PR #26, opened 2015-05-09
[WIP] SQS Implementation

TattdCodeMonkey:
Implementation of an SQS client.

Influenced by [erlcloud sqs](https://github.com/gleber/erlcloud/blob/master/src/erlcloud_sqs.erl) and [boto sqs (python)](https://github.com/boto/boto/tree/develop/boto/sqs)

### DIFF
--- a/lib/ex_aws/sqs/client.ex
+++ b/lib/ex_aws/sqs/client.ex
@@ -141,17 +141,15 @@ defmodule ExAws.SQS.Client do
     |> ExAws.Client.generate_boilerplate(opts)
 
     quote do
+      defstruct config: nil, service: :sqs
       unquote(boilerplate)
 
       @doc false
-      def request(queue_name, action, params \\ []) do
-        ExAws.SQS.Request.request(__MODULE__, queue_name, action, params)
+      def request(client, queue_name, action, params \\ []) do
+        ExAws.SQS.Request.request(client, queue_name, action, params)
       end
 
-      @doc false
-      def service, do: :sqs
-
-      defoverridable config_root: 0, request: 2, request: 3
+      defoverridable config_root: 0, request: 3, request: 4
 
     end
   end
--- a/lib/ex_aws/sqs/client.ex
+++ b/lib/ex_aws/sqs/client.ex
@@ -1,0 +1,26 @@
+defmodule ExAws.SQS.Client do
+  use Behaviour
+
+  @moduledoc """
+  """
+  
+  defmacro __using__(opts) do
+    boilerplate = __MODULE__
+    |> ExAws.Client.generate_boilerplate(opts)
+
+    quote do
+      unquote(boilerplate)
+
+      @doc false
+      def request(queue_name, action, params \\ []) do
+        ExAws.SQS.Request.request(__MODULE__, queue_name, action, params)
+      end
+
+      @doc false
+      def service, do: :sqs
+
+      defoverridable config_root: 0, request: 2, request: 3
+  
+    end     
+  end
+end
--- a/lib/ex_aws/sqs/client.ex
+++ b/lib/ex_aws/sqs/client.ex
@@ -2,8 +2,104 @@ defmodule ExAws.SQS.Client do
   use Behaviour
 
   @moduledoc """
+  The purpose of this module is to surface the ExAws.SQS API tied to a single
+  configuration chose, such that it does not need passed in with every request
+
+  Usage:
+  ```elixir
+  defmodule MyApp.SQS do
+    use ExAws.SQS.Client, otp_app: :my_otp_app
+  end
+  ```
+
+  In your config
+  ```elixir
+  config :my_otp_app, ExAws,
+    sqs: [], #SQS config goes here
+  ```
   """
+
+  @type sqs_permission :: 
+    :all | 
+    :receive_message | 
+    :delete_message |
+    :change_message_visibility | 
+    :get_queue_attributes 
+  @type sqs_acl :: [{binary, sqs_permission}]
+  @type sqs_message_attribute_name ::
+    :all |
+    :sender_id | 
+    :sent_timestamp |
+    :approximate_receive_count |
+    :approximate_first_receive_timestamp |
+    :wait_time_seconds |
+    :receive_message_wait_time_seconds 
+  @type sqs_queue_attribute_name ::
+    :all | 
+    :approximate_number_of_messages |
+    :approximate_number_of_messages_not_visible | 
+    :visibility_timeout |
+    :created_timestamp | 
+    :last_modified_timestamp | 
+    :policy  
+  @type sqs_message :: [
+    {:message_body, binary},
+    {:delay_seconds, 0..900 | none},
+    {:attributes, Keyword.t }
+  ]
+
+ @doc "Create queue"
+  defcallback create_queue(queue_name :: binary, visibility_timeout :: 0..43200 | none) :: ExAws.Request.response_t
+
+  @doc "Delete a queue"
+  defcallback delete_queue(queue_name :: binary) :: ExAws.Request.response_t
+
+  @doc "Gets attributes of a SQS Queue"
+  defcallback get_queue_attributes(queue_name :: binary, attribute_names :: [sqs_queue_attribute_name] | :all) :: ExAws.Request.response_t
+
+  @doc "Set attributes of a SQS Queue"
+  @type queue_attributes :: [
+    {:visibility_timeout, Integer} |
+    {:policy, binary}
+  ]
+  defcallback set_queue_attributes(queue_name :: binary, attributes :: queue_attributes) :: ExAws.Request.response_t
   
+  @doc "Retrieces a list of all the SQS Queues"
+  defcallback list_queues() :: ExAws.Request.response_t
+
+  @doc "Retrieves the dead letter source queues for a given SQS Queue"
+  defcallback get_dead_letter_source_queues(queue_name :: binary) :: ExAws.Request.response_t 
+  @doc "Purge all messages in a SQS Queue"
+  defcallback purge_queue(queue_name :: binary) :: ExAws.Request.response_t
+
+  @doc "Add permission"
+  defcallback add_permission(queue_name :: binary, label :: binary, permissions :: sqs_acl) :: ExAws.Request.response_t
+  
+  @doc "Remove permission"
+  defcallback remove_permission(queue_name :: binary, label :: binary) :: ExAws.Request.response_t
+  
+   @doc "Send a message to a SQS Queue"
+  defcallback send_message(queue_name :: binary, message :: sqs_message) :: ExAws.Request.response_t 
+  defcallback send_message(queue_name :: binary, message_body :: binary, delay_seconds :: 0..900 | none) :: ExAws.Request.response_t
+  
+  @doc "Send up to 10 messages to a SQS Queue in a single request"
+  defcallback send_messages(queue_name :: binary, messages :: [sqs_message, ...]) :: ExAws.Request.response_t 
+
+  @doc "Read messages from a SQS Queue"
+  defcallback receive_message(queue_name :: binary, attribute_names :: [sqs_message_attribute_name] | :all, max_number_of_messages :: 1..10, visibility_timeout :: 0..43200 | none, wait_timeout :: 0..20 | none) :: ExAws.Request.response_t
+  
+  @doc "Delete a message from a SQS Queue"
+  defcallback delete_message(queue_name :: binary, receipt_handle :: binary) :: ExAws.Request.response_t
+
+  @doc "Deletes a list of messages from a SQS Queue in a single request"
+  defcallback delete_messages(queue_name :: binary, receipt_handles :: [binary, ...]) :: ExAws.Request.response_t
+ 
+  @doc """
+  Extends the read lock timeout for the specified message from
+  the specified queue to the specified value
+  """
+  defcallback change_message_visibility(queue_name :: binary, receipt_handle :: binary, visibility_timeout :: 0..43200) :: ExAws.Request.response_t 
+
   defmacro __using__(opts) do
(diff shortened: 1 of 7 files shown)


benwilson512:
This is great stuff. Couple of general comments.

Generally speaking I use function arguments for those parameters which are Required by the API. Take http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html . Only the queue name is mandatory, the rest are optional. Take the typespec for receive_message for example made multi line for readability.
```elixir
defcallback receive_message(
    queue_name :: binary,
    attribute_names :: [sqs_message_attribute_name] | :all,
    max_number_of_messages :: 1..10,
    visibility_timeout :: 0..43200 | none,
    wait_timeout :: 0..20 | none)
``` 
If I want to specify a wait timeout but leave the rest as default, I have to specify values for the other arguments or put nil. The pattern established in the other clients is more along the lines of:

```elixir
@type receive_message_opts :: [
    {:attribute_names, [sqs_message_attribute_name] | sqs_message_attribute_name} |
    {:max_number_of_messages, 1..10} |
    {:visibility_timeout, 0..43200} |
    {:wait_timeout, 0..20}]
defcallback receive_message(queue_name :: binary)
defcallback receive_message(queue_name :: binary, opts :: receive_message_opts)
``` 
This lets me do
```elixir
receive_message("foo", wait_timeout: 10)
```
You'll notice I removed the `| none` in the typespecs because if the option is specified, it ought to have a value. I used receive_message by way of example but this applies to all of the callbacks that have "Required: No" specified in the AWS docs.

You're absolutely right to use 2 arguments to set_queue_attributes due to the two required parameters.

Great use of ranges for integer value typespecs, I totally forgot I could do that. There's a bunch of integers in the other clients where I just have `pos_integer` and ought to use a range. Thanks!

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+  ```
+  """
+
+  @type sqs_permission :: 
+    :all | 
+    :receive_message | 
+    :delete_message |
+    :change_message_visibility | 
+    :get_queue_attributes 
+  @type sqs_acl :: [{binary, sqs_permission}]
```
benwilson512: I want to avoid having types this because `[{"foo", 1}]` lacks the clarity of `%{"foo" => 1}` or `[foo: 1]` which have a clear key value aesthetic. If there isn't a good way to use atomic keys such that keyword list semantics apply, go with a map.

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+    :approximate_number_of_messages_not_visible | 
+    :visibility_timeout |
+    :created_timestamp | 
+    :last_modified_timestamp | 
+    :policy  
+  @type sqs_message :: [
+    {:message_body, binary},
+    {:delay_seconds, 0..900 | none},
+    {:attributes, Keyword.t }
+  ]
```
benwilson512: These should be `{:message_body, binary} |` not `{:message_body, binary} ,` (| not ,) because delay seconds and attributes are not required elements, and the order also does not matter. I grant that by doing so it technically allows the `{:message_body, binary}` value to be missing, so maybe
```elixir
@type sqs_message :: [
  {:message_body, binary},
  {:delay_seconds, 0..900} |
  {:attributes, Keyword.t } |
  none
]
```
is the more accurate way to do it. Thoughts?

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+  @type queue_attributes :: [
+    {:visibility_timeout, Integer} |
```
benwilson512: This should likely be the same as line 56. Perhaps there could be a `@type visibility_timeout :: 0..43200` so that the hard coded range isn't repeated.

lib/ex_aws/sqs/client.ex
benwilson512: I would say overall I want to have typespec that allows all things that the AWS service accepts, and rely on AWS to return validation errors rather than have typespec that doesn't allow something that AWS regards as valid. It's going to be a challenge to write typespec that lines up precisely with AWS, so I'm gonna err on the side of leniency in typespec so long as it at least covers all the options.

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+  @doc "Delete a queue"
+  defcallback delete_queue(queue_name :: binary) :: ExAws.Request.response_t
+
+  @doc "Gets attributes of a SQS Queue"
+  defcallback get_queue_attributes(queue_name :: binary, attribute_names :: [sqs_queue_attribute_name] | :all) :: ExAws.Request.response_t
+
+  @doc "Set attributes of a SQS Queue"
+  defcallback set_queue_attributes(queue_name :: binary, attributes :: queue_attributes) :: ExAws.Request.response_t
+  
+  @doc "Retrieces a list of all the SQS Queues"
```
benwilson512: Retrieces => Retrieves

lib/ex_aws/sqs/client.ex
TattdCodeMonkey: this doesn't compile.
```elixir
@type sqs_message :: [
  {:message_body, binary},
  {:delay_seconds, 0..900} |
  {:attributes, Keyword.t } |
  none
]
```
After reading the documentation closer, I'm like this better:
```elixir
  @type sqs_message_attribute :: %{
    :data_type => :string | :number | :binary | binary,
    :value => binary | number
  }
  @type sqs_message :: %{:message_body => binary} |
  %{:message_body => binary, :delay_seconds => 0..900} | 
  %{:message_body => binary, :attributes => [sqs_message_attribute, ...] | sqs_message_attribute } |
  %{:message_body => binary, :delay_seconds => 0..900, :attributes => [sqs_message_attribute, ...] | sqs_message_attribute }
```
Although I'm not totally sold on the attribute map inside the message map, but it does work.

Thoughts?

lib/ex_aws/sqs/client.ex
TattdCodeMonkey: agreed in the above code snippet it is  `{account_id, sqs_permission_atom}` 

it's clean and makes sense to have `{sqs_permission_atom, account_id}` instead

lib/ex_aws/sqs/client.ex
TattdCodeMonkey: :+1:

lib/ex_aws/sqs/client.ex
benwilson512: That could definitely work, although you'd want simply `[{sqs_permission, binary},...]` without the  `| {sqs_permission, binary}`.

The former defines `[all: "account_id"]` and the latter `{:all, "account_id"}` which isn't a valid Keyword arg.

However, it looks like it's petty common to set multiple permissions for a given account id, and I think using the account_ids as the key is the right option. While keyword args are the standard for function options, Maps are perfectly appropriate for regular function args (See dynamo put item). With this in mind I think we would want something like:

```elixir
SQS. add_permission("queue", "label", %{"account1" => [:receive_message, : delete_message]})
```
This gives us a typespec of:
```elixir
@type sqs_acl :: %{ binary => sqs_permission | [sqs_permission, ...]}
```

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+  
+  @doc "Remove permission"
+  defcallback remove_permission(queue_name :: binary, label :: binary) :: ExAws.Request.response_t
+  
+  @doc "Send a message to a SQS Queue"
+  defcallback send_message(queue_name :: binary, message :: sqs_message) :: ExAws.Request.response_t 
+  defcallback send_message(queue_name :: binary, message_body :: binary, delay_seconds :: 0..900 | none) :: ExAws.Request.response_t
+  
+  @doc "Send up to 10 messages to a SQS Queue in a single request"
+  defcallback send_messages(queue_name :: binary, messages :: [sqs_message, ...]) :: ExAws.Request.response_t
```
benwilson512: This should be `send_message_batch` by the way to correspond to SendMessageBatch. Trust me I prefer send_messages, but if we call it that it's no longer clear if it corresponds to AWS SendMessageBatch or if it loops over the messages file and calls `send_message`

lib/ex_aws/sqs/client.ex
benwilson512: These are some very good suggestions. I like for example that the delay_seconds parameter is part of an options keyword list, I definitely think that's where it belongs. I also like that the attributes are in that options hash for send_message. I also think your typespec for the sqs_message_attribute is spot on.

I also think you made the right choice for sqs_batch_message.

In short, :+1: on "another option". Just note the name change for send_messages to send_message_batch as mentioned in another comment.

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,122 @@
...
+  defcallback send_messages(queue_name :: binary, messages :: [sqs_message, ...]) :: ExAws.Request.response_t 
+
+  @doc "Read messages from a SQS Queue"
+  defcallback receive_message(queue_name :: binary, attribute_names :: [sqs_message_attribute_name] | :all, max_number_of_messages :: 1..10, visibility_timeout :: 0..43200 | none, wait_timeout :: 0..20 | none) :: ExAws.Request.response_t
+  
+  @doc "Delete a message from a SQS Queue"
+  defcallback delete_message(queue_name :: binary, receipt_handle :: binary) :: ExAws.Request.response_t
+
+  @doc "Deletes a list of messages from a SQS Queue in a single request"
+  defcallback delete_messages(queue_name :: binary, receipt_handles :: [binary, ...]) :: ExAws.Request.response_t
```
benwilson512: `delete_messages` => `delete_message_batch`

lib/ex_aws/sqs/client.ex
TattdCodeMonkey: :+1: I was wondering about this as I read through the AWS docs as well. I'll change the others to match "batch"  etc.

lib/ex_aws/sqs/client.ex
TattdCodeMonkey: :+1:

lib/ex_aws/sqs/request.ex
```diff
@@ -0,0 +1,28 @@
...
+      {"content-type", "application/x-www-form-urlencoded"},
+    ]
+
+    ExAws.Request.request(http_method, client.config |> url(queue_name, params), [], headers, client)
+    |> parse(client.config)
+  end
+
+  def parse({:error, result}, _), do: {:error, result}
+  def parse({:ok, body}, config) do
+    case config[:json_codec].decode(body) do
```
benwilson512: I realize you probably just copied and pasted this from another request module, but do note that (as far as I can tell) SQS only returns XML so we don't want to json parse. As of right now I don't have a policy towards the parsing of XML. JSON has a nice one to one mapping with elixir data structures. XML's structure and meta-data can vary a lot more.

I should look over the structures returned by SQS. Maybe it makes sense to parse that XML into an elixir data structure.

TattdCodeMonkey: Yea that was definitely just copy & paste, not planning on leaving it that way. this code is very raw, I just had an hour to kill with the AWS docs so I started building this out. it still needs a lot of work. notice the url function is just stubbed out with a comment on the endpoint.

I'm not sure what to do about XML yet.

benwilson512: Sounds good. I'll hold off on further review till you ping me so that I don't badger you with stuff :)

TattdCodeMonkey: No worries. I really appreciate the feedback. 

I think I will sub-branch and push here when things are a bit more polished.

benwilson512: That sounds good, I wanted to provide some early feedback on the API as defined in the Client, but I think we're on the same page there now. Note the other comment detailing a small change in the way that's handled.

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  defmacro __using__(opts) do
+    boilerplate = __MODULE__
+    |> ExAws.Client.generate_boilerplate(opts)
+
+    quote do
+      unquote(boilerplate)
+
+      @doc false
+      def request(queue_name, action, params \\ []) do
```
benwilson512: Per changes upcoming in v0.2.0 this should be `def request(client, queue_name, action, params \\ [])` and the following line should be `ExAws.SQS.Request.request(client, queue_name, action, params)`. Be sure to change the argument numbers on the defoverridable line

lib/ex_aws/sqs/client.ex
```diff
@@ -0,0 +1,158 @@
...
+    :receipt_handle => binary,
+    :visibility_timeout => visibility_timeout
+  }
+  defcallback change_message_visibility_batch(queue_name :: binary, opts :: [message_visibility_batch, ...]) :: ExAws.Request.response_t
+
+  defmacro __using__(opts) do
+    boilerplate = __MODULE__
+    |> ExAws.Client.generate_boilerplate(opts)
+
+    quote do
```
benwilson512: Per changes upcoming in v0.2.0 this should have a line with `defstruct config: nil, service: :sqs` prior to unquoting the boilerplate. Because of this, the `service/0` can be deleted.

benwilson512:
Hey! We're looking to start using SQS at cargosense, have you done any further work on this branch? If not that's totally fine, but I thought I'd let you know that we're probably going to go off what you have here and just build it out ourselves.

TattdCodeMonkey:
Thats awesome! I have not gotten a chance to get back this, I really wanted to and then time.

benwilson512:
Totally understandable. I'm gonna close this PR, but I merged your commits into the new branch / PR. Your work with the typespec will be very help, thanks!

PR #33, opened 2015-05-16
Per request configuration

### LINKED ISSUE #31: Safely change AWS credentials per request
I'm looking for a way to safely change AWS credentials for each requests, using the request I determine which account and bucket to use.

Thanks for ExAws and your help! :+1:

benwilson512:
Fixes #31

### DIFF
--- a/lib/ex_aws/client.ex
+++ b/lib/ex_aws/client.ex
@@ -22,11 +22,16 @@ defmodule ExAws.Client do
       @otp_app Keyword.get(unquote(opts), :otp_app)
       @behaviour unquote(client)
 
+      def new(opts \\ []) do
+        %__MODULE__{config: config}
+        |> struct(opts |> Enum.into(%{}))
+      end
+
       @doc false
       def config_root, do: Application.get_env(@otp_app, :ex_aws)
 
       @doc false
-      def config, do: __MODULE__ |> ExAws.Config.get
+      def config, do: %__MODULE__{} |> ExAws.Config.get
     end
   end
 
@@ -43,14 +48,12 @@ defmodule ExAws.Client do
       {:., [], [impl_module, name]}, [],
       # Append client_data to the function arguments passed to
       # The implementation function.
-      [client_data | arguments]
+      [new_client_struct | arguments]
     }]]}
   end
 
-  def client_data do
-    quote do
-      %{module: __MODULE__, config: config()}
-    end
+  def new_client_struct do
+    quote do: __MODULE__.new
   end
 
   # ExAws.Dynamo.Client #=> ExAws.Dynamo.Impl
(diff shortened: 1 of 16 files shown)


benwilson512:
@doomspork how does this look with Dynamo? Assuming you use the regular `config :ex_aws` in mix you could do:

```elixir
#somewhere
def config_with_keys({access_key, secret_key}) do
 # Inherit general configuration from mix / defaults
  config = ExAws.Dynamo.config
  |> Keyword.put(:access_key_id, access_key)
  |> Keyword.put(:secret_access_key, access_key)
  
  %{module: ExAws.Dynamo, config: config}
end

# In your request
determine_aws_auth # do whatever you do
|> config_with_keys
|> ExAws.Dynamo.Impl.list_tables
```

How's that look?

benwilson512:
The whole `client_data.module.request(:create_table, client_data)` feels a little verbose, I'll have to think about how to maybe make that less so.

doomspork:
@benwilson512 this looks great.  

When you don't pass `config` will it fall back to `config.exs`?

benwilson512:
It does not at the moment. The way things work right now (in master) is that the Impl functions take as their first argument the module name of a client. Thus a function call to Impl works like:
```elixir
ExAws.Dynamo.Impl.describe_table(ExAws.Dynamo, "MyTable")
```
Per this PR, it now works like
```elixir
ExAws.Dynamo.Impl.describe_table(%{module: ExAws.Dynamo, config: config}, "MyTable")
```
Given that, what do you mean by fallback? Allow the bare module calls as well as the mod / config map?

I definitely do NOT want to do something where the Impl modules fall back implicitly to some _particular_ client or configuration, that's precisely some of the erlcloud functionality I'm trying to get away from.

benwilson512:
Perhaps by way of clarification, if you wanted to just use the config of the ExAws.Dynamo client while still using the %{} semantics you can do
```elixir
ExAws.Dynamo.Impl.describe_table(%{module: ExAws.Dynamo, config:  ExAws.Dynamo.config}, "MyTable")
```
I'm inclined to say that the Impl modules ought to only support the latter construct. I can't imagine anyone explicitly calling Impl functions unless they're in the business of doing per-request configurations, in which case they gain nothing by the ability to pass in a bare module. Am I missing any scenarios that you can think of?

benwilson512:
Had some new thoughts, and updated the prior comment.

benwilson512:
So per the latest commit, using this is now as simple as:
```elixir
# In your request
{access_key, secret_key} = determine_aws_auth # do whatever you do

[config: [access_key_id: access_key, secret_access_key: secret_key]
|> ExAws.Dynamo.new
|> ExAws.Dynamo.Impl.list_tables
```

doomspork:
@benwilson512 this is really great, thank you so much for all your hard work.

benwilson512:
Awesome I think this is a good way to go as well. Having the client modules create a struct makes a lot of sense, should have done that to start!

While these changes won't affect ordinary usage, anyone who created a client and overrode the request function for that client will need an additional argument. Ergo, breaking change. Will release as 0.2.0 after I bring the other services up to speed.

doomspork:
Sounds great, thanks again!

PR #40, opened 2015-06-10
[WIP] SNS implementation

bruce:
Starting the SNS implementation. Not currently functional.

API docs at http://docs.aws.amazon.com/sns/latest/api/API_Operations.html


bruce:
- [ ] So far the [very] basics are in for topic-related actions, but (of course) it appears the requests for SNS have different semantics (eg, no pretty `x-amz-target` header, etc), so plenty of work left in `ExAws.SNS.Request` before things are functional.
- [ ] There are paginated actions (eg, `ListTopics`), so a `ExAws.SNS.Lazy` module needs to be built.

lib/ex_aws/utils.ex
```diff
@@ -28,13 +28,13 @@ defmodule ExAws.Utils do
     end
   end
 
-  defp camelize_key(key) when is_atom(key) do
+  def camelize_key(key) when is_atom(key) do
```
bruce: Seems reasonable to make these public for more ad-hoc camelization.

bruce:
- [ ] `publish` implementation (probably the most immediately useful SNS feature)

benwilson512:
We should also discuss parsing the XML. It looks like the responses follow a sufficiently regular format that parsing it into Elixir Maps seems to make sense.

bruce:
I agree, and that could be made pluggable -- similar to the Dynamo decoder, I'm guessing.

benwilson512:
I agree, although there are some unique challenges to allowing multiple XML decoders vs JSON decoders. JSON decoding basically just consists of a encode and decode functions that take simple single arguments. XML decoding requires something like an xpath function that often has pretty different optional arguments that turn out to be pretty useful (sweet_xml). We'll have to look at it.

lib/ex_aws/sns/request.ex
```diff
@@ -0,0 +1,23 @@
+defmodule ExAws.SNS.Request do
+  @moduledoc false
+  # SNS specific request logic.
+
+  @empty_body_hash AWSAuth.Utils.hash_sha256("")
+
+  def request(client, action, params) do
+    query = params
+    |> Map.put("Action", Mix.Utils.camelize(Atom.to_string(action)))
+    |> URI.encode_query
```
benwilson512: So the issue at the moment is that the AWSAuth library we use for signing at the moment ends up doubly encoding the parameters, because it calls URI.encode as part of its signing process. This is rather annoying, because it means you can't pass it a pre-encoded URL.

I'm pretty sure we want to just move the signing stuff in house at this point, AwsAuth provides a decent example, but suffers from having everything built into a couple massive functions.

benwilson512: OK publishing works now. In order to make this happen this branch of ExAws now depends on benwilson512/aws_auth instead of the published aws_auth package.

Before merging this branch into master I'd like to probably move auth in house so that there isn't a git based dependency on master. However, this at least lets us use this branch for our own purposes in the mean time.

lib/ex_aws/sns/client.ex
```diff
@@ -0,0 +1,149 @@
...
+  defcallback delete_topic(topic_arn :: topic_arn) :: ExAws.Request.response_t
+
+  ## Publishing
+  ######################
+
+  @type message_attribute :: %{
+    :name => binary,
+    :data_type => :string | :number | :binary,
+    :value => {:string, binary} | {:binary, binary}
+  }
```
benwilson512: @bruce let me know what you think about this typespec for message attributes. While keyword lists I think are appropriate for function options, a message attribute is really an entity in and of itself, which I prefer to use maps for. Additionally since you can have several of them, you end up with a list of keyword lists which can be visually confusing.

bruce:
Yeah, I'm familiar with the data_type switching; I struggled briefly to find the right pattern, this seems like a good middle-of-the-road approach.

TattdCodeMonkey:
As far as XML goes what are your thoughts on exml? https://github.com/paulgray/exml

I know a few people using it successfully in elixir projects.

lib/ex_aws/sns/request.ex
benwilson512: This has now been accomplished.

benwilson512:
A lot of the XML concerns we had before are largely solved now that I have an approach settled with SweetXML.

jeregrine:
@benwilson512 is there any movement on this regarding the auto-generation of the API?

benwilson512:
Hey. Autogenerated stuff will likely not hit till after ElixirConf (next weekend). There's simply been too much happening to make that a priority right now. There's several days worth of work involved, and that's several days I just haven't had unfortunately. I apologize on the delays.

jeregrine:
No worries.

vitalis:
@bruce , @benwilson512 are you planning to release this feature soon?

ASCrookes:
@bruce @benwilson512: Any plans to merge this into master soon? Also anything I could help out with if it isn't at a stable point, would love to get involved if so.

dstendardi:
also really intersted on this feature :-)

swennemans:
Me too. Would be awesome to send email, push  notifications and SMS using `ex_aws` :)

benwilson512:
Merging what we have here for now.

PR #45, opened 2015-06-30
Make use of keyword lists

mschae:
A few small things...


benwilson512:
Some of these have been integrated in another branch. Thanks for the ideas!

PR #46, opened 2015-07-08
Fix HTTP request issues

### LINKED ISSUE #47: Uploads to S3 failing
This is going to be a very unsatisfactory issue report, I apologise in advance.

This was reported by @ericmj and experienced by myself in one of our projects.

It seems that sometimes uploads to S3 stop working. No error is being raised and it looks like they just stall. Data doesn't make it into S3 but again: No error.

Best guess right now is that requests are timing out and the pool only holds one connection.

Possible fixes:

- Timeouts (e.g. recv_timeout, pool timeout, etc.)
- Increase pool size

benwilson512:
Fixes #47


lib/ex_aws/request/httpoison.ex
```diff
@@ -0,0 +1,14 @@
+defmodule ExAws.Request.HTTPoison do
+  @behaviour ExAws.Request.HttpClient
+
+  @moduledoc false
+
+  @overriding_opts [
+    recv_timeout: 5_000,
+    hackney: [pool: false]
+  ]
```
benwilson512: @mschae Do you know a good way to verify that a request is made without using the pool?

mschae:
:+1: thanks man! Appreciated!

PR #48, opened 2015-07-08
Remove timex

benwilson512:


lib/ex_aws/utils.ex
```diff
@@ -47,4 +47,29 @@ defmodule ExAws.Utils do
   def upcase(value) when is_binary(value) do
     String.upcase(value)
   end
+
+  @seconds_0_to_1970 :calendar.datetime_to_gregorian_seconds({{1970, 1, 1}, {0, 0, 0}})
+
+  def iso_z_to_secs(<<date::binary-10, "T", time::binary-8, "Z">>) do
+    <<year::binary-4, "-", mon::binary-2, "-", day::binary-2>> = date
+    <<hour::binary-2, ":", min::binary-2, ":", sec::binary-2>> = time
```
benwilson512: On the one hand, I love binary pattern matching. On the other hand, stdlib time handling sucks.

PR #55, opened 2015-07-17
Yet more s3

benwilson512:


benwilson512:
- [ ] Handling leading "/" chars on put_object_copy src objects.

PR #60, opened 2015-08-25
Add Cloud Search

mschae:
This adds cloud search.

There are currently still a few things to do:

- [x] Add documentation
- [ ] Add tests
- [ ] Implement `stream_search` that will be an enumerable that implements pagination

Also we should consider a few improvements:

- Enable users to use custom data structure (e.g. indexing and retuning a `%User{}`)
- Support more complex search queries
- Add more search options

### DIFF
--- a/lib/ex_aws/cloud_search/document.ex
+++ b/lib/ex_aws/cloud_search/document.ex
@@ -1,5 +1,16 @@
 defprotocol ExAws.CloudSearch.Document do
-  @fallback_to_any true
+  @doc """
+  The encode function returns a hash that follows the data structure of documents
+  in AWS Cloud Search as found [here](http://docs.aws.amazon.com/cloudsearch/latest/developerguide/preparing-data.html#adding-documents)
+  and [here](http://docs.aws.amazon.com/cloudsearch/latest/developerguide/preparing-data.html#deleting-documents).
 
+  More specifically there are two types of documents:
+
+  * Add: `{"type":"add","id":"<id>",fields:{"<field name>":"<field value>"}}`
+  * Delete: `{"type":"delete","id":"<id>"}
+
+  ExAws ships with a generic add and a generic delete document but by implementing
+  this protocol you can build you own document modules (or augment existing modules).
+  """
   def encode(document)
 end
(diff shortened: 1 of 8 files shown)


lib/ex_aws/cloud_search/document.ex
```diff
@@ -0,0 +1,5 @@
+defprotocol ExAws.CloudSearch.Document do
+  @fallback_to_any true
```
benwilson512: Curious about this choice. Are there data types other than the structs that we want to support?

lib/ex_aws/cloud_search/request.ex
```diff
@@ -0,0 +1,37 @@
...
+    {:ok, config[:json_codec].decode!(body)}
+  end
+
+  defp url(%{search_endpoint: search_endpoint}, path = "/search" <> _) do
+    Path.join(search_endpoint, Enum.join([@version, path]))
+  end
+  defp url(%{document_endpoint: document_endpoint}, path = "/documents" <> _) do
+    Path.join(document_endpoint, Enum.join([@version, path]))
+  end
+  defp url(_, _) do
```
benwilson512: Nice, I should do this more often.

lib/ex_aws/cloud_search/document.ex
mschae: None, really. I guess this should rather be a behaviour right?

benwilson512: Well if there are cases where I may have some `%User{}` and want to provide some customization with respect to the way in which it's encoded, I think a protocol is fine. I'm not really all that familiar with cloudsearch yet so it isn't entirely clear if such a situation is what we want.

The particular comment here was less about the use of a protocol in and of itself, and more about whether or not we want to fall back to any vs enforce particular implementations to exist.

mschae: IC so the comment was about the `@fallback_to_any`. And you're right, we don't need it.

I will add documentation cause (a) it's nice and (b) it'll explain my choice a little.

benwilson512:
Hey there!

I realize it's been an eon since this was created. I'm just letting you know that we have dramatically simplified the structure of ExAws as we go into 1.0. We would be happy to look at a contribution for Route53 again, but I'm gonna close this for now as most of what is here is related to boilerplate only necessary pre 1.0.

Thanks!

- Ben

PR #67, opened 2015-09-03
Encoding Nested Structs

### LINKED ISSUE #64: Dynamo encoding for nested structs
If a struct is nested inside another struct, dynamo will not encode it as a map properly. While other maps when encoded return `%{"M" => encoded_map}` structs just return `encoded_map` so that they can be used directly in Dynamo queries.

Ideally, current usage of `encode/1` for structs would be replaced by an `encode_root/1` or `encode(item, :root)` function, and `encode/1` would always return the `%{"M" => encoded_map}` form. How much of a breaking change this is needs to be evaluated.

benwilson512:
Fixes #64

### DIFF
--- a/lib/ex_aws/dynamo/encodable.ex
+++ b/lib/ex_aws/dynamo/encodable.ex
@@ -3,66 +3,99 @@ defprotocol ExAws.Dynamo.Encodable do
   @type t :: any
 
   @doc "Converts an elixir value into a map tagging the value with its dynamodb type"
-  def encode(value)
+  def encode(value, options)
 end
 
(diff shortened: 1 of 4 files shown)


benwilson512:
Note this introduces some breaking changes.
- `Encoder.encode_flat/1` is now `Encoder.encode_root`
- When arg in `Encoder.encode(arg)` is a struct, it will now be encoded like `%{"M" => encoded_struct}` like all other maps in order to facilitate nesting. If you want the old behaviour where it was merely `encoded_struct` use `encode_root`

PR #69, opened 2015-09-04
Implement Endpoints

adamkittelson:
There's some noise in the diff here because my sqs branch and yours have varying amounts of master merged in, but the gist here is that I've implemented and tested all the endpoint functions.

The tests in sqs_test.exs should make it easy to do any refactoring of the impl.ex functions. I've also manually integration tested them all against my AWS account in IEx so as long as the unit tests still pass after any changes they should still work in practice.

I haven't done anything with XML response parsing yet but thought this would be a good point to get my changes synced up with the primary SQS branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,18 @@
 language: elixir
-elixir: 1.0.4
+elixir: 1.0.5
 notifications:
   recipients:
     - ben.wilson@cargosense.com
 otp_release:
   - 17.1
+  - 17.5
+  - 18.0
+before_script:
+  - wget http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz -O /tmp/dynamodb_local_latest.tar.gz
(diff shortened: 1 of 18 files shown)


adamkittelson:
Travis failures look to be lack of AWS creds, not sure if those need to be entered in travis on a branch by branch basis or something.

benwilson512:
This is fantastic! I have to update the permissions via my IAM console. Will do that here in a bit and do a more thorough review over the weekend.

adamkittelson:
Hope you had a good holiday weekend (assuming you're in the states). Just checking if you've had a chance to take a look at this or if you've got any questions for me on it.

benwilson512:
Hey, would you mind merging master in again? I believe it ought to clear up the diff noise a bit.

benwilson512:
actually, this looks really good. I'm gonna merge it into the SQS branch and then handle catching up with master there.

PR #80, opened 2015-09-18
Dynamo: removed double encoding when encoding lists.

mr-bt:


mr-bt:
The build failed... I can't run the unit tests locally either but I'm pretty sure is nothing with my commit.

benwilson512:
The tests for the Encoder are found here: https://github.com/CargoSense/ex_aws/blob/master/test/lib/ex_aws/dynamo/encoder_test.exs. Can you please run them locally via.
```
mix test test/lib/ex_aws/dynamo/encoder_test.exs
```

Can you elaborate a bit on your changes?

benwilson512:
Also, the reason the tests are failing is because your branch is 10 commits behind master. Please rebase off master.

benwilson512:
Would prefer if you would rebase instead of merge, thanks.

mr-bt:
Sure, it makes sense. I'll cancel this PR and make another one.

benwilson512:
Thanks, not trying to be a pain but it can help the commit history readability a lot particularly when it's a small PR.

PR #83, opened 2015-09-20
Dynamo: fixed double list encoding.

mr-bt:
On the Enum.map/2 (line 114) the list is encoded as a generic dynamo list ([%{"S" => "abc"}, %{"N" =>  "123"}]. On the Enum.reduce/3 (line 116) the list is encoded again (line 118). The reduction happens to extract the list types which is used on the case statement (line 124) to check which dynamo list should be used.

This pull request fixes double list encoding by using a Enum.map_reduce/2 whereby the list is encoding and the list type is set: "S", "N", "B", generic ("M" or mixed values).


mr-bt:
@bruce the unit test are still failing... Nothing to do with my commit tough.

benwilson512:
Yeah I can't really explain that. Super weird. Will look at this today, thanks!

benwilson512:
Yup this looks good, thanks!

PR #87, opened 2015-09-24
Implement S3 presigned_url API function

jordi-chacon:
Hi Ben,

this is a very early phase PR (i.e. I am not done yet). I'd very much like to get some early feedback from you before I proceed.

My initial idea was to change as little as possible from the existing code, simply add on. But then I found that `auth.ex` had a lot of bits and pieces of what I needed, so I was kind of forced to refactor it to avoid duplication. I hope you are OK with this.

I have not tested this, not even manually, I just got it to a state where it compiles and looks like it could work. I will probably continue tomorrow. But I just would like to get a thumbs up or thumbs down from you in regards to what I have done so far.

I also have a question for you: I see that function `ExAws.S3.Request.url` is used to construct S3 URLs. At first I thought I could use it, but then I saw this rule about the `bucket` containing a `.` which I didn't understand. Why this rule? For the `presigned_url` function, there is a `virtual_host` flag which indicates whether I should include the `bucket` in the hostname or in the path. So for now I chose to build the URL myself until I get a better understanding of the existing function and find out if I can reuse it or not.

Thanks a lot Ben!

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -4,87 +4,59 @@ defmodule ExAws.Auth do
   @moduledoc false
 
   def headers(http_method, url, service, config, headers, body) do
-    now = :os.timestamp |> :calendar.now_to_universal_time
+    datetime = :calendar.universal_time
     headers = [
       {"host", URI.parse(url).host},
-      {"x-amz-date", amz_date(now)} |
+      {"x-amz-date", amz_date(datetime)} |
       headers
     ]
     |> handle_temp_credentials(config)
 
     auth_header = auth_header(
-      config[:access_key_id],
-      config[:secret_access_key],
-      http_method |> method_string,
+      http_method,
       url,
-      config[:region],
-      service |> service_name,
       headers,
       body,
-      now)
+      service |> service_name,
+      datetime,
+      config)
 
     [{"Authorization", auth_header} | headers ]
   end
 
-  def handle_temp_credentials(headers, %{security_token: token}) do
+  def presigned_url(http_method, url, service, config, expires) do
+    datetime = :calendar.universal_time
+    headers = presigned_url_headers(url)
+    query = presigned_url_query(datetime, service, config, expires)
+    url = "#{url}?#{query}"
+    signature = signature(http_method, url, headers, nil, service, datetime, config)
+    "#{url}&X-Amz-Signature=#{signature}"
+  end
+
+  defp handle_temp_credentials(headers, %{security_token: token}) do
     [{"X-Amz-Security-Token", token} | headers]
   end
-  def handle_temp_credentials(headers, _), do: headers
+  defp handle_temp_credentials(headers, _), do: headers
 
-  def auth_header(access_key, secret_key, http_method, url, region, service, headers, body, {date, _} = now) do
-    date  = date |> quasi_iso_format
-    scope = "#{date}/#{region}/#{service}/aws4_request"
-
-    signing_key = build_signing_key(secret_key, date, region, service)
-
-    signature = http_method
-    |> build_canonical_request(url, headers, body)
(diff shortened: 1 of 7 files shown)


jordi-chacon:
Hi again Ben,
now I have completed the implementation and the testing of URL presigning for S3.

benwilson512:
Hey thanks! I should be able to look through this over the weekend. I'll try to figure out why travis-ci with PRs doesn't work as well.

benwilson512:
http://docs.travis-ci.com/user/pull-requests/ is why the tests won't pass, which makes sense. I should tag the tests that do actual requests to AWS and use the `TRAVIS_PULL_REQUEST` env variable to not run them.

jordi-chacon:
Just wanted to mention that I have done two manual tests:
* Generate a presigned URL for `PUT`, and then upload a text file to S3 by curling such URL. (`curl -L --upload-file foo.txt "url"`)
* Generate a presigned URL for `GET`, and then fetch a text from S3 by curling such URL. (`curl -L "url"`)

lib/ex_aws/auth.ex
```diff
@@ -97,26 +69,58 @@ defmodule ExAws.Auth do
     |> Keyword.keys
     |> Enum.join(";")
 
+    payload = case body do
+                nil -> "UNSIGNED-PAYLOAD"
+                _ -> ExAws.Auth.Utils.hash_sha256(body)
+              end
```
benwilson512: As a minor coding style thing, I would prefer
```
payload = case body do
    nil -> "UNSIGNED-PAYLOAD"
    _ -> ExAws.Auth.Utils.hash_sha256(body)
end
```

This is the style used by the rest of the code base and I would prefer to keep that consistent.

lib/ex_aws/s3/client.ex
```diff
@@ -411,6 +416,16 @@ defmodule ExAws.S3.Client do
...
 
   @doc """
+  Generates a pre-signed URL for this object.
+
+  Raises an ArgumentError if `:expires_in` exceeds one week.
+
+  When option param :virtual_host is `true`, the {#bucket} name will be used as
+  the hostname. This will cause the returned URL to be 'http' and not 'https'.
+  """
+  defcallback presigned_url(http_method :: atom, bucket :: binary, object :: binary, opts :: presigned_url_opts) :: binary
```
benwilson512: Related to the comment on the implementation, I believe the return specs here should be `{:ok, binary} | {:error, binary}`

lib/ex_aws/s3/impl.ex
```diff
@@ -441,8 +441,28 @@ defmodule ExAws.S3.Impl do
     |> Parsers.parse_list_parts
   end
 
+  def presigned_url(client, http_method, bucket, object, opts \\ []) do
+    expires_in = Keyword.get(opts, :expires_in, 3600)
+    virtual_host = Keyword.get(opts, :virtual_host, false)
+    raise_if_exceeds_one_week(expires_in)
```
benwilson512: Hey, so I would prefer if we did not raise here, and instead returned an {:error, reason} tuple. Correspondingly we should have a return of `{:ok, url}` on success.

benwilson512:
Hey so other than the relatively minor tweaks this looks great. I really appreciate that you wrote comprehensive tests and typespec.

jordi-chacon:
Thanks for the feedback Ben! I have now done the minor tweaks you requested.

jordi-chacon:
I might have found a bug in my code regarding the `X-Amz-Security-Token`. Let's move this branch back to "not ready". I'll get back to you.

jordi-chacon:
It is fixed now.

benwilson512:
I'm calling this good to go, should publish it to hex later today. Thanks!

jordi-chacon:
Excellent! Thank you!

PR #104, opened 2016-01-09
Add AWS MachineLearning#Predict

dimroc:
Interacts with AWS to retrieve a real-time prediction from a previously created ML Model as described here:

http://docs.aws.amazon.com/machine-learning/latest/APIReference/API_Predict.html

Implementation includes two tests:
1. An integration test that expects a Record with "Text" to return a predictedLabel (Categorical target).
2. A stubbed test that ensures the Record map sent is accurate.

Any feedback is appreciated!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ erl_crash.dump
 /mnesia/test
 /mnesia/prod
 *.pem
+.envrc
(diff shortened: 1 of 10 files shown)


benwilson512:
Hey there!

I realize it's been an eon since this was created. I'm just letting you know that we have dramatically simplified the structure of ExAws as we go into 1.0. We would be happy to look at a contribution for Machine Learning again, but I'm gonna close this for now as most of what is here is related to boilerplate only necessary pre 1.0.

Thanks!

- Ben

PR #108, opened 2016-01-30
First cut of Route53

micahwedemeyer:
Note: I don't expect this to get merged since it's so raw, but I'm just curious if I'm on basically the right track here. I'm new to Elixir, and I'm mostly just cargo-culting the SQS and S3 implementations.

I got the Route 53 integration test to pass, meaning that it's hitting the AWS API with all the correct params.

If you want to run the test yourself, it's actually pretty easy. You can just set up a hosted zone in Route 53, even one you don't own. Just add Route 53 to your AWS account and add "example.org" as a hosted zone.

So, anyways, let me know if I'm stumbling in roughly the right direction.

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -43,6 +43,13 @@ defmodule ExAws.Config.Defaults do
         host: {"$region", "sqs.$region.amazonaws.com"},
         region: "us-east-1",
         port: 80
+      ],
+      route53: [
+        scheme: "https://",
+        host: {"$region", "route53.amazonaws.com"},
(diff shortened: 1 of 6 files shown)


benwilson512:
Hey there!

I realize it's been an eon since this was created. I'm just letting you know that we have dramatically simplified the structure of ExAws as we go into 1.0. We would be happy to look at a contribution for Route53 again, but I'm gonna close this for now as most of what is here is related to boilerplate only necessary pre 1.0.

Thanks!

- Ben

micahwedemeyer:
Thanks for the heads up. Good luck!

PR #110, opened 2016-02-03
S3: Added parsing for CommonPrefixes when listing objects

lbarratt:
This PR extends the S3 parser to include CommonPrefixes when listing objects

Thanks!


benwilson512:
This looks good, thanks!

PR #117, opened 2016-02-18
Switch to using :crypto.hash/2 to remove deprecation warning

cbarratt:
Warning: crypto:md5/1 is deprecated and will be removed in in a future release; use crypto:hash/2


benwilson512:
Thanks!

PR #127, opened 2016-03-15
switch relative path when host doesn't include 's3.' in url 

letusfly85:
I want to use below library on my local for development.

* https://github.com/jamhall/s3rver


benwilson512:
Hey!

Thanks for the PR. My initial reaction here is :-1: because it seems like this change only exists to help compatibility with one particular fake s3 implementation. It seems like some `/etc/hosts` entries are enough to make the existing functionality work, and even if that isn't enough for this particular implementation, that strikes me as a problem with `s3rver` for not handling the AWS S3 spec.

letusfly85:
@benwilson512

Thank you for replying, I see :smile:

I will see more about s3rver.

Thank you.

PR #130, opened 2016-03-22
Refactor client for per-request region selection

aeturnum:
- Clients no longer resolve host at construction
- Clients now resolve host at each request
- S3 functions with options now have support cient: options
- The only implemented client: option is {region: binary}, selecting the
  region

### DIFF
--- a/lib/ex_aws/s3/utils.ex
+++ b/lib/ex_aws/s3/utils.ex
@@ -60,6 +60,23 @@ defmodule ExAws.S3.Utils do
     |> format_and_take(param_list)
   end
 
+  def apply_client_options([], client), do: client
+  def apply_client_options(opts, %{config: config} = client) do
+    config = opts
+    |> Enum.into(%{})
+    |> set_client_region(config)
+
+    Map.put(client, :config, config)
+  end
+
+  defp set_client_region(opts, %{host: host} = config) when is_binary(host), do: config
+  defp set_client_region(opts, config) do
+    case Map.get(opts, :region) do
+      nil -> config
+      new_region -> Map.put(config, :region, new_region)
+    end
+  end
+
   @acl_headers [:acl, :grant_read, :grant_write, :grant_read_acp, :grant_write_acp, :grant_full_control]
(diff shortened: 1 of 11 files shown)


lib/ex_aws/s3/utils.ex
```diff
@@ -60,6 +60,23 @@ defmodule ExAws.S3.Utils do
     |> format_and_take(param_list)
   end
 
+  def apply_client_options([], client), do: client
+  def apply_client_options(opts, %{config: config} = client) do
```
benwilson512: This is unnecessary. You can already override the client by using the `ExAws.S3.Impl` functions which is the correct way to do this.

benwilson512:
Hey there can you talk a bit about the motivation? You can already override client config on a per request basis, and I'm not sure what the benefit is from parsing the host later. This may also mess up using dynamodb local, although I haven't had a chance to test that yet.

aeturnum:
All the tests passed on my machine - using dynamodb local. I also use fakes3 in my project and it behaved as expected with these changes. The host selection code still uses the old technique - which supported custom hosts.

The motivation for this is based on how the client is initially constructed for all services - specifically `parse_host_for_region`. At least for the S3 (and perhaps for other AWS services) client object, the region is used to select from many possible hosts (stored in `ExAws.Config.Defaults`). This process creates a final client object that is only aware of the host for one S3 region - meaning that to make a request to another region you need a new client (or need to directly re-write the host).

I'm pretty new to Elixir and it's entirely possible I missed another option. It did not seem like it was possible to switch the region of a client after that client had been created (short of custom code that specifically replaces the host field).

I thought a reasonable solution was to simply move the last step of client construction (`parse_host_for_region`) to the pre-request level and let the map of %{region => host} move around with the client. This should be completely transparent to clients that do not wish to switch regions.

benwilson512:
Why not initialize a new client per request? It isn't a particularly expensive operation.
```
ExAws.S3.new(region: "eu-west-1")
|> ExAws.S3.Impl.list_objects("my-bucket")
```

PR #131, opened 2016-03-23
Implement put_bucket_notification

evax:


lib/ex_aws/s3/utils.ex
```diff
@@ -84,6 +84,41 @@ defmodule ExAws.S3.Utils do
...
+      queue: "Queue",
+      cloud_function: "CloudFunction"
+    ]
+    ["<#{types[config.type]}>",
+     "<Id>", config.id, "</Id>",
+     build_notification_config_filter(config[:filter]),
+     "<#{arn[config.type]}>", config.arn, "</#{arn[config.type]}>",
+     Enum.map(config.events, &("<Event>#{&1}</Event>")),
+     "</#{types[config.type]}>"]
+  end
```
benwilson512: Hey there! First off <3 for working on this, it's greatly appreciated.
I'd like to do this mapping of types / arn a little differently. I think we either should simply camelize whatever they give us and let AWS validate that it's correct, or we ought to enforce that they give a value for which we have a key.

Right now if they give `topic_configuration` for example because that's what they think it should be then the value that gets sent to AWS will be `nil`, and they won't know why easily.
nil, which will be confusing.

evax:
Something like this?
```elixir
  def build_notification_config_section(config) do
    required_keys = ~w(topic arn events)a
    supported_types = ~w(topic queue cloud_function)a
    Enum.each(required_keys, fn k ->
      unless Map.has_key?(k), do: raise "missing key #{k}"
    end)
    unless config.type in ~w(topic queue cloud_function) do
      supported = Enum.map_join(supported_types, " or ", &(inspect &1))
      raise "unsupported notification type #{config.type}, must be #{supported}"
    end
    arn_tag = Macro.camelize(Atom.to_string(config.type))
    tag = arn_tag <> "Configuration"
    ["<#{tag}>",
     "<Id>", config.id, "</Id>",
     build_notification_config_filter(config[:filter]),
     "<#{arn_tag}>", config.arn, "</#{arn_tag}>",
     Enum.map(config.events, &("<Event>#{&1}</Event>")),
     "</#{tag}>"]
  end
```

lib/ex_aws/s3/utils.ex
evax: Something like this?

```elixir
  def build_notification_config_section(config) do
    required_keys = ~w(topic arn events)a
    supported_types = ~w(topic queue cloud_function)a
    Enum.each(required_keys, fn k ->
      unless Map.has_key?(k), do: raise "missing key #{k}"
    end)
    unless config.type in ~w(topic queue cloud_function) do
      supported = Enum.map_join(supported_types, " or ", &(inspect &1))
      raise "unsupported notification type #{config.type}, must be #{supported}"
    end
    arn_tag = Macro.camelize(Atom.to_string(config.type))
    tag = arn_tag <> "Configuration"
    ["<#{tag}>",
     "<Id>", config.id, "</Id>",
     build_notification_config_filter(config[:filter]),
     "<#{arn_tag}>", config.arn, "</#{arn_tag}>",
     Enum.map(config.events, &("<Event>#{&1}</Event>")),
     "</#{tag}>"]
  end
```

PR #135, opened 2016-04-05
RDS implementations (basic operations)

sashaafm:
Hello :) I'm using ex_aws for a project for my master's thesis (only the S3 wrapper so far) and I need to use RDS also. Since you don't have a RDS implementation/wrapper I decided to try to contribute and do it myself instead of building a personal wrapper that would then go to waste.

I only need some basic operations from the RDS API (mainly the ones from the Instances endpoint and a couple from the Databases endpoint) and since I'm short on time this will be my main focus. However if all goes well I'll try to do a full implementation. 

I've read the Contributing file and I'll stick to the guidelines :+1:


sashaafm:
Hello @benwilson512 :) When possible, would you mind giving me some feedback about this PR so far? 

Thank you :+1:

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,128 @@
...
+    query_params = Map.new
+    |> Map.put_new("Action", "CreateDBInstance")
+    |> Map.put_new("MasterUsername", username)
+    |> Map.put_new("MasterUserPassword", password)
+    |> Map.put_new("AllocatedStorage", storage)
+    |> Map.put_new("DBInstanceIdentifier", instance_id)
+    |> Map.put_new("DBInstanceClass", class)
+    |> Map.put_new("Engine", engine)
+    |> Map.put_new("Version", @version)
+    |> Map.merge(opts)
```
benwilson512: Hey! Minor stylistic note. Why all the put_new here? For one thing, you start with a `Map.new` so you can already be sure that the map doesn't already contain the given key. `put_new` requires a lookup with every insert and while the performance difference will hardly matter I also just don't see the point.

I would prefer all of these to be re-written in the map literal style:

```elixir
query_params = %{
  "Action" => "CreateDBInstance",
  "MasterUsername" => username,
  "MasterUserPassword" => password,
  "AllocatedStorage" => storage,
  "DBInstanceIdentifier" => instance_id,
  "DBInstanceClass" => class,
  "Engine" => engine,
  "Version" => @version,
}
```

benwilson512: There is an additional issue. Right now you simply merge in the `opts`, but this requires that people use `"Engine"` or whatever as keys in the map. One of the more useful things about `ExAws` is that I allow folks to use `db_instance_identifier: "foo"` atom style keys instead of the binary keys that get sent to AWS. See the other clients for details. There are quite a few utility functions that already exist for transforming atom keys given in options to binary keys.

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,128 @@
...
+  def describe_db_instances(client, opts \\ %{}) do 
+    query_params = Map.new
+    |> Map.put_new("Action", "DescribeDBInstances")
+    |> Map.put_new("Version", @version)
+    |> Map.merge(opts)
+
+    request(client, :get, "/", params: query_params)
+  end
+
+  def create_db_instance(client, instance_id, username, password, storage, class, engine, opts \\ %{}) do
```
benwilson512: Options in every case should be a keyword list, and be converted to a map inside this function. Keyword lists have substantially nicer syntax for users of this library.

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  def reboot_db_instance(client, instance_id, failover \\ :empty) do
+    query_params = Map.new
+    |> Map.put_new("Action", "RebootDBInstance")
+    |> Map.put_new("DBInstanceIdentifier", instance_id)
+    |> Map.put_new("Version", @version)    
+
+    if failover != :empty do 
+      query_params = Map.merge(query_params, %{"Failover" => failover})
```
benwilson512: This needs refactored to avoid the imperative warning. IE

```elixir
query_params = if failover != :empty do
  Map.put(query_params, "Failover", failover)
else
  query_params
end
```

lib/ex_aws/rds/impl.ex
benwilson512: username and password are not listed as required on http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html. This means that they should be handled via options. Only those parameters which are required should be dedicated function arguments. All optional parameters should be handled via opts.

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,128 @@
...
+    |> Map.put_new("Action", "ApplyPendingMaintenanceAction")
+    |> Map.put_new("ResourceIdentifier", resource_id)
+    |> Map.put_new("ApplyAction", action)
+    |> Map.put_new("OptInType", opt_in_type)
+    |> Map.put_new("Version", @version)
+
+    request(client, :post, "/", params: query_params)
+  end
+
+  def apply_pending_maintenance(client, resource_id, action, opt_in_type) when action == "db-upgrade" do
```
benwilson512: Why the guard clause here?

benwilson512: I can't tell any difference between this function and the other clause.

benwilson512:
General notes:

- [ ] I would go through each action and make sure that only mandatory parameters are dedicated function arguments, and that the rest are handled via opts.
- [ ] Ensure that opts handle snake case atom keys
- [ ] ALL optional parameters need documented via the typespec in the client module. See other client modules for examples.
- [ ] new lines at the end of all files.

lib/ex_aws/rds/impl.ex
sashaafm: Hey Ben, I'm just used to working with the Map module, I'll change the function to be written with map literals :)

About the opts, I actually had thought about that but was having some trouble converting the atom keys into the proper binary keys. I'll check those utils!

lib/ex_aws/rds/impl.ex
sashaafm: I'm not 100% sure right now but I believe I included the username and password because I kept getting errors saying they were required (I believe there were a couple other methods that also seem to have required parameters listed as optional).

Should I always go by the docs then?

lib/ex_aws/rds/impl.ex
sashaafm: Noted.

lib/ex_aws/rds/impl.ex
benwilson512: No, if you ran into errors without them you made the right choice :)

lib/ex_aws/rds/impl.ex
sashaafm: Yes, they're the same. Probably was late at night ahah :-1: for me

sashaafm:
I'll work on it this week :+1:

sashaafm:
I think everything's done like you asked @benwilson512. The only this missing are some specs for parameters like this one:

**Filters.member.N**

I looked around but I didn't find any other similar specs that you had already defined for the other API's. Maybe it could be something with a range?

sashaafm:
@benwilson512 everything done based on your notes :100:

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,147 @@
...
+      "Action" => "AddTagsToResource",
+      "ResourceName" => resource,
+      "Version" => @version
+    }
+
+    query_params =
+    for {k, v} <- tags do 
+      query_params
+      |> Map.put_new("Tags.member.#{Integer.to_string(n)}.Key", k)
+      |> Map.put_new("Tags.member.#{Integer.to_string(n)}.Value", v)
```
benwilson512: Hey! so this is almost right. The way this works is that suppose I have tags `[foo: "bar", baz: "buz"]` what I want to create is

```
%{
  "Tags.Member.0.Key" => "foo",
  "Tags.Member.0.Value" => "bar",
  "Tags.Member.1.Key" => "baz",
  "Tags.Member.1.Value" => "buz",
}
```

So you don't need them to pass in an `n` value, instead just use Stream.with_index in your for comprehension.

As usual, just use `Map.put` here instead of `put_new`, there's no chance it's already in the map.

benwilson512:
This is a grade A crazy amount of work, thank you so much. The primary step from here is writing some tests. Those should encompass a basic sanity test that actually hits S3 with something innocuous, like `describe_db_instances`, as well as tests which use a pass through client and tests the formatting on any of the more complicated functions.

sashaafm:
Hi @benwilson512. Yes I understand. I'll try to write the tests this week. I'm gonna check out the other tests but I'm thinking about doing something like a real use case. Create an instance, grab it, check if it appears in describe_db_instances, update it, delete it, etc...

lib/ex_aws/rds/impl.ex
```diff
@@ -0,0 +1,156 @@
...
+    |> Map.merge(%{
+      "Action"               => "RebootDBInstance",
+      "DBInstanceIdentifier" => instance_id,
+      "Version"              => @version
+    })
+
+    request(client, :get, "/", params: query_params)
+  end
+
+  @params [:duration, :end_time, [:event_categories_member], [:filter_member], :marker, :max_records, :source_identifier, :source_type, :start_time]
```
benwilson512: Are these module attributes used?

sashaafm: Dumb me. Nevermind I'll delete those.

benwilson512:
Closing in favor of https://github.com/CargoSense/ex_aws/pull/167

PR #136, opened 2016-04-05
optional handling of empty strings automatically by replacing with placeholder text

thehosepipe:
It would be useful to us, and perhaps others, to have an option to automatically handle empty strings. These can occur often, and cleaning them from maps, then adding the keys back in when retrieving is a lot of boilerplate.

This is a suggestion that the config file contains an option. When off ExAws acts as it does now. However when a replacement text is defined in the config file
... During encoding empty strings are replaced with the defined placeholder
... During decoding string containing the placeholder and replaced with empty strings.

Does this code look reasonable? Is there a preferred way to drive the encoding/decoding decision rather than config?

I have added tests to cover the existing cases, but being new to elixir can't think right now of a way to override the config to test replacement works.

What do you think? Open to suggestions.

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -15,7 +15,8 @@ defmodule ExAws.Config.Defaults do
         scheme: "https://",
         host: {"$region", "dynamodb.$region.amazonaws.com"},
         region: "us-east-1",
-        port: 80
+        port: 80,
+        empty_string: nil        
       ],
       lambda: [
         host: {"$region", "lambda.$region.amazonaws.com"},
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -15,8 +15,7 @@ defmodule ExAws.Config.Defaults do
         scheme: "https://",
         host: {"$region", "dynamodb.$region.amazonaws.com"},
         region: "us-east-1",
-        port: 80,
-        empty_string: nil        
+        port: 80
       ],
       lambda: [
         host: {"$region", "lambda.$region.amazonaws.com"},
--- a/lib/ex_aws/dynamo/decoder.ex
+++ b/lib/ex_aws/dynamo/decoder.ex
@@ -29,7 +29,14 @@ defmodule ExAws.Dynamo.Decoder do
   def decode(%{"NULL" => true}),    do: nil
   def decode(%{"NULL" => "true"}),  do: nil
   def decode(%{"B" => value}),      do: value
-  def decode(%{"S" => value}),      do: value
+  def decode(%{"S" => value}) do
+    replacement = ExAws.Dynamo.config_root[:dynamodb][:empty_string] 
+    if (replacement != nil) and (value == replacement) do
+      ""
+    else
+      value
+    end
+  end
   def decode(%{"M" => value}),      do: value |> decode
   def decode(%{"SS" => values}),    do: values
   def decode(%{"BS" => values}),    do: values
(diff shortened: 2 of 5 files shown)


benwilson512:
Hey there! Thanks for giving this a go.

Before we address implementation questions I would like to know more about what you're trying to achieve here. The test cases you have already work. In iex I can already do

```
iex(5)> ExAws.Dynamo.Encoder.encode("val")
%{"S" => "val"}
iex(6)> ExAws.Dynamo.Encoder.encode("")   
%{"S" => ""}
iex(7)> ExAws.Dynamo.Decoder.decode(%{"S" => "val"})
"val"
iex(8)> ExAws.Dynamo.Decoder.decode(%{"S" => ""})   
""
```

Can you show examples of the functionality you're trying to go for?

thehosepipe:
Of course, sorry I wasn't clearer.

Dynamo doesn't allow empty attributes. For strings this is especially painful for us as inputs can reasonably be empty strings.

Right now if I try to push this to dynamo %{"has_value" => "a value", "has_no_val" => ""}  then dynamo will reject it with a validation 'One or more parameter values were invalid: An AttributeValue may not contain an empty string....'

Because of that on saving I need to go through the whole multi-level structure and remove all empty string attributes. And worse on loading I need to go through and add them back so consuming modules can be made less smart. They expect the attributes to exist.

With the change I made, if set the config value :dynamodb, empty_string: "null-string"
then the null string is automatically replaced with a dummy value

iex(14)> ExAws.Dynamo.Encoder.encode(%{"has_value" => "a value", "has_no_val" => ""} )
%{"M" => %{"has_no_val" => %{"S" => "null-string"}, "has_value" => %{"S" => "a value"}}}

and on decoding
iex(14)> ExAws.Dynamo.Decoder.decode(%{"M" => %{"has_no_val" => %{"S" => "null-string"}, "has_value" => %{"S" => "a value"}}})
%{"has_no_val" => "", "has_value" => "a value"}

Does this help?

thehosepipe:
I forgot to say, the reason I haven't added a test for this extra functionality is that, to be honest, I couldn't work out how to override the config setting dynamodb, empty_string without changing the config file in the project, and thereby impacting the whole test suite. I need to be able to set that value for use only during a test of the 'dummy text' feature.

benwilson512:
Right. So there's a couple problems here. The first is that a global config value will not work. ExAws allows users to create custom clients, of which `ExAws.Dynamo` is merely a nice default. Configuration is always scoped to these clients, so we would need to pass that client configuration module through the entire encoding stack in order to support this.

I think this is best solved by using structs. Given this for example:

```
```

thehosepipe:
The struct idea can only be used for a lowest level struct, If I have a nested structure I would have to build a custom decoder that creates structs for every level. 

However more critically it isnt dynamic. It requires the map to be known at design time. In our case it isn't, it can change depending on user inputs. So really I need a generic solution that can handle the empty strings.

thehosepipe:
Thinking a bit further, it would also be a problem on saving. The empty strings are still in the map. These would fail on saving.

Although at least we could replace the implementation of ExAws.Dynamo.Encodable, for: Map to solve this side of the problem

benwilson512:
Right so I looked through the docs again and we have
```
String and Binary type attributes must have lengths greater than zero
```

I think the right call here is to just make sure that empty strings like null are simply stripped out entirely from what's sent.

thehosepipe:
Ok no worries then

thehosepipe:
So I've reduced the scope of the change to one that simply removes empty strings from maps. It seems better that than letting such values go to amazon and raise an exception. Easier for users of the library than stripping them all out beforehand, and less processing since the map is already being manipulated at this stage.

What do you think?

benwilson512:
Perfect! Would you mind squashing your commits at this point? Happy to merge afterward :)

benwilson512:
Turns out I can do that, yay Github :)

thehosepipe:
oh fantastic, didn't even know that could be done. Learn something new every day.

Thanks, appreciate it!

Issue #132, opened 2016-03-31
Support awscli configuration files in '.aws' folder


TattdCodeMonkey:
# Proposal

The `awscli` uses the `~/.aws` folder to store configuration files. One of those being a `credentials` file that can have default and profiles sections. The file follows an ini type format:

```
[default]
aws_access_key_id = <access_key>
aws_secret_access_key = <super_secret_thing>

[bob]
aws_access_key_id = <access_key>
aws_secret_access_key = <super_secret_thing>
```

To support this in `ExAws` we could introduce a new config tuple similar to the `{:system, "KEY"}` that is currently supported. This could support the default or named profiles. Something like the following could be supported:

```elixir
#loads config values from the "default" section
config :ex_aws,
  access_key_id: [{:shared, "aws_access_key_id"}, :instance_role],
  secret_access_key: [{:shared, "aws_secret_access_key"}, :instance_role]
```
and 
```elixir
#loads from the named section
config :ex_aws,
  access_key_id: [{:shared, "bob", "aws_access_key_id"}, :instance_role],
  secret_access_key: [{:shared, "bob", "aws_secret_access_key"}, :instance_role]
```

## Implementation Details

- To read the config file we would add an optional dependency on [ConfigParser](https://hex.pm/packages/configparser_ex) and return nil if the module is not loaded at runtime.
- ~~We would also need to load the path to the file in a cross-platform friendly way. Check for `~/.aws/credentials` on OSX, Linux, Unix and `C:\Users\<USERNAME>\.aws\credentials` on Windows~~
- use `System.user_home |> Path.join(".aws") |> Path.join("credentials")` and check `File.exists?`

## Questions
- What should the atom be for this type? `:shared`, `:awscli` or something else?
- What should the format of the config object be? `.aws` has `credentials` and `config` files. So it could be something like: `{:awscli, <file>, <section>, <key>}` where file and section both have defaults?

Thoughts, comments, concerns?

benwilson512:
Apologies for the delay. My primary concern here is the need to add an extra dependency. I wonder if there's a way to make the dependency nicely optional so that it's only used if you specify `{:shared` or something. I suppose if I simply marked it as an optional dependency, and then had just remote function calls to those modules inside my Config stuff I would be good to go.

TattdCodeMonkey:
The way I tested it, was with an optional dependency. Then I did a
`Code.ensure_loaded?` call before using it. If it's not loaded give a
warning and return nil.

[...]

PR #139, opened 2016-04-13
add :awscli option for config values

TattdCodeMonkey:
# What is this?
This adds :awscli option for config values, this will use ConfigParser
to try and read the config files used by the awscli (Python) app. It
supports loading values from the `default` or named profiles.

# Notes
This is my 1st pass for Issue #132. I would like to make this a bit more testable, moving usage of ConfigParser to load from config similar to `json_codec`. But right now the `File.exists?` would still make testing a bit harder. But without it `ConfigParser.parse_file` would raise an error if the file doesn't exist.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -1,5 +1,5 @@
 defmodule ExAws.Config do
-
+  require Logger
   @moduledoc false
 
   # Generates the configuration for a client.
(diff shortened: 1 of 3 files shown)


benwilson512:
I have some ideas about how to make the code more testable, but one question I had is if we actually want the warnings. The usual procedure for a lot of AWS libraries is that you just specify a list of possible ways to get the access keys and then it just falls back until it's done. I could see maybe debug output, but warnings seem not quite right to me.

We'll also want to make sure this gets cached one way or another because adding file read + parsing time to every requests seems undesirable.

TattdCodeMonkey:
I hadn't thought about caching, but your definitely right. I'll rework this removing the warnings and caching the parsed config.

benwilson512:
Just as a note, I already do some caching around looking up auth from instance roles, whatever caching we do here may already be covered by that mechanism, it's been long enough since I've looked at it that I'm not sure.

TattdCodeMonkey:
I think I can do this as a separate dependency and module. Something like
`awscli_config` that will read the file and set application env variables
on startup. I'll hack on that some tomorrow and if it works I'll close this
PR.

Issue #145, opened 2016-04-29
Kinesis Firehose


cwc:
The Firehose API appears to be separate from the Kinesis API. Is anyone working on an implementation? If not, I'm prepared to take it on myself, and would appreciate any insight, tips, potential gotchas, etc.

cwc:
Got some work done here, would appreciate any comments:
https://github.com/CargoSense/ex_aws/compare/master...cwc:firehose

benwilson512:
I'm impressed, please open a full PR.

benwilson512:
Gonna close this in favor of the PR.

PR #146, opened 2016-04-29
Add Kinesis Firehose client

cwc:
Refs #145 

I'm currently trying to test this to ensure that it works. I'm running the app on my dev machine, rather than within AWS, so it's choking in `ExAws.InstanceMeta.request` while trying to get IAM stuff.

My `AWS_*` environment vars are set correctly, is there something else I need to do?


benwilson512:
That's odd, it should only be trying to hit the instance meta if it can't get the values from the ENV. if you open just a regular `iex` and do

```
System.get_env("AWS_ACCESS_KEY_ID")
System.get_env("AWS_SECRET_ACCESS_KEY")
```

do you see your values?

cwc:
I'm an idiot; wrong shell :grimacing:. It seems to be working!

cwc:
So I wasn't getting any errors, but no data was seen in my delivery streams. I poked around in the official SDK and found that the Firehose service has its own domain. Couldn't find it in the actual documentation so I was assuming it went to the same place as Kinesis, but whatever.

Latest change gets records into Firehose for sure. It might be worth looking into why no errors were emitted when it was posting to the Kinesis endpoint with invalid data.

cwc:
Anything else needed before this can be merged?

benwilson512:
- [ ] Basic sanity test that hits the actual API (preferably a basic list function)
- [ ] unit tests on the individual functions

cwc:
When I send several records, I sometimes see errors logged like:

```
[error] ExAws: HTTP ERROR: :connect_timeout
[error] ExAws: HTTP ERROR: :closed
```

Is ExAws automatically retrying when this happens? Do I need to fix the Firehose client so these errors can be handled in user code?

benwilson512:
ExAws does indeed auotmatically retry, and you do not need to worry about that yourself.

benwilson512:
I'm happy to merge this btw the minute there are some basic tests. Please see some of the other services as examples.

PR #148, opened 2016-05-07
#75 delete_multiple_objects handles more than 1000

thehosepipe:
chunks them automatically, and deletes recursively


thehosepipe:
This should add functionality requested in #75 .
chunks objects into groups of 1000 and deletes each chunk separately. Recursively sends the tail objects to same method.
Method will therefore return up the stack the full AWS error if any failures occur. Thus user has full knowledge of what failed in the event of an error.

benwilson512:
So, setting aside some implementation questions, you'll note that the path determined in #75 was a `delete_all_objects/n` function that would make the necessary calls.

One of the things that I think is generally pretty important is that people know how many AWS requests each function does, because requests are a high latency operation. In all places where ExAws makes additional requests on behalf of the user, we do so from differently named functions like `stream_objects!` that make it clear it isn't an ordinary AWS operation.

So, what we really want is `delete_all_objects`. I would be happy to see a PR that does that. I'll be providing some feedback on this one mostly for educational purposes, but closing it anyway.

Other note: Please rebase off of master so that there aren't merge commits in the PR, it makes for a confusing git history.

I would expect delete_all_objects to look basically like:
```elixir
list_of_more_than_1k_things
|> Stream.chunk(1000, [], [])
|> Enum.map(&S3.delete_multiple_objects("bucket", &1))
```

However, we'll need to think through error handling

lib/ex_aws/s3/impl.ex
```diff
@@ -232,6 +234,14 @@ defmodule ExAws.S3.Impl do
     body_binary = body |> IO.iodata_to_binary
 
     request(client, :post, bucket, "/?delete", body: body_binary, headers: %{"content-md5" => content_md5})
+    |> case do
+      {:ok, body} -> 
+        case tail == [] do
+          true -> {:ok, body}
```
benwilson512: just pattern match on tail IE

```
case tail do
  [] -> {:ok, body}
  _ -> etc
end
```

lib/ex_aws/s3/impl.ex
```diff
@@ -232,6 +234,14 @@ defmodule ExAws.S3.Impl do
...
+          false -> delete_multiple_objects(client, bucket, tail, opts)
```
benwilson512: not a big fan that we're just throwing out results here.

thehosepipe: regarding this, would we really want to pass back a list of results where each result is ok? I was thinking rather that the happy path result should be either {:ok, _} or {:ok, number_deleted}

When a failure occurs we of course have to return the failure reason. I guess we could return the number successfully deleted, as well as the error for the failure... I am a little wary of returning the list of items deleted and the error as returned data load gets large. But what do you think?

thehosepipe: so options
Happy path
{:ok, number deleted} or {:ok, enumerable of items deleted}

error path
{:error, number deleted, error message} or  {:error, enumerable of items deleted, error message}

PR #152, opened 2016-05-18
Allow ExAws.S3.Client.list_objects to handle unowned objects

nickmabry:
Hi! Thanks for the awesome library! I tried out `ExAws` for the first time tonight and ran into a small issue when attempting to list the objects in one of our companies' S3 buckets. I had been able to list buckets without issue using `ExAws.S3.list_buckets/0`, but the following failed:

``` elixir
iex> ExAws.list_objects("my-bucket-name")
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil
stacktrace:
  (elixir) lib/enum.ex:1: Enumerable.impl_for!/1
  (elixir) lib/enum.ex:116: Enumerable.reduce/3
  (elixir) lib/enum.ex:1486: Enum.reduce/3
  (elixir) lib/enum.ex:1025: Enum.into/4
  (ex_aws) lib/ex_aws/s3/parsers.ex:41: anonymous fn/1 in ExAws.S3.Parsers.list_objects_binaries/1
  (elixir) lib/enum.ex:1019: anonymous fn/4 in Enum.into/3
  (stdlib) lists.erl:1262: :lists.foldl/3
  (elixir) lib/enum.ex:1025: Enum.into/4
```

The root cause seemed to be that this bucket was not returning ownership information for its objects, and after the response XML was parsed, `ExAws.S3.Parsers.list_objects_binaries({:owner, nil})` would eventually be called, and fail when it tried to recurse with `nil`.

My quick solution was to reorder the anonymous function clauses to handle any nil values before matching specific keys (other than `:contents`, which should always be present.) It doesn't appear to break anything, but I'm not super-familiar with the library. If it needs some changes, I'm happy to take another stab. If this is a weird use case and it shouldn't be merged, I understand.

### DIFF
--- a/lib/ex_aws/s3/parsers.ex
+++ b/lib/ex_aws/s3/parsers.ex
@@ -38,9 +38,9 @@ if Code.ensure_loaded?(SweetXml) do
     defp list_objects_binaries(result) do
       Enum.into(result, %{}, fn
         {:contents, contents} -> {:contents, Enum.map(contents, &list_objects_binaries/1)}
+        {k, nil} -> {k, nil}
         {:owner, v} -> {:owner, list_objects_binaries(v)}
         {:common_prefixes, prefixes} -> {:common_prefixes, Enum.map(prefixes, &list_objects_binaries/1)}
-        {k, nil} -> {k, nil}
         {k, v} -> {k, IO.chardata_to_string(v)}
       end)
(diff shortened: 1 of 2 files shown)


benwilson512:
Perfect thanks!

Issue #90, opened 2015-10-20
Can't configure ExAws to use Dynamodb Local anymore


jordi-chacon:
Hi Ben,

I was using Dynamodb Local via ExAws just fine until this commit:https://github.com/CargoSense/ex_aws/commit/d68ae5f2c0cfa2facc804c8951759e55eb9fb378

I have `host: "localhost"` which seems to cause the following crash:
```
    ** (EXIT) an exception was raised:
        ** (ArgumentError) argument error
            (stdlib) binary.erl:317: :binary.replace/4
            (ex_aws) lib/ex_aws/config.ex:67: ExAws.Config.parse_host_for_region/1
            (ex_aws) lib/ex_aws/config.ex:16: ExAws.Config.build/2
```

Is it still possible to use DynamoDB Local via ExAws? If so, how? If not, could you consider adding that option back?

Thanks!

benwilson512:
Ah yea that's odd because that's test suite uses dynamodb local and it passes without issue. Let me see if I can figure out what the issue is one sec.

benwilson512:
Can you post the relevant parts of your config.exs and dev.exs files please?

jordi-chacon:
Here you go:

```
config :ex_aws,
  access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :instance_role],
  secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, :instance_role],
  http_client: HTTPoison

config :ex_aws, :dynamodb,
  scheme: {:system, "DYNAMODB_SCHEME"},
  host: {:system, "DYNAMODB_HOST"},
  port: {:system, "DYNAMODB_PORT"},
  region: "eu-west-1"
```

The error happens when trying to start my application when running `mix test` like this:
```
DYNAMODB_SCHEME="http://" \
  DYNAMODB_HOST="localhost" \
  DYNAMODB_PORT="7000" \
  mix test --include external:true
```

benwilson512:
Ah yeah I see the issue, this is my bad. I can do a fix, but for the moment a workaround would be to do

```
```

benwilson512:
I would make sure not to alter the host key at all in your prod.exs file.

benwilson512:
Can you alter your ex_aws dependency to read
```
 {:ex_aws, github: "cargosense/ex_aws", branch: "bug-fixes"}
```
as well as return your config to how it was and let me know if that fixes it?

I'll have a release out shortly if so.

[...]

PR #153, opened 2016-05-19
Be able to use ENV variables to define host parameter

sylverlion:
Following the discussion in https://github.com/CargoSense/ex_aws/issues/90

This PR removes the `{:host, host}, config ->` clause from `retrieve_runtime_config/2` to be able to set the host value within an ENV variable and be able to use `host: {:system, "AWS_HOST"}` in the `config*.exs`

I don't see any reason to keep this clause, but I just started using your project, so maybe this PR is a bit too rash.
At least it opens the discussion on this matter, since i'm stuck if i don't do this modification, and i'm obviously not fond to keep working with a fork to be able to use it.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -42,8 +42,6 @@ defmodule ExAws.Config do
   def retrieve_runtime_config(%{config: config} = client) do
     new_config = config
     |> Enum.reduce(%{}, fn
-      {:host, host}, config ->
-        Map.put(config, :host, host)
       {k, v}, config ->
         case retrieve_runtime_value(v, client) do
           %{} = result -> Map.merge(config, result)


benwilson512:
Hey!

I apologize for not responding on the other thread, the issue was already closed and so I simply missed that there was an update to it. I _think_ that this PR is fine, but it's been long enough that I don't really remember why I had a special case for `{:host`. I need to test this against other regions some locally first. Host can be a finicky thing due to amazon's various conventions around host names for services.

sylverlion:
Hey Ben,

Thanks for your answer. 
Sure, I'm not able yet to see all the implications here and you should have had a good reason to put this special case here i suppose.
One way or the other, if we could be able to use ENV parameters for the host, that would be great !

Thanks

benwilson512:
This will be fine, thank you!

benwilson512:
So, this didn't actually work, but I'll be able to push a fix to master here in a bit that'll contain the desired functionality.

benwilson512:
Fixed in master.

PR #154, opened 2016-05-20
Add support for poison 2.0

steffenix:
Tested using poison 2.0 didn't raise any test.


badoet:
would love to have this. the 1.2 seems few versions too far from the head.
at least have that or options. to relax the dependencies.
Thank you for this!

benwilson512:
Thanks! :)

PR #161, opened 2016-05-24
Add Qualifier to Lambda

sgtpepper43:
Add a "qualifier" option to lambda functions. Gets set in opts, but is not added to headers (it's part of the query string).

### DIFF
--- a/lib/ex_aws/lambda/impl.ex
+++ b/lib/ex_aws/lambda/impl.ex
@@ -84,7 +84,7 @@ defmodule ExAws.Lambda.Impl do
   end
 
   def invoke(client, function_name, payload, client_context, opts \\ []) do
-    opts = Enum.into(opts, %{})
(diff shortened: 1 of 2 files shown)


lib/ex_aws/lambda/impl.ex
```diff
@@ -101,7 +101,9 @@ defmodule ExAws.Lambda.Impl do
         header = {"X-Amz-Client-Context", context |> client.config[:json_codec].encode! |> Base.encode64}
         [header | headers]
     end
-    request(client, :invoke, payload, "/2015-03-31/functions/#{function_name}/invocations", [], headers)
+    url = "/2015-03-31/functions/#{function_name}/invocations"
+    url = if qualifier != nil, do: url <> "?Qualifier=#{qualifier}", else: url
```
benwilson512: you can simply do `if qualifier, do: ` here as `nil` evaluates to false.

benwilson512:
Thank you!

PR #168, opened 2016-06-06
Remove virtual-style hosting

bradgessler:
When I setup a docker container running fakes3 at `http://s3`, the virtual-host style S3 buckets try to access `http://#{bucke}.s3` which results in an `[error] ExAws: HTTP ERROR: :nxdomain`. This removes virtual-hosted-style URLs in favor of the more consistent bucket path, which makes standing up fakes3 in a docker-compose cluster easier.


benwilson512:
This fails on any operations like `list_buckets` because it results in a double slash. However, I will be making these changes in the new beta branch.

PR #171, opened 2016-06-13
Add HTTPotion 3.0.0 as it now compiles on Elixir 1.3

cbarratt:
I noticed how it was recently removed but the author has now released a working version that is compatible with Elixir 1.3 👍


benwilson512:
This is basically identical to #172

PR #172, opened 2016-06-13
Remove unused imports

cbarratt:
There are some more warnings but I'm struggling to understand where or how to fix them. There's quite a bit of whitespace removal here too! (My editor does this automatically).

screen shot 2016-06-13 at 11 24 24


benwilson512:
Hey, I realize this is an annoyance for 1.3. However, https://github.com/CargoSense/ex_aws/pull/165 handles this I think in the best way, which is to remove the metaprogramming approach entirely. I hope to release this as a beta here in the next couple days, and will continue pushing further updates there.

PR #177, opened 2016-06-29
Fix 'could not find operation' error on SNS.

milkcocoa:
When I tried to use `SNS.publish/2` with the latest version, I got the following error.

```
{:error, {:http_error, 400, "<ErrorResponse xmlns=\"http://webservices.amazon.com/AWSFault/2005-15-09\">\n  <Error>\n    <Type>Sender</Type>\n    <Code>InvalidAction</Code>\n    <Message>Could not find operation publish for version 2010-03-31</Message>\n  </Error>\n  <RequestId>xxxx</RequestId>\n</ErrorResponse>\n"}}
```

So, I've fixed this issue, and added some brief tests.

### DIFF
--- a/lib/ex_aws/sns.ex
+++ b/lib/ex_aws/sns.ex
@@ -78,19 +78,19 @@ defmodule ExAws.SNS do
     |> Map.new
     |> camelize_keys
 
-    request(:list_topics, opts)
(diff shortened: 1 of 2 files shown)


benwilson512:
Thanks!

PR #181, opened 2016-07-04
Fix syntax error in sample IAM policy

sandisk:
"This policy contains the following JSON error on line 11: Unexpected ']'"


benwilson512:
Thanks!

PR #185, opened 2016-07-21
make retry backoff exponential with jitter

adamkittelson:
I implemented exponential backoff with jitter based on the "full jitter" variant of the formula described at https://www.awsarchitectureblog.com/2015/03/backoff.html

The article didn't specify values for the @max_backoff_in_ms or @base_backoff_in_ms values. I picked 10 seconds for the max because that is how long the longest retry would be for the 10th attempt currently in master, and I chose 10ms for the base because that value resulted in a cap of around 10 seconds on the 10th attempt. Feel free to fiddle with those or with the formula.

sleep times between requests with this formula and base / max values are as follows:

attempt # | between 0 and x ms
----------|----------------------
1 | 20
2 | 40
3 | 80
4 | 160
5 | 320
6 | 640
7 | 1280
8 | 2560
9 | 5120
10 | 10000

Thanks!
Adam

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -1,6 +1,8 @@
 defmodule ExAws.Request do
   require Logger
   @max_attempts 10
+  @base_backoff_in_ms 10
+  @max_backoff_in_ms 10_000
 
   @moduledoc """
   Makes requests to AWS.
@@ -84,9 +86,11 @@ defmodule ExAws.Request do
     {:attempt, attempt + 1}
   end
 
-  # TODO: make exponential
-  # TODO: add jitter
   def backoff(attempt) do
-    :timer.sleep(attempt * 1000)
+    (@base_backoff_in_ms * :math.pow(2, attempt))
+    |> min(@max_backoff_in_ms)
+    |> trunc
+    |> :rand.uniform
+    |> :timer.sleep
   end
 end


benwilson512:
Hey thanks! This is definitely a valuable addition!

lib/ex_aws/request.ex
```diff
@@ -84,9 +86,11 @@ defmodule ExAws.Request do
...
   end
 
-  # TODO: make exponential
-  # TODO: add jitter
   def backoff(attempt) do
-    :timer.sleep(attempt * 1000)
+    (@base_backoff_in_ms * :math.pow(2, attempt))
+    |> min(@max_backoff_in_ms)
+    |> trunc
+    |> :rand.uniform
```
benwilson512: So, I think the issue here is that if you do `:rand.uniform(10000)` it's going to be _any_ number between 0 and 10000. We definitely want jitter, but jitter is supposed to be some level of fuzz around an otherwise increasing value.

We need to do some thing like:

```
base_value =
  (@base_backoff_in_ms * :math.pow(2, attempt))
  |> min(@max_backoff_in_ms)
  |> trunc

:timer.sleep(base_value + trunc(base_value / 10))
```

This gives us a 10% random jitter on each sleep value, while still guaranteeing a general curve.

adamkittelson: It's a little unintuitive, but based on the article I believe that's on purpose.

They have an "equal" jitter variant which is closer to what you're describing, they describe it like "The intuition behind this one is that it prevents very short sleeps, always keeping some of the slow down from the backoff."

They also describe a 3rd variant they call "decorated" jitter you could look into.

At any rate, they saw better results with the "full" jitter approach though vs the "equal" jitter approach. (And even the equal jitter approach had more jitter than your example, from around 50% to 150% of the base value).

"Of the jittered approaches, “Equal Jitter” is the loser. It does slightly more work than “Full Jitter”, and takes much longer. The decision between “Decorrelated Jitter” and “Full Jitter” is less clear. The “Full Jitter” approach uses less work, but slightly more time. Both approaches, though, present a substantial decrease in client work and server load."

Like I said though, feel free to play with it, but I'm not sure that guaranteeing a general curve actually matters.

benwilson512: Hm interesting, I do actually remember reading that article now when it came out. I suppose the reason I was leaning toward "equal" jitter was because the scenario I'm using fallback for is to compensate for network irregularities instead of throughput. The general assumption is that the AWS service being hit is more than capable of handling whatever load our system wants to throw at them. Thus if we're hitting issues, it's some kind of intermittent network issue, and with those it is important to wait out the issue (within bounds).

That said, I have no empirical evidence that equal jitter is better at avoiding network issues. I suppose there's also an argument to be made in favor of full jitter here in such cases, because by minimizing the probability that a group of retries all happen at once, we minimize local network load from the box.

Long story short, I think your approach here with full jitter is likely correct.

adamkittelson: How do you feel about the possibility of me amending to the PR to make some of these values configurable at the service / request level (e.g. the base/max sleep time, retry count, and type of jitter)?

It looks like the aws java sdk allows some configuration of some things related to retries, and as an end user I could see wanting to handle retries differently in different situations. e.g. if I have a service with an API call that does an insert into SQS or Dynamo I may only want to retry 5 times over the course of 2-3 seconds before passing the backpressure amazon is giving me back to my client in the form of a 503 or something so they can engage their retry logic of my service, but in another scenarios, like some background process doing something with s3 files maybe i'm fine with retrying more times at potentially much longer intervals that would possibly be so long they'd cause something like nginx to declare a gateway timeout if the api service had those values.

benwilson512: That sounds good as well. This should be much simpler to do with the new architecture.

adamkittelson: Cool, I'll try to find some time this weekend to do another PR with some configurability. Thanks Ben!

PR #188, opened 2016-07-25
Drop dependency on httpoison, use hackney directly

michalmuskala:
This completely removes `HTTPoison` including the adapter. It might be desired to leave the adapter anyway, but I don't see the advantage especially since it's coming to a bigger release so some breaking may be fine.


benwilson512:
Heh I'll run stuff locally, I'm very 👍  on this so no worries :)

benwilson512:
Tests are green :)

sashaafm:
@michalmuskala would you mind commenting on the benefits of using hackney directly? Less overhead and possible points of failure from httpoison?

michalmuskala:
If you can avoid a dependency you should. It's yet another thing that can go wrong. 

Some dependencies are worth it, but others really aren't. I don't think httpoison adds anything significant over hackney in the most use cases. I made similar PRs to 6 projects - in most of them the whole change was basically limited to changing `HTTPoison.post` to `:hackney.post`.

sashaafm:
Yeah that's what I was thinking. Might do the same in a couple of my projects. Thanks @michalmuskala ! :+1:

PR #189, opened 2016-07-26
Support for message attributes in receive_messages

picharras:
Now you can fetch all message attributes in message

### DIFF
--- a/lib/ex_aws/sqs.ex
+++ b/lib/ex_aws/sqs.ex
@@ -183,7 +183,7 @@ defmodule ExAws.SQS do
     {:max_number_of_messages, 1..10} |
     {:visibility_timeout, 0..43200} |
     {:wait_time_seconds, 0..20} |
-    {:meessage_attribute_names, []}
+    {:message_attribute_names, []}
   ]
 
   @doc "Read messages from a SQS Queue"
@@ -199,7 +199,7 @@ defmodule ExAws.SQS do
     params =
       attrs
       |> format_queue_attributes
-      |> Map.put("MessageAttributeNames", format_message_attributes(message_attr))
+      |> Map.merge(format_message_attributes(message_attr))
       |> Map.merge(format_regular_opts(opts))
 
     request(queue, "ReceiveMessage", params)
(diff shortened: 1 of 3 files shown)


benwilson512:
Hey! I'm happy to look at this in greater detail tomorrow, in the mean time can you rebase and clear up the merge conflicts? Best I can tell it's just issues with the mix.lock file.

Thanks!

lib/ex_aws/sqs.ex
```diff
@@ -307,6 +312,13 @@ defmodule ExAws.SQS do
     |> ExAws.Utils.camelize
   end
 
+  defp format_message_attributes(attributes) do
+    attributes
+    |> Enum.with_index
+    |> Enum.map(&format_message_attribute/1)
+    |> Enum.reduce(%{}, &Map.merge(&1, &2))
```
benwilson512: This really should just be a Map.new or Enum.into instead of a map with a reduce via merge. IE
```
attributes
|> Enum.with_index(1)
|> Map.new(fn {attr, i} ->
  {"MessageAttributeNames.#{i}", to_string(attr)}
end)
```

benwilson512: This removes the need for `defp format_message_attribute`.

benwilson512: Do note I use `to_string` instead of `Atom.to_string`. I think requiring atoms is unnecessary, people may want  to hook this up to dynamic sources where using atoms would be the wrong approach.

lib/ex_aws/sqs.ex
```diff
@@ -182,7 +182,8 @@ defmodule ExAws.SQS do
     {:attribute_names, :all | [sqs_message_attribute_name, ...]} |
     {:max_number_of_messages, 1..10} |
     {:visibility_timeout, 0..43200} |
-    {:wait_time_seconds, 0..20}
+    {:wait_time_seconds, 0..20} |
+    {:message_attribute_names, []}
```
benwilson512: This should be `{:message_attribute_names, [String.Chars.t]}`

benwilson512:
Closed in favor of https://github.com/CargoSense/ex_aws/pull/240

Issue #193, opened 2016-07-29
HTTPoison dependency failling on master


eniolopes:
I'm having the following error pointing to HEAD on master

```
(UndefinedFunctionError) function ExAws.Request.HTTPoison.request/4 is undefined (module ExAws.Request.HTTPoison is not available)
```

I believe the source is https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/config/defaults.ex#L10, shouldn't it be using hackney instead?

benwilson512:
It should, that's a bug, PR?

Wonder how tests passed...

eniolopes:
Sure, I can do a PR, what should the default value be? `ExAws.Request.HttpClient` ?

eniolopes:
Ops, I meant `ExAws.Request.Hackne`

benwilson512:
Yup that should do it.

benwilson512:
Fixed by #195

PR #195, opened 2016-08-01
Fix default http client reference

eniolopes:


eniolopes:
Fix https://github.com/CargoSense/ex_aws/issues/193

benwilson512:
Thanks!

PR #196, opened 2016-08-02
concurrent downloader with gen stage

benwilson512:


lib/ex_aws/s3/download.ex
```diff
@@ -0,0 +1,48 @@
...
+    {:ok, sink} = Download.Sink.start_link(op.dest)
+
+    for _ <- 1..Keyword.get(op.opts, :max_concurrency, 8) do
+      {:ok, worker} = Download.Worker.start_link(%{bucket: op.bucket, path: op.path, config: config})
+
+      GenStage.sync_subscribe(sink, to: worker, min_demand: 0, max_demand: 1)
+      GenStage.sync_subscribe(worker, to: source, min_demand: 0, max_demand: 1)
+    end
+
+    # how do I block till this is done?
```
benwilson512: @josevalim This worked out super well, and using GenStage really simplified stuff. The Source just emits byte delimiters that need to be downloaded, workers use those to actually pull down the data, and then the chunks are folded back in to process which handles writing them out to file.

What I'm not clear on is how I might block until it's done. I looked into just returning `{:stop, ...etc}` from the producer when it reached the end of the byte chunks but that seemed to make all the subscribers and definite sink quit prior to actually downloading / writing all the chunks.

benwilson512:
@ericmj you may find this interesting as well given our discussions on the topic.

lib/ex_aws/s3/download.ex
```diff
@@ -0,0 +1,48 @@
...
+  def perform(op, config) do
+    {:ok, source} =
+      op.bucket
+      |> get_file_size(op.path, config)
+      |> Download.Source.start_link
+
+    {:ok, sink} = Download.Sink.start_link(op.dest)
+
+    for _ <- 1..Keyword.get(op.opts, :max_concurrency, 8) do
+      {:ok, worker} = Download.Worker.start_link(%{bucket: op.bucket, path: op.path, config: config})
```
josevalim: You can return `subscribe_to: [{source, min_demand: 0, max_demand: 1}]` as the third option in `init/1` so you don't need to subscribe with another request later.

lib/ex_aws/s3/download.ex
josevalim: I am copying @fishcakez in case we want to improve this in GenStage.

The default subscription is permanent which means that, if the producer terminates, the subscriber will terminate immediately (even if reason is normal). Maybe we should have a mechanism for sync termination throughout the whole pipeline (i.e. the producer only terminate after consumers are done). One option you can do for now is to set the subscription mode to temporary (instead of permanent) and monitor the producer and than exit accordingly depending if the producer exited with reason `:normal` or not when you are done processing. Another option is to use the notifications system in GenStage. When the producer is done, you issue a GenStage.async_notify with a message that will be sent to all consumers. Once the consumers receive those messages, they terminate. The producer then watches its consumers (using handle_subscribe and handle_cancel callbacks) and exit when they are all done.

Both approaches currently have race conditions. If you return `{:stop, :normal, ...}` even after monitoring, you can exit before all messages are delivered (in case the dispatcher is buffering messages). There is no approach to solve this.

The second approach has race conditions in case the producer is fast and send the notification before all consumers are subscribed. We have mechanism to solve the second currently:

    1. The producer sets `demand: :accumulate` on `init/1`
    2. Once all consumers are subscribed, we set demand to `:forward` using `GenStage.demand/2`
    3. Once we have no more items, we emit a notification
    4. The producer then wait for all consumers to terminate so it can exit

The questions I have are:

  1. Should we make the first approach (the one using monitoring) simpler by either guaranteeing all queued messages are delivered before `:stop` (it may never happen) or by providing a `flush_and_stop` function?

  2. Can we automate any of the steps on the second approach? I am certain the fourth step (where the producer waits for all consumers and then exits) can be automated by supporting a `consumers: :permanent` option. That's what we do in GenStage.Flow and GenStage.Streamer.

Feedback is appreciated.

lib/ex_aws/s3/download/sink.ex
```diff
@@ -3,17 +3,42 @@ alias Experimental.GenStage
...
+  def handle_events(chunks, _from, state) do
+    :ok = :file.pwrite(state.file, chunks)
+
+    state = %{state | bytes_written: state.bytes_written + total_bytes(chunks)}
+
+    if done?(state) do
+      {:stop, :normal, state}
+    else
+      {:noreply, [], state}
+    end
```
benwilson512: @josevalim I think in my particular use case I can get around needing to answer the questions you raised by tracking the bytes written in the sink, and have it choose to stop when it's written `file_size` number of bytes.

My mistake earlier was using `:stop` from the source to determine completion, but it's really the sink that determines that in this case.

lib/ex_aws/s3/download.ex
```diff
@@ -0,0 +1,51 @@
...
+
+      GenStage.sync_subscribe(sink, to: worker, min_demand: 0, max_demand: 1)
+      GenStage.sync_subscribe(worker, to: source, min_demand: 0, max_demand: 1)
+    end
+
+    receive do
+      {:DOWN, ^ref, :process, ^sink, :normal} ->
+        {:ok, :done}
+    end
```
benwilson512: Do I need to worry about other `{:DOWN` messages? It's linked to all the processes so if they die this one will die as well.

lib/ex_aws/s3/download/sink.ex
josevalim: @benwilson512 the issue is that the producer won't terminate if the consumers exit. If you are exiting with normal reason, no link will break. So you will either need to have the producers check when all consumers are dead (handle_cancel will suffice) or use a non-normal exit or call GenStage.stop.

benwilson512: Gotcha. `GenStage.stop` inside the receive block seems easiest.

PR #198, opened 2016-08-02
initial pass on concurrent multipart upload

benwilson512:


lib/ex_aws/s3/upload.ex
```diff
@@ -0,0 +1,64 @@
...
+
+  def perform(op, config) do
+    upload_id = start_upload(op, config)
+    stream = build_stream(op)
+
+    Flow.new(stages: op.opts[:max_concurrency] || 8, max_demand: 1)
+    |> Flow.from_enumerable(stream)
+    |> Flow.map(&upload_chunk(&1, upload_id, op, config))
+    |> Enum.to_list
+    |> complete_upload(upload_id, op, config)
```
benwilson512: @josevalim the cool bit

benwilson512: Memory usage seems a bit high:
image

How much pre-fetching is it doing?

benwilson512: Side note: I feel like the `:stages` option would be better named `:workers`. If I say a process has 8 stages I tend to think of those as 8 things that a given item will go through, when instead we mean that items will be spread across those 8 things.

PR #200, opened 2016-08-17
Add application platform support to SNS

jeremyboles:
SNS has a couple of endpoints so you can use it to send push notifications to mobile devices and this PR adds support for that. I had trouble running the full test suite but `mix test test/lib/ex_aws/sns_test.exs` runs fine with no errors.


benwilson512:
Hey there sorry I'm just noticing this, our slack integrations got messed up and I missed out on some PRs. Thanks!

PR #202, opened 2016-08-19
Add support for additional query parameters to S3.presigned_url

boydm:
Added an optional :query_params field to presigned_url_opts

If this is set to an array of key value pairs [{"key", "value}], then those are added to the url to be signed and included in the calculated signature. This is required to enable uploading parts of a multipart upload directly from the browser to S3.

Added tests to s3_test.exs and auth_test.exs and explained it in the docs for presigned_url.

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -24,10 +24,10 @@ defmodule ExAws.Auth do
     [{"Authorization", auth_header} | headers ]
   end
 
-  def presigned_url(http_method, url, service, datetime, config, expires) do
+  def presigned_url(http_method, url, service, datetime, config, expires, query_params \\ []) do
     service = service_name(service)
     headers = presigned_url_headers(url)
-    query = presigned_url_query(service, datetime, config, expires)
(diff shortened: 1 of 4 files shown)


benwilson512:
This is an exemplary PR, thank you! I'll review further in the morning. Have you tried this against an actual S3 bucket in the wild?

boydm:
Yes. It works against a S3 bucket in the wild. I did have to mess around to get the AWS-side CORS settings right. Thinking I'll write that up in a separate blog post or gist or something.

benwilson512:
I've been thinking about adding some pages to the hex docs, may be worth including there as well whenever you get a chance to write it up.

boydm:
I won't get around to writing it up until at least late next week. (need to work on a talk for ElixirConf...)

boydm:
Just checking...

I assume you're not blocked on me writing up a blog post. I'm planning to do that after the PR gets merged. Maybe even after it's published to hex so that readers have a clean way to try it out.

Good assumption?

benwilson512:
Correct.

Issue #209, opened 2016-09-01
Dynamo.batch_write_item generates invalid DeleteItem syntax


kyleaa:
Using Dynamo.batch_write_item to attempt to delete items generates a request item in the format `%{"DeleteRequest" => %{"Key" => %{"S" => "api_key1"}}}`. This errors with the following message:
```
{:error, {:http_error, 400, "{\"__type\":\"com.amazon.coral.service#SerializationException\",\"Message\":\"Expected null\"}"}}
```

Per the [API docs ](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) this should be formatted as 
`%{"DeleteRequest" => %{"Key" => %{ "name_of_key" => %{"S" => "api_key1"}}}}`. We also need to be able to supply all key values in the event of a composite primary key.

benwilson512:
Fixed in master.

PR #210, opened 2016-09-01
Set key name for DeleteRequest on batch_write_item

kyleaa:
This addresses issue #209.


benwilson512:
Awesome! Can you have a test case that validates this against dynamodb local? Does the version of dynamodb local specified in travis need to be updated?

kyleaa:
I've added an integration test that passes against dynamodb local using the "latest" version specified in `travis.yml`

benwilson512:
Thank you!!!

PR #212, opened 2016-09-05
improve jsx.ex

talentdeficit:
return proper error tuples from `encode/1` and `decode/1` and take advantage of jsx support for maps


benwilson512:
Does this require bumping the version for jsx on the optional dependencies list?

talentdeficit:
no, altho updating to 2.7 or 2.8 isn't a bad idea. 2.7 has a couple bugfixes

benwilson512:
Thank you!!!

PR #214, opened 2016-09-09
Extracted Auth.Signatures into separate module

aaronrenner:
I am working on adding S3 presigned post functionality (#114) and found it a little confusing how the actual signature was being generated. I extracted an `ExAws.Auth.Signatures` module to make this more clear, and so I can share this functionality with the PresignedPostGenerator module I'm working on.

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -1,6 +1,8 @@
 defmodule ExAws.Auth do
   import ExAws.Auth.Utils
 
+  alias ExAws.Auth.Signatures
+
   @moduledoc false
 
   def headers(http_method, url, service, config, headers, body) do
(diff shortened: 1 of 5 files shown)


aaronrenner:
I just realized I used a `describe` block that will only work in Elixir 1.3. What is your minimum elixir version for this library? Travis only tests on 1.3, but the mix file says it supports down to 1.0.

aaronrenner:
I also realized I needed credential generation in an easily accessible place, so I extracted another module.

benwilson512:
Fortunately test files are not included in a hex package so that shouldn't be an impediment. I'm pretty sure I'm using `with` / `else` in there somewhere too so I should probably update that.

This looks pretty good, will review more tomorrow, thanks!

aaronrenner:
I just rebased these on top of your latest changes, so you shouldn't need to do a merge commit. Please let me know if you have any feedback. Thanks for all your work on this library!

benwilson512:
Looks good, thanks!

PR #216, opened 2016-09-13
Implement ExAws.Operation, for ExAws.Operation.RestQuery

kanmo:
I want to get the result of EC2-API running such code.

```
ExAws.EC2.describe_instances
|> ExAws.request
```

so, implementation of the ExAws.Operation Protocol to ExAwsOperation.RestQuery.


benwilson512:
Can you include some tests for an API that uses a rest query?

benwilson512:
Thanks!!!

PR #217, opened 2016-09-14
Update Credentials functions arg order to match Signatures arg order

aaronrenner:
I just realized right after you committed #214, the argument order of functions in the `Credentials` module don't match the argument order in the `Signatures` module.

Here are how they are currently

```elixir
Credentials.generate_credential_scope_v4(service, datetime, config)
Signatures.generate_signature_v4(service, config, datetime, string_to_sign)
```
This change updates the argument order of functions in the `Credentials` module to match the argument order of functions in the `Signatures` module.

```elixir
Credentials.generate_credential_scope_v4(service, config, datetime)
Signatures.generate_signature_v4(service, config, datetime, string_to_sign)
```

Sorry about this oversight.


aaronrenner:
Thanks for merging!

PR #219, opened 2016-09-19
Adds Flow support to S3.Upload

aaronrenner:
Currently, you can only upload streams using `S3.upload`. This change adds the ability to upload flows as well.

I've also added some some integration tests that ensure the S3 upload operations send the expected http requests. I put those tests in `test/lib/ex_aws/s3/upload_test.exs`, but can move them if you'd like.

Please let me know what you think and thanks for the work you've already done to support GenStage.


benwilson512:
I'm not entirely convinced this is necessary. If you're gonna do your own flow you're missing basically 2 lines until you've done the whole thing anyway. I don't want to have people needing to worry about the part where you gotta wrap it in index either.

aaronrenner:
I'm trying to do a streaming file transfer with hackney + ex_aws, and hackney will only allow the process that owns the connection (aka `client_ref`) to read the body of the response. That means I need to be able to transfer ownership of the connection to the process that will be reading the body before I start the upload Flow. If the upload Flow is built using `Flow.from_enumerable\2`, there is no way to grab the PID of the producer stage and change the controlling process before the flow starts.

Here's some code:

```elixir
def transfer_file(source_url) do
  with {:ok, download_stream} <- start_request(source_url),
          :ok <- stream_upload(download_stream), 
          do: :ok
end

defp start_request(source_url)
  {:ok, status, _headers, client_ref} = :hackney.request("get", source_url)

  case status do
    200 ->
      download_stream = hackney_response_to_stream(client_ref)
      {:ok, download_stream}
    _ ->
      :error
  end
end

defp stream_upload(download_stream) do
  stream_for_ex_aws =Stream.with_index(download_stream, 1)

    # I know GenStage.Streamer isn't part of the Public API, but it's exactly what I needed
    # and figured it's better than writing my own code to do the same thing.
    {:ok, download_stage} = GenStage.start_link(GenStage.Streamer, {stream_for_ex_aws, [consumers: :permanent]})

    # This is the key part where I transfer the controlling process so the 
    # body can be read from the Flow producer.
    :hackney.controlling_process(client_ref, download_stage)

    Flow.new(stages: 2, max_demand: 1)
    |> Flow.from_stage(download_stage)
    |> ExAws.S3.upload(bucket_name(), filename())
    |> ExAws.request!
end

defp hackney_response_to_stream(client_ref) do
  # ...
end
```
  
I absolutely agree that it's not that great to require the producer emit a tuple of `{binary, index}`, but once the flow is running and things have become concurrent, I'm not sure how to add that index in for the user.

I also thought about having the stream make the http request as part of the stream's setup function, but I only wanted to read the body and stream the upload if I received a 200 status code.

I did make sure the current api that allows the user to pass a plain old stream still works and I'm guessing 90% of the users will use it this way.  However, if you also allow a Flow to be passed in (even if the format is weird), it will still allow it to be extensible for users that need something custom like this. If it's hard coded to use `Flow.from_enumerable/2` it doesn't appear you can use a GenStage-based producer because you can't convert a Flow to a Stream and back to a Flow for uploading due to limitations noted in the `GenStage.stream` documentation.

> Known limitations

>from_enumerable
>This module also provides a function called from_enumerable/2 which receives an enumerable (like a stream) and creates a stage that emits data from the enumerable.

>Given both GenStage.from_enumerable/2 and GenStage.stream/1 require the process inbox to send and receive messages, it is impossible to run a stream/1 inside a from_enumerable/2 as the stream/1 will never receive the messages it expects.

I really appreciate you looking this over and giving me some push-back when I didn't explain my situation as well as I could have. Please let me know what you think, and if there is a different way I should be approaching this problem.

benwilson512:
Can you elaborate on the use case here? I'm not entirely sure what you're trying to do with the hackney stream.

aaronrenner:
I'm writing an otp app that downloads a file from a url and uploads it to S3. Since the file from the url could be very large, I didn't want to read it from memory, or write it to disk. Instead I'll use hackney to read the response a chunk at a time and concurrently upload to S3. 

I make the initial request with

```
{:ok, status, _headers, client_ref} = :hackney.request("get", source_url)
```
and then I read it a chunk at a time with

```
:hackney.stream_body(client_ref)
```

I do wrap this behavior into a Stream using `Stream.resource/3` and passing in the client_ref to the response.

Reason I had couldn't use just hand the stream to ExAws.Upload is because hackney only allows you to call `:hackney.stream_body` for a client ref that's owned by the current process. When ExAws does `GenStage.from_enumerable/2`, it tries (and fails) to read the stream under a new process that is not the owner of the client_ref. That's why I have to manually start up the GenStage, then call `:hackney.controlling_process(client_ref, download_stage)` to transfer ownership and then hand it off to ExAWS to do the upload part.

Please let me know if that answers your questions. It took quite a bit of troubleshooting to figure this out, so I hope this makes sense without having to dive into the bowels of hackney. This PR would allow me to my own GenStage to ExAws.Upload so I don't have to rewrite the batch upload code myself.

aaronrenner:
I just got back from vacation and found this branch has fallen pretty far behind. Just rebased on top of master.

lib/ex_aws/s3.ex
```diff
@@ -488,22 +489,54 @@ defmodule ExAws.S3 do
...
 
-  ## Notes
+  ## Uploading a flow
+
+  GenStage Flows can also be uploaded directly to S3. However, due to the concurrent
+  nature of flows, the producer must emit events in the following format
+  `{binary, index}` to ensure they are uploaded in the correct order. Binaries
+  will be uploaded as a chunk and must be at least 5 megabytes in size.
+  ```
+  enumerable = ["hello world"] |> Enum.with_index(1)
```
benwilson512: We probably want to promote `Stream.with_index` here as most people's enumerables are gonna be lazy. If it isn't lazy we're gonna lose a lot of the memory savings this approach offers.

lib/ex_aws/s3.ex
```diff
@@ -488,22 +489,54 @@ defmodule ExAws.S3 do
...
-  |> ExAws.request! #=> {:ok, :done}
+  |> ExAws.request! #=> :done
   ```
 
-  ## Notes
+  ## Uploading a flow
+
+  GenStage Flows can also be uploaded directly to S3. However, due to the concurrent
+  nature of flows, the producer must emit events in the following format
+  `{binary, index}` to ensure they are uploaded in the correct order. Binaries
```
benwilson512: > However, due to the concurrent nature of flows, the producer must emit events in the following format `{binary, index}` to ensure they are uploaded in the correct order.

The chunks can be uploaded in arbitrary order. Rather, the indices are used by AWS to determine in what order to combine the pieces when completing the result. I would say:

" All parts uploaded to AWS must be tagged with a `1` based index so that they can be assembled in order when finalizing the multi part upload."

lib/ex_aws/s3.ex
```diff
@@ -488,22 +489,54 @@ defmodule ExAws.S3 do
...
+      path: path,
+      opts: opts,
+    }
+  end
+  defp do_upload(source, bucket, path, opts) do
+    source_stream = Stream.with_index(source, 1)
+
+    flow =
+      Flow.new(stages: opts[:max_concurrency] || 4, max_demand: 2)
+      |> Flow.from_enumerable(source_stream)
```
benwilson512: Can you actually update GenStage to the latest version? They've made some changes so this would now be

```elixir
flow = Flow.from_enumerable(source_stream, stages: opts[:max_concurrency] || 4, max_demand: 2)
```

lib/ex_aws/s3.ex
```diff
@@ -488,22 +489,54 @@ defmodule ExAws.S3 do
...
+  Flow.new(stages: 4, max_demand: 2)
+  |> Flow.from_enumerable(enumerable)
+  |> S3.upload("my-bucket", "test.txt")
+  |> ExAws.request! #=> :done
+  ```
   """
-  def upload(source, bucket, path, opts) do
+  def upload(source, bucket, path, opts \\ []) do
+    do_upload(source, bucket, path, opts)
+  end
```
benwilson512: `do_upload` largely unnecessary. You could just have
```elixir
def upload(source, bucket, path, opts \\ [])
def upload(%Flow{} = flow, bucket, path, opts) do
  # flow stuff
end
def upload(stream, bucket, path, opts) do
  # non flow stuff
end
```

We should also add typespecs. In this case that should be:
```elixir
@spec upload(source :: Enumerable.t | Flow.t, bucket :: binary, path :: binary) :: __MODULE__.Upload.t
@spec upload(source :: Enumerable.t | Flow.t, bucket :: binary, path :: binary, opts :: Keyword.t) :: __MODULE__.Upload.t
```

benwilson512:
> Reason I had couldn't use just hand the stream to ExAws.Upload is because hackney only allows you to call :hackney.stream_body for a client ref that's owned by the current process. When ExAws does GenStage.from_enumerable/2, it tries (and fails) to read the stream under a new process that is not the owner of the client_ref. That's why I have to manually start up the GenStage, then call :hackney.controlling_process(client_ref, download_stage) to transfer ownership and then hand it off to ExAWS to do the upload part.

This isn't exactly correct. `GenStage.from_enumerable` does indeed spawn a new process, however Flow.from_enumerable does not. However Flow.from_enumerable does require exclusive access to the process mailbox, which is likely your trouble here.

It's that last point that has convinced me :). Basically there are definitely scenarios where people want to stream stuff based on the process inbox, and that means targetting an API that makes that possible: Flow.

lib/ex_aws/s3.ex
aaronrenner: That's super cool that you can use a bodiless function clause for default arguments. I learned something today. 🎸 😎

lib/ex_aws/s3.ex
```diff
@@ -517,19 +517,21 @@ defmodule ExAws.S3 do
...
-  def upload(source, bucket, path, opts \\ []) do
-    do_upload(source, bucket, path, opts)
-  end
-
-  defp do_upload(%Flow{} = flow, bucket, path, opts) do
+  @spec upload(
+    source :: Enumerable.t | Flow.t,
+    bucket :: String.t,
+    path :: String.t,
+    opts :: Keyword.t) :: __MODULE__.Upload.t
```
aaronrenner: Here's what I ended up with for a typespec. I saw that ExDoc only displays one typespec, and checked in Ecto to see how they handle an optional keyword list: https://github.com/elixir-ecto/ecto/blob/v2.1.0-rc.0/lib/ecto/changeset.ex#L1065. I also went with `String.t` for the bucket and path instead of binary because I think we want a standard UTF-8 encoded string not a possibly non-printable binary.

Let me know if you prefer the typespec how you previously had it.

lib/ex_aws/s3.ex
aaronrenner: I absolutely agree we need to upgrade, but these Flow API changes are on 0.6.0-dev and are not released yet. I'm a little hesitant to point the GenStage dependency to the master branch. Should I do another PR as soon as GenStage 0.6.0 is released?

lib/ex_aws/s3.ex
aaronrenner: Good point. Done.

lib/ex_aws/s3.ex
aaronrenner: Done.

aaronrenner:
Apparently I don't know how to correctly use Github Reviews yet. Comments are above, though.

lib/ex_aws/s3.ex
benwilson512: Yes please!

lib/ex_aws/s3.ex
benwilson512: I think this is fine. It might be nice however to document the opts, which would at this point just be `[max_concurrency: pos_integer]`

lib/ex_aws/s3.ex
aaronrenner: I see GenStage 0.6.0 was just released last night. I'll make the change and include it in this PR.

lib/ex_aws/s3.ex
```diff
@@ -511,35 +511,30 @@ defmodule ExAws.S3 do
...
+
+  ## Options
+
+  These options are specific to this function
+
+  * `:max_concurrency` -- The number of concurrent processes reading from this
+     stream. Only applies when uploading a stream.
+
+  All other options (ex. `:content_type`) are passed through to
+  `ExAws.S3.initiate_multipart_upload/3`.
```
aaronrenner: I added documentation for the options, but didn't add it to the typespec because I wasn't sure how to easily include all of the options that are passed through to `ExAws.S3.initiate_multipart_upload/3`

test/lib/ex_aws/s3/download_test.exs
```diff
@@ -0,0 +1,47 @@
+defmodule ExAws.S3.DownloadTest do
```
aaronrenner: There was an update that needed to be made to `ExAws.S3.Download` for the GenStage upgrade, but I found there was no test coverage. Now there's test coverage 😄.

lib/ex_aws/s3.ex
```diff
@@ -511,35 +511,30 @@ defmodule ExAws.S3 do
...
-      opts: opts,
-    }
-  end
-  def upload(source, bucket, path, opts) do
-    source_stream = Stream.with_index(source, 1)
-
-    flow =
-      Flow.new(stages: opts[:max_concurrency] || 4, max_demand: 2)
-      |> Flow.from_enumerable(source_stream)
-
```
aaronrenner: I moved the flow logic to the `ExAws.Operation.perform` function for `ExAws.S3.Upload` because it sounded like `Flow.from_enumerable` creates a GenServer immediately when it was called in the 0.5.0 documentation. In  0.6.0 it seems more unclear when the gen server is created, so moving this logic may not have been necessary. Fortunately wherever this logic lives its behind a private api so we can refactor this later if needed.

lib/ex_aws/s3/download.ex
```diff
@@ -72,10 +72,9 @@ defimpl ExAws.Operation, for: ExAws.S3.Download do
       file
     end
 
-    chunk_stream = Download.build_chunk_stream(op, config)
-
-    Flow.new(stages: op.opts[:max_concurrency] || 8, max_demand: 2)
-    |> Flow.from_enumerable(chunk_stream)
+    op
+    |> Download.build_chunk_stream(config)
+    |> Flow.from_enumerable(stages: op.opts[:max_concurrency] || 8, max_demand: 2)
```
benwilson512: Thanks for grabbing this one :)

lib/ex_aws/s3/upload.ex
```diff
@@ -89,4 +89,16 @@ defimpl ExAws.Operation, for: ExAws.S3.Upload do
...
+  defp build_producer(%Flow{} = flow, _opts) do
+    flow
+  end
+  defp build_producer(source, opts) do
+    source
+    |> Stream.with_index(1)
+    |> Flow.from_enumerable(
+          stages: Keyword.get(opts, :max_concurrency, 4),
+          max_demand: 2
+        )
```
benwilson512: 👍

lib/ex_aws/s3.ex
benwilson512: Ah.
```
@type upload_opts [max_concurrency: pos_integer] | initiate_multipart_upload_opts
```

test/lib/ex_aws/s3/download_test.exs
benwilson512: Thanks!

Haven't really used bypass before, it could likely be a valuable addition to this library's test suite given that it's an API client and all.

aaronrenner:
Ok, this is rebased and I think it's ready to go. 

Let me know if you see anything else, and thanks for all of your help!

benwilson512:
Looks good to me, thanks a bunch!

PR #225, opened 2016-09-23
Move retry options to configuration

### LINKED ISSUE #224: Could we make retry options configurable?
I'm using [Bypass](https://github.com/PSPDFKit-labs/bypass) in my test suite to intercept AWS calls:

```elixir
# config/test.exs

use Mix.Config

config :ex_aws, sqs: %{
  scheme: "http://",
  host: "127.0.0.1",
  region: "us-east-1",
  port: 1234
}

# a test might look like

defmodule Test do
  use ExUnit.Case, async: true

  setup do
    port = Application.fetch_env!(:ex_aws, :sqs).port
    bypass = Bypass.open(port: port)
    {:ok, bypass: bypass}
  end

  test "when AWS is down", %{bypass: bypass} do
    Bypass.expect bypass, fn conn -> Plug.Conn.resp(conn, 500, "down") end

    assert_raise ExAws.Error, fn ->
      ExAws.SQS.list_queues |> ExAws.request!
    end
  end
end
```

But this test takes a long time to run due to all the retries with exponential backoff, which are hard-coded into `ExAws`: https://github.com/CargoSense/ex_aws/blob/11dcf1d45abd99c53e7665c67b4c7dcaae0392d8/lib/ex_aws/request.ex#L3-L5

If only for speeding up test suites, could we make the retry options configurable? It might also be useful for people who want to tweak the defaults.

ajvondrak:
No tests, but there weren't tests around this code in the first place. I did try it out on my project locally (using a `path:` dependency) and it seems to work as planned. Hope the code isn't too ugly!

Fixes #224.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,14 +114,13 @@ ExAws will retry failed AWS API requests using exponential backoff per the "Full
 
 The algorithm uses three values, which are configurable:
 
-```
+```elixir
 # default values shown below
 
-config :ex_aws, retries: %{
+config :ex_aws, :retries,
   max_attempts: 10,
   base_backoff_in_ms: 10,
-  max_backoff_in_ms: 10_000,
-}
+  max_backoff_in_ms: 10_000
 ```
 
 * `max_attempts` is the maximum number of possible attempts with backoffs in between each one
--- a/README.md
+++ b/README.md
@@ -108,6 +108,26 @@ config :ex_aws,
 
 This means it will first look for the AWS standard `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables, and fall back using instance meta-data if those don't exist. You should set those environment variables to your credentials, or configure an instance that this library runs on to have an iam role.
 
+## Retries
+
+ExAws will retry failed AWS API requests using exponential backoff per the "Full Jitter" formula described in https://www.awsarchitectureblog.com/2015/03/backoff.html
+
+The algorithm uses three values, which are configurable:
+
+```
+# default values shown below
+
+config :ex_aws, retries: %{
+  max_attempts: 10,
+  base_backoff_in_ms: 10,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -108,6 +108,26 @@ config :ex_aws,
 
 This means it will first look for the AWS standard `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables, and fall back using instance meta-data if those don't exist. You should set those environment variables to your credentials, or configure an instance that this library runs on to have an iam role.
 
+## Retries
+
+ExAws will retry failed AWS API requests using exponential backoff per the "Full Jitter" formula described in https://www.awsarchitectureblog.com/2015/03/backoff.html
+
+The algorithm uses three values, which are configurable:
+
+```
```
benwilson512: we want to tag this as elixir code via ```elixir

README.md
```diff
@@ -108,6 +108,26 @@ config :ex_aws,
...
+# default values shown below
+
+config :ex_aws, retries: %{
+  max_attempts: 10,
+  base_backoff_in_ms: 10,
+  max_backoff_in_ms: 10_000,
+}
```
benwilson512: Config by convention uses keyword lists, I would recommend those here as well.

```
config :ex_aws, :retries,
  max_attempts: 10,
  base_backoff_in_ms: 10,
  max_backoff_in_ms: 10_000
```

benwilson512: Because you're using the Access behaviour when accessing retries this probably doesn't really require any changes to the config module, but you should confirm that it works happily anyway.

ajvondrak: Cf. the commit I just pushed, `Access` works happily, but I _did_ uncover another function that broke. Seems to be in working order now (so long as all three parameters are always configured).

benwilson512:
Looks pretty good to me! Thank you!!!

PR #227, opened 2016-09-25
Add a policy of ec2 in order to pass the ec2-integration-test

kanmo:
In order to in pass the `test/lib/ex_aws/ec2/integration_test.exs`  of TravisCI, 
you need to add the EC2 of the policy to the IAM.

Can you add the policy of EC2 to IAM that are used in TravisCI?


benwilson512:
Thanks!!!

PR #232, opened 2016-09-30
Improvements for the SQS portion of the library

scohen:
This fixes a bug in the list_queues function as well as adding parsing for all XML responses. 
Added unit tests for each response type, and I'll add more for errors as I encounter them.

@benwilson512

### DIFF
--- a/lib/ex_aws/sqs.ex
+++ b/lib/ex_aws/sqs.ex
@@ -169,7 +169,10 @@ defmodule ExAws.SQS do
   @spec list_queues() :: ExAws.Operation.Query.t
   @spec list_queues(opts :: [queue_name_prefix: binary]) :: ExAws.Operation.Query.t
   def list_queues(opts \\ []) do
-    request("", "ListQueues", Enum.into(opts, %{}))
+    params = opts
+    |> format_regular_opts
+
+    request("", "ListQueues", params)
   end
 
   @doc "Purge all messages in a SQS Queue"
--- a/lib/ex_aws/sqs.ex
+++ b/lib/ex_aws/sqs.ex
@@ -65,7 +65,7 @@ defmodule ExAws.SQS do
       |> format_permissions
       |> Map.put("Label", label)
 
-    request(queue, "AddPermission", params)
+    request(queue, :add_permission, params)
   end
 
   @doc """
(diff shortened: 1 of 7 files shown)


lib/ex_aws/sqs.ex
```diff
@@ -259,10 +262,16 @@ defmodule ExAws.SQS do
   end
 
   defp request(queue, action, params) do
+    action_atom = action
+    |> Macro.underscore
+    |> String.to_atom
```
benwilson512: Would it be preferable to have for example `request("", :list_queues, params)` and then

```
defp request(queue, action_atom, params) do
  action = action_atom |> Atom.to_string |> Macro.camelize
end

File: test/lib/ex_aws/sqs/parser_test.exs [hunk 2]
```diff
@@ -0,0 +1,434 @@
+defmodule ExAws.SQS.ParserTest do
+  use ExUnit.Case, async: true
```
benwilson512: You're the best person on earth for including all of these tests.

lib/ex_aws/sqs/parsers.ex
```diff
@@ -0,0 +1,204 @@
...
+
+      {:ok, Map.put(resp, :body, parsed_body)}
+    end
+
+    def attribute_list_to_map(list_of_maps) do
+      list_of_maps
+      |> Enum.reduce(%{}, fn(%{name: name, value: val}, acc) ->
+        attribute_name = name
+        |> Macro.underscore
+        |> String.to_atom
```
benwilson512: The other `String.to_atom` call is fine cause all the inputs are basically hardcoded.

Aren't attribute lists basically arbitrary input?

lib/ex_aws/sqs/parsers.ex
```diff
@@ -0,0 +1,204 @@
+if Code.ensure_loaded?(SweetXml) do
+  defmodule ExAws.SQS.Parsers do
+    import SweetXml, only: [sigil_x: 2]
+
+    def parse({:ok, resp=%{body: xml}}, :list_queues) do
```
benwilson512: I'm gonna be "that guy" and note that as a stylistic thing I tend to prefer `%{body: xml} = resp`

lib/ex_aws/sqs.ex
scohen: Ok. I was trying to make as few changes as possible, but atoms are great.

test/lib/ex_aws/sqs/parser_test.exs
scohen: I like tests. A lot.

lib/ex_aws/sqs/parsers.ex
scohen: I don't have a preference. NBD.

lib/ex_aws/sqs/parsers.ex
scohen: Oh, derp.

Issue #207, opened 2016-08-30
Presigned URLs are borked for some reason on master


benwilson512:

benwilson512:
#228 is a better issue.

PR #233, opened 2016-09-30
Fix some issues in generating process of presigned url

kenrota:
Fix some issues in generating process of presigned url

Summary
* Remove unnecessary url encoding/decoding
 * No need to replace "+" to " "
* Use canonical query in both query to sign and query of result `presigned_url`
* Keep ordering of original query in presigned url
* Update test cases
 * expected `X-Amz-Signature`s are calculated by using https://github.com/kenrota/scriptbox/blob/master/ruby/gen_presigned_url_base_on_aws_example.rb

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -168,13 +168,13 @@ defmodule ExAws.Auth do
       acc <> "#{to_string(key)}=#{to_string(value)}&"
     end)
     <> "X-Amz-Algorithm=AWS4-HMAC-SHA256&"
-    <> "X-Amz-Credential=#{uri_encode(Credentials.generate_credential_v4(service, config, datetime))}&"
+    <> "X-Amz-Credential=#{Credentials.generate_credential_v4(service, config, datetime)}&"
     <> "X-Amz-Date=#{amz_date(datetime)}&"
     <> "X-Amz-Expires=#{expires}&"
     <> "X-Amz-SignedHeaders=host"
     <> case config[:security_token] do
          nil -> ""
-         token -> "&X-Amz-Security-Token=#{uri_encode(token)}"
+         token -> "&X-Amz-Security-Token=#{token}"
        end
   end
 end
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -33,9 +33,7 @@ defmodule ExAws.Auth do
 
     org_query_params = query_params |> Enum.map(fn({k, v}) -> {to_string(k), v} end)
     amz_query_params = build_amz_query_params(service, datetime, config, expires)
-    {org_query, amz_query} = [org_query_params, amz_query_params]
-    |> Enum.map(&(canonical_query_params(&1)))
-    |> List.to_tuple
+    [org_query, amz_query] = [org_query_params, amz_query_params] |> Enum.map(&canonical_query_params/1)
     query_to_sign = org_query_params ++ amz_query_params |> canonical_query_params
     query_for_url = if Enum.count(org_query_params) == 0 do amz_query else org_query <> "&" <> amz_query end
 
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -51,7 +51,7 @@ defmodule ExAws.Auth do
   defp auth_header(http_method, url, headers, body, service, datetime, config) do
     uri = URI.parse(url)
     path = uri_encode(uri.path)
-    query = uri.query |> canonical_query_params
+    query = if uri.query, do: uri.query |> URI.decode_query |> Enum.to_list |> canonical_query_params, else: ""
     signature = signature(http_method, path, query, headers, body, service, datetime, config)
     [
       "AWS4-HMAC-SHA256 Credential=", Credentials.generate_credential_v4(service, config, datetime), ",",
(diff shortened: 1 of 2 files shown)


lib/ex_aws/auth.ex
```diff
@@ -30,10 +30,19 @@ defmodule ExAws.Auth do
...
-    query = presigned_url_query(service, datetime, config, expires, query_params)
-    url = "#{url}?#{query}"
-    signature = signature(http_method, url, headers, nil, service, datetime, config)
-    "#{url}&X-Amz-Signature=#{signature}"
+
+    org_query_params = query_params |> Enum.map(fn({k, v}) -> {to_string(k), v} end)
+    amz_query_params = build_amz_query_params(service, datetime, config, expires)
+    {org_query, amz_query} = [org_query_params, amz_query_params]
+    |> Enum.map(&(canonical_query_params(&1)))
+    |> List.to_tuple
```
benwilson512: These 3 lines could just easily be:
```
[org_query, amz_query] = [org_query_params, amz_query_params] |> Enum.map(&canonical_query_params/1)
```

lib/ex_aws/auth.ex
```diff
@@ -30,10 +30,19 @@ defmodule ExAws.Auth do
...
+    query_to_sign = org_query_params ++ amz_query_params |> canonical_query_params
+    query_for_url = if Enum.count(org_query_params) == 0 do amz_query else org_query <> "&" <> amz_query end
```
benwilson512: This should be:
```
if Enum.any?(org_query_params), do : org_query <> "&" <> amz_query, else: amz_query
```

lib/ex_aws/auth.ex
```diff
@@ -114,9 +123,13 @@ defmodule ExAws.Auth do
   end
 
   defp canonical_query_params(nil), do: ""
-  defp canonical_query_params(params) do
+  defp canonical_query_params(params) when is_binary(params) do
+    params
+    |> String.split("&") |> Enum.map(&List.to_tuple(String.split(&1, "=")))
+    |> canonical_query_params
+  end
```
benwilson512: When might it be binary? I'm not really a fan of having a basic URI params parser here. Doesn't URI.parse do this anwyay?

benwilson512:
FYI because this fixes a pretty critical bug in master I'd like to get this merged in today if possible.

kenrota:
>Hey there, I appreciate the changes. I do know that master is broken for presigned URLs per #207 . Have you tried this against S3 with presigned URLs?

Hi Mr. Ben Wilson

Thank you for your comments.
I've tried this fixes on local machine using access_key_id/secret_access_key, and on EC2 using  IAM role.

That test cases are:
* put without optional query. It includes only X-AMZ-XXX.
* put with optional query. (x-amz-acl. etc)
 * I wrote this kind of query with "original" in PR.
* get without optional query.

All are successed.

I've not tried multipart upload.

kenrota:
>When might it be binary? I'm not really a fan of having a basic URI params parser here. Doesn't URI.parse do this anwyay?

>My bigger question is why are we doing this at all? Why might this function get a binary at all that  we have to manually parse. If it gets a binary then that tells me we're encoding to a binary too soon.

Sorry, i'm not good at english.
Please let me confirm your questions.

Your question is why I added `when is_binary(params)` in  https://github.com/CargoSense/ex_aws/pull/233/commits/3bac4df21b9e5f46c3a9a526b63aac8b50c0cd68 ?

kenrota:
> Your question is why I added when is_binary(params) in 3bac4df ?

If yes, the reason is below.

* Source code makes `canonical_query_params` in `build_canonical_request`.
* But my changes makes `canonical_query_params` out of `build_canonical_request` and `signature`.
  So, I thought `auth_header` needs `canonical_query_params` before `signature`
  In this case, argument of `canonical_query_params` is string.
  I thought string is binary (Is it correct?), Then, I added `when is_binary(params)` for string case.

kenrota:
I will fix my previous comment.

* wrong
 * out of `build_canonical_request`
* correct
 * outside of  `build_canonical_request`

lib/ex_aws/auth.ex
kenrota: Thank your for your simple code. I fixed it. d452679cd9e00987e684608ecb1c475e614fb06a

lib/ex_aws/auth.ex
kenrota: Thank you for your readable and better code. I fixed it. https://github.com/CargoSense/ex_aws/pull/233/commits/bba05c30569408ad9b4950a829b8861408998e71

benwilson512:
I see, so in line https://github.com/kenrota/ex_aws/blob/e0ea7dfd6f9e9bd3effbcef804755bd5d8d2e810/lib/ex_aws/auth.ex#L54 `uri.query` is a binary. Is there a way we can use http://elixir-lang.org/docs/stable/elixir/URI.html#decode_query/2 there instead of having our own decoder?

```
uri.query |> URI.decode_query |> canonical_query_params
```

And then all we need is:
```
  defp canonical_query_params(params) do
    params
    |> Enum.sort(fn {k1, _}, {k2, _} -> k1 < k2 end)
    |> Enum.map_join("&", &pair/1)
  end
```

kenrota:
ce47317779437f2afbd865a144d02b84d96a441d I fixed for https://github.com/CargoSense/ex_aws/pull/233#issuecomment-250891867.
Correct commit message is `"Refactoring: Use URI.decode_query instead of manual decoding"`.

benwilson512:
This looks good, thank you very much!

PR #236, opened 2016-10-03
Added :phone_number to SNS publish_opts & updated spec.

jglover:
As per AWS SNS [API docs](http://docs.aws.amazon.com/sns/latest/api/API_Publish.html), you can specify PhoneNumber as a request parameter; provided you don't specify :target_arn or :topic_arn. 
This allows for the sending of SMS on a one-time basis rather than to subscribers of a particular topic (useful for sending one-time access codes etc).

Whilst the code still executes without this change, it should probably be made to clarify the options that `publish_opts` has, and should help towards breaking changes in the future.

Tests are passing as before.


benwilson512:
Thanks!

PR #237, opened 2016-10-03
SNS message_attributes struct should accept atoms representing type

jglover:
Current SNS message_attributes struct only accepts strings representing the type, rather than atoms. Not sure if this is intended behaviour or not but it seems odd.


benwilson512:
Thanks!

PR #240, opened 2016-10-04
Handled Message Attributes

scohen:
SQS has the concept of Attributes *and* MessageAttributes. Attributes
are intrinsic to a message, and are set by the system.
MessageAttributes, however are set by the user and can be
anything.

This PR adds support for getting and setting message attributes, and can
parse numeric, binary and string attributes.

@benwilson512

### DIFF
--- a/lib/ex_aws/sqs.ex
+++ b/lib/ex_aws/sqs.ex
@@ -183,6 +183,7 @@ defmodule ExAws.SQS do
 
   @type receive_message_opts :: [
     {:attribute_names, :all | [sqs_message_attribute_name, ...]} |
+    {:message_attribute_names, :all | [String.t, ...]} |
     {:max_number_of_messages, 1..10} |
     {:visibility_timeout, 0..43200} |
     {:wait_time_seconds, 0..20}
@@ -195,10 +196,13 @@ defmodule ExAws.SQS do
     {attrs, opts} = opts
     |> Keyword.pop(:attribute_names, [])
 
-    params =
-      attrs
-      |> format_queue_attributes
-      |> Map.merge(format_regular_opts(opts))
+    {message_attrs, opts} = opts
+    |> Keyword.pop(:message_attribute_names, [])
+
+    params = attrs
+    |> format_queue_attributes
+    |> Map.merge(format_message_attributes(message_attrs))
+    |> Map.merge(format_regular_opts(opts))
 
     request(queue, :receive_message, params)
   end
@@ -326,6 +330,21 @@ defmodule ExAws.SQS do
     key = "AttributeName.#{index + 1}"
 
     Map.put(%{}, key, format_param_key(attribute))
+  end
+
+  defp format_message_attribute({attribute, index}) do
+    %{"MessageAttributeName.#{index + 1}" => attribute}
+  end
+
+  defp format_message_attributes(:all) do
+    %{"MessageAttributeNames" => "All"}
+  end
+
+  defp format_message_attributes(attributes) do
+    attributes
+    |> Enum.with_index
+    |> Enum.map(&format_message_attribute/1)
+    |> Enum.reduce(%{}, &Map.merge(&1, &2))
   end
 
   defp format_batch_message(message, index) do
--- a/lib/ex_aws/sqs.ex
+++ b/lib/ex_aws/sqs.ex
@@ -333,7 +333,7 @@ defmodule ExAws.SQS do
   end
 
   defp format_message_attribute({attribute, index}) do
-    %{"MessageAttributeName.#{index + 1}" => attribute}
+    %{"MessageAttributeName.#{index + 1}" => to_string(attribute)}
   end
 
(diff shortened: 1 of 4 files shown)


benwilson512:
Did you have a chance to look at the existing PR for this? https://github.com/CargoSense/ex_aws/pull/189 I'll take a look here in a min, just want to see what the difference is.

scohen:
I just noticed it. 

So, mine does a couple things differently; namely, it parses the responses coming back into binaries, strings or numbers, and my tests are better. 

I do wish I had seen the other PR though.

lib/ex_aws/sqs.ex
```diff
@@ -328,6 +332,21 @@ defmodule ExAws.SQS do
     Map.put(%{}, key, format_param_key(attribute))
   end
 
+  defp format_message_attribute({attribute, index}) do
+    %{"MessageAttributeName.#{index + 1}" => attribute}
```
scohen: I might need to call to_string here.

benwilson512: Yup let's do that, and change the type signature for these to `String.Chars.t` cause that'll cover anything handled by `to_string`

lib/ex_aws/sqs/parsers.ex
```diff
@@ -175,6 +179,55 @@ if Code.ensure_loaded?(SweetXml) do
...
+
+    defp parse_attribute_value(<<"String", _rest :: binary>>, %{string_value: string_value}) do
+      string_value
+    end
+
+    defp parse_attribute_value(<<"Binary", _rest :: binary>>, %{binary_value: b64_encoded}) do
+      :base64.decode(b64_encoded)
+    end
+
+    defp parse_attribute_value(<<"Number", _rest :: binary>>, %{string_value: string_value}) do
```
benwilson512: Minor note, you can just do these as `"Number" <> _` which I like in these cases where it's a nice string literal.

lib/ex_aws/sqs/parsers.ex
```diff
@@ -175,6 +179,55 @@ if Code.ensure_loaded?(SweetXml) do
       {:ok, Map.put(resp, :body, parsed_body)}
     end
 
+    def message_attributes_to_map(list_of_message_attribtues) do
+      list_of_message_attribtues
+      |> Enum.reduce(%{}, fn(%{name: name, data_type: data_type} = attr, acc) ->
+        case parse_attribute_value(data_type, attr) do
+          ^attr ->
+            Map.put(acc, name, attr)
```
benwilson512: Probably a dumb question, but why do we not want a value and data type map here?

benwilson512:
I'm happy to merge this and close the other one after we've hashed out details.

lib/ex_aws/sqs.ex
scohen: done.

lib/ex_aws/sqs/parsers.ex
scohen: Oh, and you will have the data type in there. There's a test that exercises this case.

benwilson512: Right ok.

Would it make sense to do:
```
Map.put(acc, name, Map.put(attr, :value, parsed))
```

on line 189 there so that if someone wanted to do something fancy they had the original value around?  If not that's fine.

lib/ex_aws/sqs/parsers.ex
```diff
@@ -175,6 +179,55 @@ if Code.ensure_loaded?(SweetXml) do
...
+          update_in(parsed_body, attribute_path, fixup_fn)
+      end
+    end
+
+    defp parse_attribute_value(<<"String", _rest :: binary>>, %{string_value: string_value}) do
+      string_value
+    end
+
+    defp parse_attribute_value(<<"Binary", _rest :: binary>>, %{binary_value: b64_encoded}) do
+      :base64.decode(b64_encoded)
```
benwilson512: This function raises with argument error if given non base64 encoded input. Other encoding errors merely return the original value as per our discussion above. We may want
```
case Base.decode64(b64_encoded) do
  {:ok, decoded} -> decoded
  _ -> b64_encoded
end
```

lib/ex_aws/sqs/parsers.ex
scohen: sure.

scohen: done.

lib/ex_aws/sqs/parsers.ex
scohen: I'll change it, but the SQS docs say that all binary data is base64 encoded for you.

benwilson512: Ah in that case don't. `Base.decode64!` however might be better

benwilson512: lol sorry should have replied in time before you pushed the commit

PR #242, opened 2016-10-05
Handled multiple messages from receive_message

scohen:
If multiple messages are requested, the response contains the key
`:messages`, which contains a list of the messages. If one is requested,
the response contains the `:message` key.

@benwilson512, I'm not sure how I feel about changing the key based on the number of messages returned. Thoughts appreciated.

### DIFF
--- a/lib/ex_aws/sqs/parsers.ex
+++ b/lib/ex_aws/sqs/parsers.ex
@@ -127,17 +127,9 @@ if Code.ensure_loaded?(SweetXml) do
         |> fix_attributes([:message_attributes], &message_attributes_to_map/1)
       end)
 
-      fixed_attributes = case new_messages do
-                           [message] ->
-                             parsed_body
-                             |> Map.put(:message, message)
-                             |> Map.delete(:messages)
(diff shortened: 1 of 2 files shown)


lib/ex_aws/sqs/parsers.ex
```diff
@@ -119,11 +119,17 @@ if Code.ensure_loaded?(SweetXml) do
                           ]
                         ])
 
-      fixed_attributes = parsed_body
-      |> fix_attributes([:message, :attributes], &attribute_list_to_map/1)
-      |> fix_attributes([:message, :message_attributes], &message_attributes_to_map/1)
+      new_messages = parsed_body
+      |> Map.get(:messages)
```
benwilson512: Since `:messages` is marked as optional in theory `Map.get(parsed_body, :messages)` could return `nil`. We either want it non optional or with a `|> Map.get(:messages, [])`

scohen: Will do, but it looks like sweetXML returns an empty list if messages isn't present (There's a test for this case).
I'll still add it as it's clearer.

PR #243, opened 2016-10-05
Fixed error parsing

scohen:
Something changed and caused the error shape to change, which broke the
error parsers.

We should figure out how to prevent regressions like this in the future.
@benwilson512

### DIFF
--- a/lib/ex_aws/sqs/parsers.ex
+++ b/lib/ex_aws/sqs/parsers.ex
@@ -157,7 +157,7 @@ if Code.ensure_loaded?(SweetXml) do
       parse_request_id(resp, ~x"//SetQueueAttributesResponse")
     end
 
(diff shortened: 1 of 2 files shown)


benwilson512:
Ah shoot that change happened because certain S3 errors IE `HEAD bucket` where the bucket is non existent or forbidden can result in bodiless errors.

scohen:
@benwilson512, Is there a common builder for responses? If we all used that, it would ensure that unit tests would fail if the parsing changed.

benwilson512:
At the moment the possible error formats are contained almost entirely here: https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/request.ex

Each individual operation type can customize error handling like we do for parsing. There's a couple things in the request.ex module I need to move out to the dynamo operation module actually.

PR #244, opened 2016-10-05
Sns delete endpoint

jeffskelton3:
Added the ability to delete an SNS endpoint (useful if one wishes to re-assign a device token).  Already using on a production app and its working as expected. 

Thanks for making this. Its been very useful.


benwilson512:
Thanks a bunch!

PR #245, opened 2016-10-05
Add ini file configuration support

tomasz-tomczyk:
This adds support for ~/.aws/credentials file used by most other AWS SDK libraries including the [official CLI](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html)

Where I work we switched to temporary keys generated as and when we need them and store them in that file - found out today ex_aws does not read them :)

This also helped me run the tests, as by default, 17 failed for me, I believe due to the default configuration pointing at US AWS (I'm located in London).

I'm new to Elixir so I'd appreciate feedback and comments for the PR! I'd like to add support to determining profile that should be loaded, but wanted your thoughts first on if I'm on the right path.


benwilson512:
Hey there, thanks for the PR!

We did have another PR before on this issue here https://github.com/CargoSense/ex_aws/pull/139 although they were using another external library https://github.com/easco/configparser_ex

I sort of thing we want to use that library for parsing the INI files because it's likely to handle all the possibilities better, plus I don't want to have to maintain a custom INI parser.

The only other issue is that this makes the INI parsing mandatory. I would prefer a solution like we have with stuff like `:instance_role` where if you want `:awscli` you can stick it on the list IE
```
access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :awscli, :instance_role],
```

This way there's a clear ordering to what's chosen. It'd also be nice if the ini config parser library was not actually necessary unless you were choosing to use `:awscli`.

Lastly we'll also need caching, because reading and parsing the file on every request is definitely not optimal. We've already got an `auth_cache` mechanism in place for handling instance roles and we could probably just integrate into that.

tomasz-tomczyk:
Hey,
Thanks for the feedback! I've started implementing the suggestions from you as I can't see any outcome of the other PR you've linked. I'll ping you once I'm ready for round 2 of feedback if that's cool :)

- [x] Replace ini parsing library with configparser_ex
- [ ] Make the ini file parsing optional (ideally including loading the library, if possible)
- [ ] Push config file lookup to a cache service

test/lib/ex_aws/config_test.exs
```diff
@@ -5,7 +5,7 @@ defmodule ExAws.ConfigTest do
     value = "foo"
     System.put_env("ExAwsConfigTest", value)
     assert :s3
-    |> ExAws.Config.new([access_key_id: {:system, "ExAwsConfigTest"}])
+    |> ExAws.Config.new([access_key_id: {:system, "ExAwsConfigTest"}, secret_access_key: {:system, "AWS_SECURITY_TOKEN"}])
```
tomasz-tomczyk: It appears that if you don't overwrite this, the default way of loading secret_access key will be used (and I updated the defaults to include `:awscli`) -- and the way the Config module works, it loads all variables the config options provide -- overwriting other settings. Does this make sense?

Likewise, :instance_profile will override a few keys as per https://github.com/tomasz-tomczyk/ex_aws/blob/bceaf3bc55bda27df995b778629d35d1ec3d2695/lib/ex_aws/config.ex#L64-L68

In theory, it should be possible to load access_key_id from ENV and secret_access_key from another source, but doesn't seem possible by design with current implementation. Should be an edge case, anyway, so not sure if worth refactoring?

tomasz-tomczyk:
@benwilson512 I've made some progress! Would appreciate further feedback. :)

I've added the cache layer, but worried it might be too much of a replication of AuthCache? They do have different responsibilities, but a lot of it could be DRY'd up.

I'm also not sure about the expiration setting; that setting isn't available in the files, so we can either make it an option passed in the config (my preferred solution) or have an arbitrary value. Happy to add that in if you agree.

Not entirely sure if I've done optional loading of library correctly.. Let me know if that's all I needed!

Also, have a look at the comment I've made above regarding another test I had to edit to make it pass. I hope you agree that it's out of scope of this PR and we can tackle it separately if needed.

Thanks!

tomasz-tomczyk:
The more I think about it, the less I like replicating the entire Cache bit - let me try and integrate it into the existing AuthCache!

tomasz-tomczyk:
Ok, i've addressed both the DRY aspect of cache layer and configurability of expiration time. How does that look?

lib/ex_aws/credentials_ini.ex
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+    def parse_ini_file(_, _), do: %{}
+
+    def strip_key_prefix(credentials) do
+      credentials
+      |> Enum.reduce(%{}, fn({key, val}, acc) ->
+        updated_key = key
+        |> String.replace_leading("aws_", "")
+        |> String.to_atom
```
benwilson512: I would prefer if we avoided a `String.to_atom` here and whitelisted the acceptable transformations.

mix.exs
```diff
@@ -41,6 +41,7 @@ defmodule ExAws.Mixfile do
       {:gen_stage, "~> 0.6.0"},
       {:dialyze, "~> 0.2.0", only: :dev},
       {:bypass, "~> 0.1", only: :test},
+      {:configparser_ex, "~> 0.2.1", optional: true, only: [:dev, :test]},
```
benwilson512: We actually don't want the `only: [:dev, test]` here.

lib/ex_aws/config/defaults.ex
```diff
@@ -4,8 +4,8 @@ defmodule ExAws.Config.Defaults do
   """
 
   @common %{
-    access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :instance_role],
-    secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, :instance_role],
+    access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, {:awscli, "default", 30}, :instance_role],
+    secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, {:awscli, "default", 30}, :instance_role],
```
benwilson512: Question: is an expiration all that necessary? I don't really see people changing the file and then expecting a running application to pick it up at some specified number of seconds later.

lib/ex_aws/credentials_ini.ex
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def replace_token_key(credentials) do
+      token = Map.get(credentials, :session_token)
+
+      credentials
+      |> Map.delete(:session_token)
+      |> Map.put(:security_token, token)
+    end
+  end
```
benwilson512: We need an `else` here that has all the public functions but with a body that just says `raise "ConfigParser required to use"`

benwilson512: Otherwise if you don't have config parser there's gonna be `xref` warnings.

mix.exs
tomasz-tomczyk: Curious, why is that? Regardless, removed :)

lib/ex_aws/config/defaults.ex
tomasz-tomczyk: [Temporary credentials](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) can be set to expire after any time, so perhaps I can increase the default to a few hours (this is in seconds, right?), but I think it's better to have the option than not?

benwilson512: Is there any way that the expiration time could be determined automatically or are we stuck just having users hard code it?

tomasz-tomczyk:
@benwilson512 is there anything else I can do? Thanks for your help again :)

benwilson512:
Apologies on the delay, this looks good to go, thanks!

benwilson512:
Hey, so a bit of post-merge feedback:

- Tabs instead of spaces: Please always use spaces and not tabs
- `:awscli` added to the default config, but the required dependency was optional. This broke default configuration.
- Alarmingly small default expiration. Maybe you meant to multiply it by 1000 to get seconds, but the default expiration is 30 milliseconds. I should have caught that one, but I was also unclear (and am still unclear) about whether expiration actually makes a lot of sense for this.

tomasz-tomczyk:
Thanks @benwilson512 sorry this caused you some headache in the end! In hindsight, not sure what I was thinking with the 30ms, brain fart. Thanks for your help with this.

tomasz-tomczyk:
FWIW, tabs came from #292 :)

benwilson512:
Noted, thanks again for your help!

PR #246, opened 2016-10-06
Firehose with tests

peillis:


lib/ex_aws/config/defaults.ex
```diff
@@ -17,7 +17,7 @@ defmodule ExAws.Config.Defaults do
       region: "us-east-1",
       port: 80
     },
-    kinesis_firehose: %{
+    firehose: %{
```
benwilson512: Although it's kind of verbose I'd prefer if we kept this as `kinesis_firehose` unless we're also going to change the module from `KinesisFirehose` to `Firehose`.

peillis: Actually I had to change that in order to get it working. Looks like the service it's just called  firehose on amazon. 
I can change the module to `Firehose` if you prefer that.

benwilson512: Yes please, thanks!

peillis: ok, done

PR #249, opened 2016-10-07
Add S3.presigned_post to generate form data for browser uploads

aaronrenner:
This adds the ability to generate presigned post api for direct browser uploads to S3. This is similar to using presigned urls, but it uses a different api and allows you to use keys like "folder/${filename}" where ${filename} is replaced by S3 based on the name of the file that is uploaded.

Here are the docs for the API I am using: http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTForms.html

Let me know what you think.

### DIFF
--- a/lib/ex_aws/auth/presigned_posts/policy.ex
+++ b/lib/ex_aws/auth/presigned_posts/policy.ex
@@ -1,0 +1,34 @@
+defmodule ExAws.Auth.PresignedPosts.Policy do
+  @moduledoc false
+  import ExAws.Auth.Utils
+
+  defstruct [expiration: nil, conditions: []]
+
+  def new do
+    %__MODULE__{}
+  end
+
+  def set_expiration(policy, datetime) do
+    %__MODULE__{policy | expiration: datetime}
+  end
+
+  def add_condition(policy = %__MODULE__{conditions: conditions}, condition) do
+    new_conditions =
+      conditions
+      |> Enum.reverse
+      |> List.insert_at(0, condition)
+      |> Enum.reverse
+    %__MODULE__{policy | conditions: new_conditions}
+  end
+
+  def encode(policy = %__MODULE__{}, config) do
+    policy |> to_map |> config.json_codec.encode! |> Base.encode64
+  end
+
+  defp to_map(policy = %__MODULE__{}) do
+    %{
+      "expiration" => iso_8601_format(policy.expiration),
+      "conditions" => policy.conditions
+    }
+  end
+end
--- a/lib/ex_aws/s3.ex
+++ b/lib/ex_aws/s3.ex
@@ -889,6 +889,72 @@ defmodule ExAws.S3 do
     end
   end
 
+  @type presigned_post_opts :: [
+    {:expires_in, pos_integer()}
+    | {:current_datetime, :calendar.datetime}
+    | {:fields, [{String.t, String.t}]}
+  ]
+  @type presigned_post_response :: %{
+    url: String.t,
+    form_fields: %{optional(String.t) => String.t}
+  }
+
+  @doc """
+  Generates a presigned POST request that can be used for uploading an object.
+
+  This generates a presigned post request that can be used by the browser to
+  directly upload an object to s3. Given the following request,
+
+  ```
+  response = S3.presigned_post(config, "my-bucket", "path/to/my/file.txt")
+  ```
+
+  the browser must send a POST request with enctype of `multipart/form-data` to `response.url`.
+  This request should have `response.form_fields` as its body, plus an additional field
+  named `file` which contains the contents of the file to be uploaded.
+
+  ## Options
+
+  * `fields`: Additional form fields to be signed and included in request.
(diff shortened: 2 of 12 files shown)


benwilson512:
Hey, thanks for the PR! This definitely looks interesting. It's a lot more code that I would have expected, so I probably won't get to it today. I'll try to get to it this weekend however.

lib/ex_aws/s3.ex
```diff
@@ -889,6 +889,72 @@ defmodule ExAws.S3 do
...
+  S3.presigned_post(config, "my-bucket", "path/to/my/${filename}")
+
+  # Presigned post that expires in 5 minutes
+  S3.presigned_post(config, "my-bucket", "path/to/my/file.txt", expires_in: 60 * 5)
+  ```
+
+  For more information, see http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTForms.html
+  """
+  @spec presigned_post(
+    config :: %{},
```
benwilson512: This should be `config :: ExAws.Config.t`

lib/ex_aws/s3.ex
```diff
@@ -889,6 +889,72 @@ defmodule ExAws.S3 do
     end
   end
 
+  @type presigned_post_opts :: [
+    {:expires_in, pos_integer()}
+    | {:current_datetime, :calendar.datetime}
+    | {:fields, [{String.t, String.t}]}
```
benwilson512: Would `fields` here be more naturally a map or can there be duplicate keys?

lib/ex_aws/auth/presigned_posts/policy.ex
```diff
@@ -0,0 +1,34 @@
...
+    %__MODULE__{policy | expiration: datetime}
+  end
+
+  def add_condition(policy = %__MODULE__{conditions: conditions}, condition) do
+    new_conditions =
+      conditions
+      |> Enum.reverse
+      |> List.insert_at(0, condition)
+      |> Enum.reverse
+    %__MODULE__{policy | conditions: new_conditions}
```
aaronrenner: The policy submitted to amazon is an array, and I assume if you had 2 conflicting rules, they have logic about which one would win. I'd like to keep it so when you add rules to a policy, they come out in the same order you added them. However I do see this is not very efficient and maybe the Erlang `:queue` module would help.

luccasmaso:
+1

lib/ex_aws/auth/presigned_posts/policy.ex
benwilson512: Here's my problem with the approach here. This is an internal module; we have `@moduledoc false`. This is a relatively enormous amount of logic for what, as far as I could tell, could just be a list literal.

It is really very hard to follow what is going on. Every time I go to some function I have to jump over to 3 different modules just to figure out what's going on. It does not feel functional.

For example consider the simple line: `|> PresignedPost.set_bucket(bucket_name)`

To figure out what this does I have to read
```
def set_bucket(presigned_post, bucket_name) do
  put_policy_condition(presigned_post, %{"bucket" => bucket_name})
end
```
but this doesn't actually tell me anything so I go to `put_policy_condition`
```
def put_policy_condition(%{policy: policy} = presigned_post, condition) do
  policy = Policy.add_condition(policy, condition)
  %{presigned_post | policy: policy}
end
```
this still doesnt' tell me anything so I go to `Policy.add_condition`
```
def add_condition(policy = %__MODULE__{conditions: conditions}, condition) do
  new_conditions =
    conditions
    |> Enum.reverse
    |> List.insert_at(0, condition)
    |> Enum.reverse
  %__MODULE__{policy | conditions: new_conditions}
end
```
It's here I finally learn that we're just sticking the value at the end of a list somewhere.

As best I can tell this could all be replaced with just
```
policy = [
  ... stuff before
  %{"bucket" => bucket}
  ... stuff after
]
```

Your test cases actually illustrate this well:
```
test "adding multiple entries retains order" do
  policy =
    Policy.new
    |> Policy.add_condition(%{"acl" => "public-read"})
    |> Policy.add_condition(["starts_with", "$key", "something/"])
    |> Policy.add_condition(%{"Content-Type" => "image/jpeg"})

  assert policy.conditions == [
    %{"acl" => "public-read"},
    ["starts_with", "$key", "something/"],
    %{"Content-Type" => "image/jpeg"},
  ]
end
```
The actual `assert` stuff is shorter AND simpler than the API itself.

benwilson512: I sincerely appreciate the effort you put into to the docs and typespecs of these modules, but I feel like they obscure what's going on rather than actually helping to make it easier.

benwilson512:
Closing this for inactivity. Feel free to re-open if desired.

tajchumber:
@benwilson512 can I take over this PR? I feel like a lot of people will benefit from this :)

benwilson512:
@tajchumber you are absolutely welcome to open a PR on this subject if that's what you're asking.

PR #251, opened 2016-10-07
Add support for SNS create and list subscriptions

taylorbrooks:


benwilson512:
Thanks! Any interest in creating parsers for these btw?

taylorbrooks:
What do you mean by parsers?

benwilson512:
SQS would be a good example. https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/sqs.ex#L276 adds a parser function which is implemented here https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/sqs/parsers.ex

These serve to automatically parse the responses from AWS. This would also enable a streaming version of for example `list_subscriptions.`

PR #254, opened 2016-10-12
Add SNS verify_message

### LINKED ISSUE #248: Helper to verify SNS message
Hi,

I was wondering if an helper to verify SNS message would be within the scope of the library ? Hopefully it is :)

Doc: http://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.verify.signature.html

tlvenn:
Hi @benwilson512,

Here is a initial stab at the SNS verify_message, please let me know your comments and how I should proceed with testing. Do we want to use something like Exvcr with a fixture to properly test the whole flow ?

This resolves #248

Thanks!

### DIFF
--- a/lib/ex_aws/sns.ex
+++ b/lib/ex_aws/sns.ex
@@ -303,7 +303,8 @@ defmodule ExAws.SNS do
   defp get_string_to_sign(message_params) do
     message_params
     |> Map.take(get_params_to_sign(message_params["Type"]))
-    |> Enum.map_join(&(elem(&1,0)<>"\n"<>elem(&1,1)<>"\n"))
+    |> Enum.map(fn {key, value} -> [to_string(key), "\n", to_string(value), "\n"] end)
+    |> IO.iodata_to_binary
   end
 
   defp get_params_to_sign(message_type) do
--- a/lib/ex_aws/sns.ex
+++ b/lib/ex_aws/sns.ex
@@ -261,15 +261,14 @@ defmodule ExAws.SNS do
   @message_types ["SubscriptionConfirmation", "UnsubscribeConfirmation", "Notification"]
 
   @doc "Verify message signature"
+  @spec verify_message(message_params :: %{String.t => String.t}) :: [:ok | {:error, String.t}]
   def verify_message(message_params) do
     with :ok <- validate_message_params(message_params),
-      :ok <- validate_signature_version(message_params["SignatureVersion"]),
-      {:ok, public_key} <- ExAws.SNS.PublicKeyCache.get(message_params["SigningCertURL"]) do
-        message_params
-        |> get_string_to_sign
-        |> verify(message_params["Signature"], public_key)
-    else
-      {:error, error} -> {false, error}
+         :ok <- validate_signature_version(message_params["SignatureVersion"]),
+         {:ok, public_key} <- ExAws.SNS.PublicKeyCache.get(message_params["SigningCertURL"]) do
+      message_params
+      |> get_string_to_sign
+      |> verify(message_params["Signature"], public_key)
     end
   end
 
@@ -278,7 +277,7 @@ defmodule ExAws.SNS do
       case required_params -- Map.keys(message_params) do
         [] -> :ok
         missing_params -> {:error, "The following parameters are missing: #{inspect missing_params}"}
-      end        
+      end
     end
   end
 
@@ -288,23 +287,23 @@ defmodule ExAws.SNS do
       "SubscriptionConfirmation" -> {:ok, @confirmation_params ++ @signature_params}
       "UnsubscribeConfirmation" -> {:ok, @confirmation_params ++ @signature_params}
       type when is_binary(type) -> {:error, "Invalid Type, expected one of #{inspect @message_types}"}
-      type when is_nil(type) -> {:error, "Missing Type c (Type)"}
-      type -> {:error, "Invalid Type's type, expected a String, got #{inspect type}"}
+      type when is_nil(type) -> {:error, "Missing message type parameter (Type)"}
+      type -> {:error, "Invalid message type's type, expected a String, got #{inspect type}"}
     end
   end
 
   defp validate_signature_version(version) do
     case version do
       "1" -> :ok
-      sig when is_binary(sig) -> {:error, "Unsupported SignatureVersion, expected \"1\", got #{version}"}
-      sig -> {:error, "Invalid SignatureVersion format, expected a String, got #{version}"}
+      val when is_binary(val) -> {:error, "Unsupported SignatureVersion, expected \"1\", got #{version}"}
+      _ -> {:error, "Invalid SignatureVersion format, expected a String, got #{version}"}
     end
   end
 
   defp get_string_to_sign(message_params) do
     message_params
     |> Map.take(get_params_to_sign(message_params["Type"]))
-    |> Enum.reduce("", fn {key, value}, acc -> acc <> key <> "\n" <> value <> "\n" end)
+    |> Enum.map_join(&(elem(&1,0)<>"\n"<>elem(&1,1)<>"\n"))
   end
 
   defp get_params_to_sign(message_type) do
@@ -317,8 +316,8 @@ defmodule ExAws.SNS do
 
   defp verify(message, signature, public_key) do
     case :public_key.verify(message, :sha, Base.decode64!(signature), public_key) do
-      true -> true
-      false -> {false, "Signature is invalid"}
+      true -> :ok
+      false -> {:error, "Signature is invalid"}
     end
   end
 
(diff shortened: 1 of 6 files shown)


lib/ex_aws/sns/public_key_cache.ex
```diff
@@ -0,0 +1,59 @@
...
+      {:ok, public_key} <- get_public_key(cert) do
+        :ets.insert(__MODULE__, {certUrl, public_key})
+        {:reply, {:ok, public_key}, ets}
+    else
+      error -> {:reply, error, ets}
+    end
+  end
+
+  defp fetch_certificate(cert_url) do
+    with {:ok, 200, _resp_headers, client} <- :hackney.get(cert_url, [], "", []),
```
benwilson512: We should use `:hackney.get(cert_url, [], "", [:with_body])` because
1) it's simpler
2) Right now we only read the body on `200`. However the body doesn't go away for other status codes, and will result in basically a memory leak if not read.

lib/ex_aws/sns/public_key_cache.ex
```diff
@@ -0,0 +1,59 @@
...
+      {:ok, status_code, _resp_headers, _client} -> {:error, "Could not fetch certificate from #{cert_url}, expected http code 200, got: #{status_code}"}
+      {:error, error} -> {:error, "Unexpected error, could not fetch certificate from #{cert_url}, got #{inspect error}"}
+    end
+  end
+
+  defp get_pem_entry(pem_entries) do
+    case length(pem_entries) do
+      0 -> {:error, "invalid argument"}
+      x when x > 1 -> {:error, "found multiple PEM entries, expected only 1"}
+      x when x == 1 ->
```
benwilson512: This could more easily be:
```
case pem_entries do
  [entry] ->
    try do # etc
  entries ->
    {:error, "Invalid PEM entries: #{inspect entries}"}
end
```

lib/ex_aws/sns/public_key_cache.ex
```diff
@@ -0,0 +1,59 @@
...
+        try do
+          {:ok, :public_key.pem_entry_decode(Enum.at(pem_entries, 0))}
+        catch
+          kind, error -> {:error, "Unexpected error while decoding pem entry: #{inspect error}"}
+        end
+    end
+  end  
+  
+  defp get_public_key(cert) do
+    {:ok, :public_key.der_decode(:RSAPublicKey, cert |> elem(1) |> elem(7) |> elem(2))}
```
benwilson512: Not sure what value there is in wrapping this with `:ok`.

lib/ex_aws/sns/public_key_cache.ex
```diff
@@ -0,0 +1,59 @@
+defmodule ExAws.SNS.PublicKeyCache do
+  use GenServer
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, :ok, opts)
+  end
+
+  def get(certUrl) do
+      case :ets.lookup(__MODULE__, certUrl) do
```
benwilson512: should be `cert_url`

lib/ex_aws/sns.ex
```diff
@@ -169,6 +169,77 @@ defmodule ExAws.SNS do
...
+  @doc "Verify message signature"
+  def verify_message(message_params) do
+    with :ok <- validate_message_params(message_params),
+      :ok <- validate_signature_version(message_params["SignatureVersion"]),
+      {:ok, public_key} <- ExAws.SNS.PublicKeyCache.get(message_params["SigningCertURL"]) do
+        message_params
+        |> get_string_to_sign
+        |> verify(message_params["Signature"], public_key)
+    else
+      {:error, error} -> {false, error}
```
benwilson512: Not really a fan of `{false, error}`. What's wrong with `{:ok, msg} | {:error, reason}` as a return signature?

On that note, we need typespecs.

lib/ex_aws/sns.ex
```diff
@@ -169,6 +169,77 @@ defmodule ExAws.SNS do
...
+      "1" -> :ok
+      sig when is_binary(sig) -> {:error, "Unsupported SignatureVersion, expected \"1\", got #{version}"}
+      sig -> {:error, "Invalid SignatureVersion format, expected a String, got #{version}"}
+    end
+  end
+
+  defp get_string_to_sign(message_params) do
+    message_params
+    |> Map.take(get_params_to_sign(message_params["Type"]))
+    |> Enum.reduce("", fn {key, value}, acc -> acc <> key <> "\n" <> value <> "\n" end)
```
benwilson512: This would probably be better built up as iodata and then concatenated at the end.

lib/ex_aws/sns.ex
```diff
@@ -169,6 +169,77 @@ defmodule ExAws.SNS do
...
+        missing_params -> {:error, "The following parameters are missing: #{inspect missing_params}"}
+      end        
+    end
+  end
+
+  defp get_required_params(message_type) do
+    case message_type do
+      "Notification" -> {:ok, (@notification_params -- @optional_notification_params) ++ @signature_params}
+      "SubscriptionConfirmation" -> {:ok, @confirmation_params ++ @signature_params}
+      "UnsubscribeConfirmation" -> {:ok, @confirmation_params ++ @signature_params}
```
benwilson512: Are these types coming from amazon? Is there any reason to expect it would NOT be one of these?

lib/ex_aws/sns.ex
```diff
@@ -169,6 +169,77 @@ defmodule ExAws.SNS do
...
+  ######################
+
+  @notification_params ["Message", "MessageId", "Subject", "Timestamp", "TopicArn", "Type"]
+  @optional_notification_params ["Subject"]
+  @confirmation_params ["Message", "MessageId", "SubscribeURL", "Timestamp", "Token", "TopicArn", "Type"]
+  @signature_params ["SignatureVersion", "Signature", "SigningCertURL"]
+  @message_types ["SubscriptionConfirmation", "UnsubscribeConfirmation", "Notification"]
+
+  @doc "Verify message signature"
+  def verify_message(message_params) do
```
benwilson512: Do we want to automatically verify all messages?

lib/ex_aws/sns/public_key_cache.ex
```diff
@@ -0,0 +1,59 @@
+defmodule ExAws.SNS.PublicKeyCache do
```
benwilson512: Do these public keys ever expire?

tlvenn: SNS certificates are valid for 1year.
sns_amazonaws_com

lib/ex_aws/sns.ex
tlvenn: I simplified it with `|> Enum.map_join(&(elem(&1,0)<>"\n"<>elem(&1,1)<>"\n"))`

lib/ex_aws/sns.ex
tlvenn: Those types are indeed coming from Amazon SNS documentation. It should always be one of those.

tlvenn:
Hi @benwilson512,

Sorry for the late feedback on this issue, got side tracked pretty badly last 2 weeks...
I just pushed a new commit, hopefully addressing all your concerns.

lib/ex_aws/sns/public_key_cache.ex
tlvenn: Pretty neat, I usually don't interact with Hackney directly and did not see that option.

lib/ex_aws/sns/public_key_cache.ex
tlvenn: Yep fixed.

tlvenn:
btw shouldn't I use http_client instead of using hackney directly ?

benwilson512:
Yes yeah, we can't assume people are using hackney.

lib/ex_aws/sns.ex
benwilson512: Sorry I wasn't super clear. What I had in mind was something more like
```
|> Enum.map(fn {key, value} -> [to_string(key), "\n", to_string(value), "\n"] end)
|> IO.iodata_to_binary
```

It's more efficient to build up IOData in this case than concatenating binaries. Doesn't make a huge difference but it's also pretty simple code. I almost always prefer `{key, value}` over `elem` btw, just seems clearer.

tlvenn: Ha thanks for enlightening me on this, I did not know about iodata_to_binary. And yes I agree, pattern matching on tuples is much nicer and cleaner than trying to use shorthand function.

tlvenn:
Hi @benwilson512, the PR should be ready for final review and hopefully if all good, ready to be merged

tlvenn:
Hi @benwilson512 , when you get the time, could you please do final review and merge if ready ? Thanks !

benwilson512:
I think this looks good, thanks!

PR #258, opened 2016-10-16
Added code to check for valid SNS MessageStructure attributes and con…

jglover:
Added code to check for valid SNS MessageStructure attributes and convert their types (atom to string representation), and tests for SMS and MessageStructure values in SNS requests.


benwilson512:
The tests however are always welcome :)

test/lib/ex_aws/sns_test.exs
```diff
@@ -51,12 +51,58 @@ defmodule ExAws.SNSTest do
...
+    }
+
+    message_attributes = [
+      %{name: "AWS.SNS.SMS.SenderID", data_type: :string, value: {:string, "sender"}},
+      %{name: "AWS.SNS.SMS.MaxPrice", data_type: :string, value: {:string, "0.8"}},
+      %{name: "AWS.SNS.SMS.SMSType", data_type: :string, value: {:string, "Transactional"}}
+    ]
+    publish_opts = [
+      message_attributes: message_attributes,
+      phone_number: "+15005550006"
```
benwilson512: Do make sure that this isn't a valid phone number.

jglover:
Yeah I did wonder that myself, it also seems like a hassle to maintain rather than let AWS handle it as it will. I'll close this one and open a new one specific to the tests :)

benwilson512:
Thanks!

test/lib/ex_aws/sns_test.exs
jglover: It was harder to find a test number than you'd expect it to be ha. 
These are Twilio test credentials, they're an underlying provider for AWS. Not documented via AWS but seems to work just fine, and more importantly isn't going to spam anyone :)
https://www.twilio.com/docs/api/rest/test-credentials

PR #259, opened 2016-10-16
Add SNS publish SMS test

jglover:
Added test for SNS publish SMS attributes. Tests to support #236


benwilson512:
Awesome, is this good to go?

jglover:
¯\_(ツ)_/¯ Tests pass so if you're happy with it, sure.

PR #261, opened 2016-10-18
Export normalize_opts method to Utils module

hirocaster:
Refactor duplicates methods.


benwilson512:
Hey, thanks for the contribution!

Unfortunately I'm actually 👎 on this. The problem is that what "normalized" means for different services is different. Duplicate code _can_ be a problem, but coupling is also a problem.

hirocaster:
Thanks, speedy feedback.

> The problem is that what "normalized" means for different services is different.

I understand it.

PR #269, opened 2016-10-19
Add parser for SNS and get_endpoint_attributes and set_endpoint_attributes endpoints

jnicklas:
See #267
Closes #266

PR #270, opened 2016-10-20
Fix delete endpoint

jnicklas:
This PR is based on #269, the diff will make more sense once that is merged, but I thought I'd create this anyway so I don't forget ;)


benwilson512:
This branch now has conflicts FYI

jnicklas:
Updated!

Issue #267, opened 2016-10-19
[SNS] Implement missing API calls


jnicklas:
Sticking to the main ones which seem obviously useful as opposed to listing all the esoteric ones :)
- [x] [DeletePlatformApplication](http://docs.aws.amazon.com/sns/latest/api/API_DeletePlatformApplication.html)
- [x] [ListPlatformApplications](http://docs.aws.amazon.com/sns/latest/api/API_ListPlatformApplications.html)
- [x] [GetEndpointAttributes](http://docs.aws.amazon.com/sns/latest/api/API_GetEndpointAttributes.html)
- [x] [SetEndpointAttributes](http://docs.aws.amazon.com/sns/latest/api/API_SetEndpointAttributes.html)
- [x] [GetSubscriptionAttributes](http://docs.aws.amazon.com/sns/latest/api/API_GetSubscriptionAttributes.html)
- [x] [SetSubscriptionAttributes](http://docs.aws.amazon.com/sns/latest/api/API_SetSubscriptionAttributes.html)
- [x] [Unsubscribe](http://docs.aws.amazon.com/sns/latest/api/API_Unsubscribe.html)
- [x] [GetPlatformApplicationAttributes ](http://docs.aws.amazon.com/sns/latest/api/API_GetPlatformApplicationAttributes.html)

asgoel:
@jnicklas mind if I help out with this? using SNS mobile push pretty heavily in an app I'm building and need some of the un-implemented ones.

jnicklas:
That would be great! I haven't had time to implement these yet.

jnicklas:
That would be really great! Please do!

tisdag 25 oktober 2016 skrev Ashu Goel <notifications@github.com>:

asgoel:
Completed with #274. @jnicklas feel free to close.

jnicklas:
@asgoel thank you ❤️

benwilson512:
Thanks folks!

PR #274, opened 2016-10-25
Add missing SNS API calls

asgoel:
Adds the missing SNS API calls per #267. cc @jnicklas

### DIFF
--- a/lib/ex_aws/sns/parsers.ex
+++ b/lib/ex_aws/sns/parsers.ex
@@ -88,6 +88,37 @@ if Code.ensure_loaded?(SweetXml) do
       {:ok, Map.put(resp, :body, parsed_body)}
     end
 
+    def parse({:ok, %{body: xml}=resp}, :list_platform_applications) do
+      parsed_body = xml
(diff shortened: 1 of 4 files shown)


lib/ex_aws/sns/parsers.ex
```diff
@@ -108,6 +139,39 @@ if Code.ensure_loaded?(SweetXml) do
                           subscription_arn: ~x"./SubscriptionArn/text()"s,
                           topic_arn: ~x"./TopicArn/text()"s,
                         ],
+                        next_token: ~x"./ListSubscriptionsResult/NextToken/text()"s,
```
asgoel: @jnicklas I added this in because it looks like it was missing. I added a separate test for it as well.

benwilson512:
This looks fine to me, is it ready to go as far as features you wanted to add?

asgoel:
@benwilson yes

asgoel:
@benwilson512 sorry - meant to say that this one is done, and you can go ahead and merge if it's ok.

PR #275, opened 2016-10-28
Do not constrain to one week

jfrolich:
There is a hard coded limit to one week in the S3 signed URLs. My use case is to create cachable urls that are still protected (not completely public). The expire time needs to be high to enable browser caching, otherwise there is not really a point. I propose to remove this constraint.


jfrolich:
Ah I see, AWS does not allow this. Sorry 🙃

benwilson512:
Yup sorry!

PR #280, opened 2016-11-07
Add AWS Key Management Service (KMS)  

hirocaster:
Added new service(KMS). 
http://docs.aws.amazon.com/ja_jp/kms/latest/APIReference/Welcome.html

I needed KMS library for my products.
So, added extend your library.


benwilson512:
This is the most thorough, complete, and well documented PR I have ever received. You're awesome. Let me take a bit longer to look through everything, but this looks very good, thank you!

hirocaster:
Thank speedy reply. I appreciate your review.

test/lib/ex_aws/kms/integration_test.exs
```diff
@@ -0,0 +1,107 @@
+defmodule ExAws.KMSIntegratinTest do
+  use ExUnit.Case, async: true
+
+  defp key_id do
+    System.get_env("AWS_KEY_ARN")
+  end
```
benwilson512: So, I'm slightly wary of having this test in. Is this an environment variable that is normally set?

hirocaster: `AWS_KEY_ARN` is decided name by me. It is not defacto standard name.  
It might be good to name it for testing just like `TEST_EX_AWS_KEY_ARN`.

The normal app has a `KEY_ARN or KEY_ID` at encrypting by KMS.`KEY_ARN or KEY_ID` is set by the developer.
I guess with the standard approach, environment variables or config files.
So, I use environment variable `AWS_KEY_ARN`.

Do you think about it?

hirocaster: Change it https://github.com/CargoSense/ex_aws/pull/280/commits/a7c60c3263b4e7dc944b5d59bd4ac8886090943d

benwilson512:
Thank you so much!!!

PR #286, opened 2016-11-21
Add Route53

kagux:
Hi,

Thanks a lot for your work on this project! 
I'd like to contribute with Route53 endpoints. This PR is just a single endpoint to list hosted zone and  I will push a few more later today or tomorrow. But I'd love to get your feedback on the direction I took with this implementation.

Endpoints:
- list_hosted_zones
- create_hosted_zone
- delete_hosted_zone
- change_record_sets
- list_record_sets

### DIFF
--- a/lib/ex_aws/operation/rest_query.ex
+++ b/lib/ex_aws/operation/rest_query.ex
@@ -3,6 +3,7 @@ defmodule ExAws.Operation.RestQuery do
     http_method: nil,
     path: "/",
     params: %{},
+    body: "",
     service: nil,
     action: nil,
     parser: &ExAws.Utils.identity/2,
@@ -13,10 +14,9 @@ end
 
 defimpl ExAws.Operation, for: ExAws.Operation.RestQuery do
   def perform(operation, config) do
-    data = ""
     headers = []
     url = ExAws.Request.Url.build(operation, config)
-    ExAws.Request.request(operation.http_method, url, data, headers, config, operation.service)
+    ExAws.Request.request(operation.http_method, url, operation.body, headers, config, operation.service)
     |> operation.parser.(operation.action)
   end
 
--- a/lib/ex_aws/operation/rest_query.ex
+++ b/lib/ex_aws/operation/rest_query.ex
@@ -4,6 +4,8 @@ defmodule ExAws.Operation.RestQuery do
     path: "/",
     params: %{},
     service: nil,
+    action: nil,
+    parser: &ExAws.Utils.identity/2,
   ]
 
   @type t :: %__MODULE__{}
@@ -11,29 +13,14 @@ end
 
 defimpl ExAws.Operation, for: ExAws.Operation.RestQuery do
   def perform(operation, config) do
-    query = operation.params
-    |> URI.encode_query
-
-    headers = [
-      {"content-type", "application/x-www-form-urlencoded"},
-    ]
-
-    ExAws.Request.request(operation.http_method, url(config, operation.path), query, headers, config, operation.service)
+    data = ""
+    headers = []
+    url = ExAws.Request.Url.build(operation, config)
+    ExAws.Request.request(operation.http_method, url, data, headers, config, operation.service)
+    |> operation.parser.(operation.action)
   end
 
   def stream!(%{stream_builder: fun}, config) do
     fun.(config)
   end
-
-  def url(%{scheme: scheme, host: host, port: port}, queue_name) do
-    [
-      scheme,
-      host,
-      port |> port(),
-      queue_name
-    ] |> IO.iodata_to_binary
-  end
-
-  defp port(80), do: ""
-  defp port(p),  do: ":#{p}"
 end
--- a/lib/ex_aws/parsers.ex
+++ b/lib/ex_aws/parsers.ex
@@ -1,0 +1,18 @@
+defmodule ExAws.Parsers do
+  defmacro __using__(_opts) do
+    quote do
+      import SweetXml, only: [sigil_x: 2]
+      def parse({:error, {type, http_status_code, %{body: xml}}}, _) do
+        parsed_body = xml
+        |> SweetXml.xpath(~x"//ErrorResponse",
+                          request_id: ~x"./RequestId/text()"s,
+                          type: ~x"./Error/Type/text()"s,
+                          code: ~x"./Error/Code/text()"s,
+                          message: ~x"./Error/Message/text()"s,
+                          detail: ~x"./Error/Detail/text()"s)
+
+        {:error, {type, http_status_code, parsed_body}}
+      end
+    end
+  end
+end
--- a/lib/ex_aws/request/url.ex
+++ b/lib/ex_aws/request/url.ex
@@ -1,7 +1,5 @@
 defmodule ExAws.Request.Url do
-  @moduledoc """
-  Utility module for request URLs
-  """
+  @moduledoc false
 
   @doc """
   Builds URL for an operation and a config"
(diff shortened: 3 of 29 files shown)


benwilson512:
Hey man lots of good stuff here, thanks for the PR! I'll try to get a review in here later today.

kagux:
I've added a couple of new endpoints

lib/ex_aws/utils.ex
```diff
@@ -100,4 +100,6 @@ defmodule ExAws.Utils do
 
     greg_secs - @seconds_0_to_1970
   end
+
+  def uuid, do: DateTime.utc_now |> :erlang.phash2
```
kagux: I think it is good enough for a simple uuid. If we were to consider improving it, then using external package like https://hex.pm/packages/uuid could be a better alternative. I didn't want to bring in a package unless really needed.

kagux:
Added error response handling

lib/ex_aws/parsers.ex
```diff
@@ -0,0 +1,18 @@
...
+    quote do
+      import SweetXml, only: [sigil_x: 2]
+      def parse({:error, {type, http_status_code, %{body: xml}}}, _) do
+        parsed_body = xml
+        |> SweetXml.xpath(~x"//ErrorResponse",
+                          request_id: ~x"./RequestId/text()"s,
+                          type: ~x"./Error/Type/text()"s,
+                          code: ~x"./Error/Code/text()"s,
+                          message: ~x"./Error/Message/text()"s,
+                          detail: ~x"./Error/Detail/text()"s)
```
benwilson512: Is this correct? Every service that uses XML has this pattern? Including S3?

I'm a little reluctant to do this because AWS services are not known for being particularly consistent with respect to one another. There is some level of consistency that exists with respect to a given Operation type like Query or JSON or RestQuery but across those there is very little.

lib/ex_aws/operation/rest_query.ex
```diff
@@ -3,37 +3,24 @@ defmodule ExAws.Operation.RestQuery do
...
   def perform(operation, config) do
-    query = operation.params
-    |> URI.encode_query
-
-    headers = [
-      {"content-type", "application/x-www-form-urlencoded"},
-    ]
-
-    ExAws.Request.request(operation.http_method, url(config, operation.path), query, headers, config, operation.service)
+    headers = []
```
benwilson512: Why are we not setting the header here now?

lib/ex_aws/utils.ex
```diff
@@ -100,4 +100,15 @@ defmodule ExAws.Utils do
...
+
+  def rename_keys(keywords, mapping) do
+    keywords |> Enum.map(fn {k, v} ->
+     cond do
+       Keyword.has_key?(mapping, k) -> {mapping[k], v}
+       true -> {k, v}
+     end
+    end)
+  end
```
benwilson512: I think this would be better as:
```
def rename_keys(params, mapping) do
  mapping = Map.new(mapping)
  Enum.map(params, fn {k, v} ->
    {Map.get(mapping, k, k), v}
  end)
end
```

lib/ex_aws/request/url.ex
```diff
@@ -0,0 +1,23 @@
...
+    |> (&struct(URI, &1)).()
+    |> URI.to_string
+  end 
+
+  defp normalize_scheme(%{scheme: scheme} = url) do
+    scheme = scheme |> String.replace("://", "")
+    url |> Map.put(:scheme, scheme)
+  end
+  defp normalize_scheme(url), do: url
+end
```
benwilson512: I appreciate moving this into its own module, I think there's a lot of potential here for cleaning up code from other operations. However if we're gonna have a module named `ExAws.Request.URL` we should make sure that it works for all services. If it does not, it belongs under the operation namespace not the request namespace.

Could you try and see if you can use this with all operations?

lib/ex_aws/parsers.ex
kagux: This is correct for `Route53`, `SNS` and `SQS`. Error response formats are identical. Relevant links:

- http://docs.aws.amazon.com/Route53/latest/APIReference/requests-rest-responses.html
-http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/UnderstandingResponses.html

I didn't find reference for SNS, but it is exactly the same and my initial trigger to extract this were identical test examples. 
How would you suggest renaming this module?

lib/ex_aws/operation/rest_query.ex
kagux: it became more complicated as headers vary depending on request method and body type. It works works for all current use cases without headers. So I decided to skip it to avoid bloating PR and make a separate PR to set correct headers.

lib/ex_aws/request/url.ex
kagux: Sure, I'll replace URL manipulation with this module for all operations.

lib/ex_aws/utils.ex
kagux: nice one, thanks, I'll update!

kagux:
I've addressed all your comments except for `headers` list. I think it shouldn't be done within the scope of current PR. 
To set correct headers we either have to pass content type to operation or set it based on combination of request method and body content.

lib/ex_aws/parsers.ex
benwilson512: Fair enough. I think it should likely go under `ExAws.Operation.Query.Parser` because the parsing it does seems specific to that particular operation format. It does not relate to either S3 or say Dynamo.

benwilson512:
There's one other relatively minor issue I'm noticing: Route 53 isn't actually a Rest Query operation, it's a Rest XML operation. This explains why some of the header stuff wasn't making sense. While we don't actually have a Rest XML operation yet, it should be pretty easy to build by just cloning Rest Query version. This lets us set the headers on the rest query back to what they were.

kagux:
it depends on how you define each operation. To be honest, I don't think that splitting would solve the problem. In original rest query implementation, header was always set to "application/x-www-form-urlencoded", which technically wasn't correct.  Adding XML POST body exposed this issue.  
Splitting out XML would introduce quite some duplication and would solve headers issue only for one particular case. 

 If you insist, I'll clone a new operation. I do think that splitting header logic from operation or making content type part of operation would be a  better long term solution. 

How do you want me to proceed?

kagux:
Updated `ExAws.Operation.Query.Parser` module. Headers should be the only thing left pending in this PR.

kagux:
@benwilson512, how do you want to proceed?

kagux:
I've added new endpoint `get_record_sets` and rebased to remove merge conflicts on mix.lock

kagux:
My last change is related to this issue in `sweet_xml` https://github.com/awetzel/sweet_xml/pull/20. Text nodes with `&quot;` are split in `xmerl_scan`. They fixed it in `sweet_xml` for a single node. In our case, we want a list of strings and that ends up splitting some nodes into multiple.
Amazon wraps TXT zone records in quotes and as a result triggers this issue. I couldn't figure any other way around this issue but decoding quotes.

kagux:
hey @benwilson512, 
I was wondering if we could wrap up with this PR. We've been using PR-fork in production for last month with no issues.
Please, let me know if there's still a blocking issue in current PR state.

lib/ex_aws/utils.ex
benwilson512: I'm gonna switch this to `:erlang.unique_integer([:positive])` since concurrent requests could easily call this function at the same time.

benwilson512: Actually never mind, those can repeat over instances. I suppose concurrently creating hosted zones is unlikely so we'll stick with the current implementation.

PR #289, opened 2016-11-23
Support loading of credentials from the aws cli file

hamstah:
Hi,
This PR allows loading of credentials from ~/.aws/credentials
This is useful when using assumed roles that change often.
The new keys are stored in the credentials file under a specific profile.

```elixir
config :ex_aws,
  profile: :"test-assumed",
  access_key_id: [:credentials_file],
  secret_access_key: [:credentials_file],
```

I didn't want to change too much how the config works, but ideally specifying `profile` should be enough to not have to provide the other keys, but as they default to `:instance_role` we need to overwrite it.

Thanks


benwilson512:
Hey there thanks for the PR!

However, a comprehensive PR that addressed this issue was just merged into master: https://github.com/CargoSense/ex_aws/pull/245

hamstah:
Oh, bad timing :)
Didn't check the pending PR when I wrote mine.

PR #290, opened 2016-11-23
Added module for client side KMS encryption

hamstah:
Hi
The new module makes it easy to use KMS for client side encryption/decryption

```elixir
message = "waffles"
assert ExAws.KMS.ClientEncryption.encrypt(message, key_id) |> ExAws.KMS.ClientEncryption.decrypt == message
```

Any suggestion on how to write tests given the dependency on a valid master key?

Thanks

### DIFF
--- a/lib/ex_aws/kms_client_encryption.ex
+++ b/lib/ex_aws/kms_client_encryption.ex
@@ -5,6 +5,23 @@ end
 
 defmodule ExAws.KMS.ClientEncryption do
 
+  @doc "Encrypts a map as json with a data key generated for the master key_id provided"
+  @spec encrypt(data :: Map.t(), key_id :: String.t()) :: ExAws.KMS.ClientEncryption.Result
+  def encrypt_map(%{} = data, key_id) do
+    config = ExAws.Config.new(:kms)
+    json = config[:json_codec].encode!(data)
+    encrypt(json, key_id)
+  end
+
+  @doc "Decrypts the result of a call to encrypt_map() (Note: keys will be strings)"
+  @spec decrypt(message :: ExAws.KMS.ClientEncryption.Result) :: Map.t()
+  def decrypt_map(message) do
+    plaintext = decrypt(message)
+    config = ExAws.Config.new(:kms)
+    {:ok, obj} = config[:json_codec].decode(plaintext)
+    obj
+  end
+
   @doc "Encrypts plaintext with a data key generated for the master key_id provided"
   @spec encrypt(plaintext :: String.t(), key_id :: String.t()) :: ExAws.KMS.ClientEncryption.Result
   def encrypt(plaintext, key_id) do
--- a/lib/ex_aws/kms_client_encryption.ex
+++ b/lib/ex_aws/kms_client_encryption.ex
@@ -1,0 +1,35 @@
+defmodule ExAws.KMS.ClientEncryption.Result do
+  @enforce_keys [:encrypted_data_key, :ciphertext]
+  defstruct [:encrypted_data_key, :ciphertext]
+end
+
+defmodule ExAws.KMS.ClientEncryption do
+
+  @doc "Encrypts plaintext with a data key generated for the master key_id provided"
+  @spec encrypt(plaintext :: String.t(), key_id :: String.t()) :: ExAws.KMS.ClientEncryption.Result
+  def encrypt(plaintext, key_id) do
+    case  ExAws.KMS.generate_data_key(key_id, [{:key_spec, "AES_256"}]) |> ExAws.request do
+      {:ok, %{"CiphertextBlob" => encrypted_data_key, "KeyId" => _, "Plaintext" => data_key}} ->
+
+	iv    = :crypto.strong_rand_bytes(16)
+	state = :crypto.stream_init(:aes_ctr, :base64.decode(data_key), iv)
+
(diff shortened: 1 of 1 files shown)


hamstah:
Hey guys,
Any feedback on this PR?
Thanks

benwilson512:
Would it make sense to just call the module `ExAws.KMS.Client` ?

hamstah:
I guess it would.
I used "ClientEncryption" as opposed to server side encryption, which is something a KMS client would use too. (server side encryption is done with the encrypt call, not generate data key then encrypt client side).

I'd say ExAws.KMS is already the client.

Nico

hamstah:
Hey @benwilson512 
any other feedback?

benwilson512:
Nope, this seems good, thanks!

benwilson512:
Be advised, this PR had tabs instead of spaces in some place. Please always use spaces instead of tabs

benwilson512:
Actually sorry I had to revert on further investigation, making more review comments.

lib/ex_aws/kms_client_encryption.ex
```diff
@@ -0,0 +1,52 @@
...
+    config = ExAws.Config.new(:kms)
+    {:ok, obj} = config[:json_codec].decode(plaintext)
+    obj
+  end
+
+  @doc "Encrypts plaintext with a data key generated for the master key_id provided"
+  @spec encrypt(plaintext :: String.t(), key_id :: String.t()) :: ExAws.KMS.ClientEncryption.Result
+  def encrypt(plaintext, key_id) do
+    case  ExAws.KMS.generate_data_key(key_id, [{:key_spec, "AES_256"}]) |> ExAws.request do
+      {:ok, %{"CiphertextBlob" => encrypted_data_key, "KeyId" => _, "Plaintext" => data_key}} ->
```
benwilson512: If you're only going to handle the `{:ok` case then you should just use `ExAws.request!` and this function should be `def encrypt!`. The same goes for the `decrypt` function.

Ideally I think you take a different approach that looks like the following:
```
key_id
|> ExAws.KMS.generate_data_key(key_spec: "AES_256")
|> ExAws.request!
|> ExAws.KMS.ClientEncryption.encrypt(plain_text)
```

One of the guiding principles behind ExAws is the separation of pure logic from actual requests to AWS. This module in its current state violates those principles because calling these functions has hard coded requests to AWS. This makes it increasingly difficult to support supplying optional config to ExAws.request!, as well as choosing between the `!` and non `!` versions.

By making the ClientEncryption stuff just pure logic, we give all that power back to the users.

benwilson512: Also, why is `"AES_256"` hard coded? Is this specified by AWS as the only supported encryption scheme or should we be supporting others?

lib/ex_aws/kms_client_encryption.ex
```diff
@@ -0,0 +1,52 @@
...
+
+	iv    = :crypto.strong_rand_bytes(16)
+	state = :crypto.stream_init(:aes_ctr, :base64.decode(data_key), iv)
+
+	{_state, ciphertext} = :crypto.stream_encrypt(state, to_string(plaintext))
+	%ExAws.KMS.ClientEncryption.Result{encrypted_data_key: encrypted_data_key, ciphertext: :base64.encode(iv <> ciphertext)}
```
benwilson512: Why are we concatenating `iv` and `cyphertext`?

lib/ex_aws/kms_client_encryption.ex
```diff
@@ -0,0 +1,52 @@
...
+  @doc "Encrypts plaintext with a data key generated for the master key_id provided"
+  @spec encrypt(plaintext :: String.t(), key_id :: String.t()) :: ExAws.KMS.ClientEncryption.Result
+  def encrypt(plaintext, key_id) do
+    case  ExAws.KMS.generate_data_key(key_id, [{:key_spec, "AES_256"}]) |> ExAws.request do
+      {:ok, %{"CiphertextBlob" => encrypted_data_key, "KeyId" => _, "Plaintext" => data_key}} ->
+
+	iv    = :crypto.strong_rand_bytes(16)
+	state = :crypto.stream_init(:aes_ctr, :base64.decode(data_key), iv)
+
+	{_state, ciphertext} = :crypto.stream_encrypt(state, to_string(plaintext))
```
benwilson512: Why are we using the stream encryption functions when not actually stream encrypting?

lib/ex_aws/kms_client_encryption.ex
benwilson512: Also, is base64 encoding some kind of standard here or just a preference? Might someone choose base16 encoding?

lib/ex_aws/kms_client_encryption.ex
```diff
@@ -0,0 +1,52 @@
+defmodule ExAws.KMS.ClientEncryption.Result do
+  @enforce_keys [:encrypted_data_key, :ciphertext]
+  defstruct [:encrypted_data_key, :ciphertext]
+end
+
+defmodule ExAws.KMS.ClientEncryption do
```
benwilson512: This module needs documentation showing what example usage is like. I'm honestly not sure what role this module fills.

benwilson512:
Please re-open this PR.

hamstah:
Thanks for the feedback Ben, I will re-open the PR once I've addressed your questions

PR #292, opened 2016-11-24
Awscli credentials

hamstah:
Hi,
I've noticed a few issues while trying to use the new INI file implementation.

Thanks

Nico

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -65,10 +65,12 @@ defmodule ExAws.Config do
     config
     |> ExAws.Config.AuthCache.get
     |> Map.take([:access_key_id, :secret_access_key, :security_token])
+    |> valid_map_or_nil
   end
   def retrieve_runtime_value({:awscli, profile, expiration}, _) do
     ExAws.Config.AuthCache.get(profile, expiration)
     |> Map.take([:access_key_id, :secret_access_key, :region, :security_token])
+    |> valid_map_or_nil
   end
   def retrieve_runtime_value(values, config) when is_list(values) do
     values
@@ -87,5 +89,8 @@ defmodule ExAws.Config do
     end
   end
   def parse_host_for_region(config), do: config
+
+  defp valid_map_or_nil(map) when map == %{}, do: nil
+  defp valid_map_or_nil(map), do: map
 
 end
--- a/lib/ex_aws/credentials_ini.ex
+++ b/lib/ex_aws/credentials_ini.ex
@@ -18,9 +18,10 @@ if Code.ensure_loaded?(ConfigParser) do
       |> ConfigParser.parse_string
 
       case parsed_contents do
-        {:ok, map} ->
-          Map.get(map, profile_name)
+        {:ok, %{^profile_name => config}} ->
+	  config
           |> strip_key_prefix
+	{:ok, %{}} -> raise "Missing #{profile_name} entry in configuration"
         _ -> %{}
       end
     end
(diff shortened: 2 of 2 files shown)


hamstah:
@tomasz-tomczyk Thought this might be relevant to you.

benwilson512:
Can you describe what these issues are?

hamstah:
Hi,
The issues are described in the commits, but basically
- The code assumes the profile is present in both config files
- The lookup in the `.aws/config` will not work as it's using the wrong section name
- The code merges the 2 config with the wrong precedence

tomasz-tomczyk:
At a glance, this is reasonable; I should have checked the naming for the profiles - did not pick up on the prefix ([source](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-multiple-profiles)).

Is the config file not generated unless you provide the region?

Looks good!

hamstah:
Thanks for the review @tomasz-tomczyk 

Actually I double checked and it seems that the tool will generate an empty entry in the profile file if you don't specify the region/output. 

I have some other tools that update the shared credentials but not the config, but if the default aws behaviour is to update both, we can assume that profile is present in both.

I will amend the PR to raise an error if the profile isn't in both.
I will also add back the merge with fixed precedence instead of ignoring the config file to match that.

Thanks

hamstah:
Updated the commit as discussed.

hamstah:
Hi,
Could you please have a look at the new commits?

Without those fixes, the default configuration will never resolve correctly with `:instance_role` because `CredentialsIni `always returns a map (with only the `security_token` key) if the configuration is not found and this map gets used as configuration without falling back to `ec2_instance`.

Please let me know if you have more feedback I need to address to get this merged in.

Thanks

lib/ex_aws/credentials_ini.ex
```diff
@@ -39,11 +33,30 @@ if Code.ensure_loaded?(ConfigParser) do
 
     def replace_token_key(credentials) do
       token = Map.get(credentials, :session_token)
+      cond do
+	not is_nil(token) ->
+	  credentials
+	  |> Map.delete(:session_token)
+	  |> Map.put(:security_token, token)
+	true -> credentials
```
benwilson512: You could replace lines 35 through 42 with:
```
case Map.pop(credentials, :session_token) do
  {nil, credentials} ->
    credentials
  {token, credentials} ->
    Map.put(credentials, :security_token, token)
end
```

lib/ex_aws/config.ex
```diff
@@ -88,4 +90,11 @@ defmodule ExAws.Config do
   end
   def parse_host_for_region(config), do: config
 
+  defp valid_map_or_nil(arg) do
+    cond do
+      map_size(arg) == 0 -> nil
+      true -> arg
+    end
+  end
```
benwilson512: This would be better as:
```
defp valid_map_or_nil(map) when map == %{}, do: nil
defp valid_map_or_nil(map), do: map
```

lib/ex_aws/config.ex
```diff
@@ -65,10 +65,12 @@ defmodule ExAws.Config do
     config
     |> ExAws.Config.AuthCache.get
     |> Map.take([:access_key_id, :secret_access_key, :security_token])
+    |> valid_map_or_nil
   end
   def retrieve_runtime_value({:awscli, profile, expiration}, _) do
-    ExAws.Config.AuthCache.get(profile, expiration)
+    result = ExAws.Config.AuthCache.get(profile, expiration)
```
benwilson512: This variable looks unused to me.

hamstah:
Thanks @benwilson512 I've updated the commits as you suggested.

benwilson512:
Thanks!

hamstah:
Thanks @benwilson512 !

PR #293, opened 2016-11-27
Use ini file configuration without session token

kanmo:
Thanks to [this pull request](https://github.com/CargoSense/ex_aws/pull/245)  for use ini file configuration.

aws cli configuration without session token cannot use this feature.

For example, profile without session token(we use this when simply change account)

```
[profile]
aws_access_key_id=yourawsaccesskey
aws_secret_access_key=yourawssecretaccesskey
```

This pull request is a change to enable load ini file configuration without session token.


benwilson512:
This is being handled in https://github.com/CargoSense/ex_aws/pull/292

benwilson512:
Thanks for catching it though!

PR #296, opened 2016-11-29
Add Amazon STS service

whw:


lib/ex_aws/sts.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  @type get_federation_token_params :: %{
+    :duration => pos_integer(),
+    :name => binary,
+    :policy => policy
+  }
+
+  @doc "Get Federation Token"
+  @spec get_federation_token(params :: get_federation_token_params) :: ExAws.Operation.Query.t
+  def get_federation_token(%{duration: duration, name: name, policy: policy}) do
```
benwilson512: Mandatory parameters should be individual arguments, options should be a keyword list of options.

lib/ex_aws/sts/parsers.ex
```diff
@@ -0,0 +1,39 @@
...
+      {:error, {type, http_status_code, parsed_body}}
+    end
+
+    def parse(val, _), do: val
+
+    defp request_id_xpath do
+      ~x"./ResponseMetadata/RequestId/text()"s
+    end
+  end
+else
```
benwilson512: Thanks for including parsers!

whw:
@benwilson512 thanks for the comments! I'll push a fix right quick.

I only really needed this small slice of the STS API for now, so I was hoping that I could plant this nugget and myself or someone else could extend it in the future.

lib/ex_aws/sts.ex
whw: So to be clear, in this case duration and policy are optional, so the signature should be:

`get_federation_token(name, [duration: duration, policy: policy])`

benwilson512: Yup! From  typespec perspective that comes out to
```
@type get_federation_token_opt :: {:duration, pos_integer} | {:policy, policy}
@spec get_federation_token(name :: String.t, [get_federation_token_opt]) :: ExAws.Operation.Query.t
```

PR #299, opened 2016-12-05
implement dynamodb streams

kyleaa:
Contains an implementation of the DynamoDB Streams API based on http://docs.aws.amazon.com/dynamodbstreams/latest/APIReference/API_Operations.html. There are integration tests included that run against local dynamodb.

### DIFF
--- a/lib/ex_aws/dynamo_streams.ex
+++ b/lib/ex_aws/dynamo_streams.ex
@@ -1,0 +1,125 @@
+defmodule ExAws.DynamoStreams do
+  @moduledoc """
+  Operations on DynamoDB Streams
+
+  http://docs.aws.amazon.com/dynamodbstreams/latest/APIReference/API_Operations.html
+
+  NOTE: When Mix.env in [:test, :dev] dynamo clients will run by default against
+  Dynamodb local.
+
+  Enabling/Disabling streams on a table is performed through the `Dynamo.update_table` operation.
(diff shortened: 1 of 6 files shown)


lib/ex_aws/dynamo_streams.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  @type stream_arn :: binary
+
+  @doc """
+  Lists all of the streams associated with an account.
+  If you have multiple streams on a table (created by disabling/enabling the stream)
+  you may have difficulty identifying the active stream from this operation.
+  """
+  @spec list_streams() :: ExAws.Operation.JSON.t
+  def list_streams() do
```
benwilson512: The optional parameters should be supported from: http://docs.aws.amazon.com/dynamodbstreams/latest/APIReference/API_ListStreams.html

kyleaa:
Added! Thanks!

benwilson512:
Good stuff, thank you!

benwilson512:
Hey so, did you try actually running this against AWS? Because based on the testing I'm doing it's got signature errors.

benwilson512:
This has been reverted as I couldn't easily make it work. I confess I'm a bit annoyed that this was submitted without ever having tested it against the actual AWS service. Merging it to master in that state is however my fault. Please include some tests that actually hit AWS like `list_streams` and when those pass we can try merging this again.

kyleaa:
I sincerely apologize - I am upset with myself as well for this oversight. Thank you for your patience and the time you put into this project!

PR #307, opened 2016-12-13
Dynamo streams implementation

kyleaa:
Attempt 2, this time it actually works 🙁.

Dynamo Streams uses the "dynamodb" service name for the signing of all of the operations, but requires a different URL. Rather than trying to hack something into the `:dynamodb` config key I added a `:service_override` config option... it's not super elegant, I'm certainly open to suggestions on improving this.

I've added a `list_streams` call that hits live aws from the integration test and have tested the remaining operations against real streams now.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -12,6 +12,12 @@ config :ex_aws, :dynamodb,
   port: 8000,
   region: "us-east-1"
 
+config :ex_aws, :dynamodb_streams,
+  scheme: "http://",
+  host: "localhost",
+  port: 8000,
+  region: "us-east-1"
+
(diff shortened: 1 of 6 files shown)


nerdyworm:
I took this branch out for a test drive, seems to be working splendidly.   Thanks!

benwilson512:
Can you include a test that uses config pointing it to the actual AWS dynamodb streams service?

kyleaa:
I added one in this PR, its on line 29 of integration_test calling `list_streams` in us-east-1

benwilson512:
My bad, thanks!

PR #309, opened 2016-12-14
Remove configparser_ex requirement in ec2 environments

kyleaa:
Looking at the discussion on #245 I think the intention was for configparser_ex to be optional, but with the `raise` blocks and the addition of the `:awscli` to the default config resolution list you'll always get an error before attempting to retrieve ec2 credentials.

also fixed an unused variable warning.

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -4,8 +4,8 @@ defmodule ExAws.Config.Defaults do
   """
 
   @common %{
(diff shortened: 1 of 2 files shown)


benwilson512:
I'm inclined to think that the real issue is that the defaults here https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/config/defaults.ex#L8 include `{:awscli`. If I want to keep the configparser dependency as optional, I think the `{:awscli` option ought to not be a default.

If someone wants to use it they can include the dependency and add it to the config.

Thoughts?

kyleaa:
That definitely sounds good to me. I'll change this up.

benwilson512:
Thanks!

PR #313, opened 2016-12-19
WIP Feature: Cloudformation support

joffotron:
Hi!

I've started implementing API support for Cloudformation in ex_aws (as I need a couple of functions for the project I'm on).

At the moment I have http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackResource.html and http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ListStackResources.html

to (I think) a usable point. 

At this stage - wanting feedback as to whether what I've built is going in the right direction for eventual inclusion, once the rest of the operations have been implemented?


test/lib/ex_aws/cloudformation_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule ExAws.CloudformationTest do
```
benwilson512: You need a test that actually hits the CloudFormation API, generally a basic list action.

lib/ex_aws/cloudformation/parsers.ex
```diff
@@ -0,0 +1,60 @@
...
+    defp request_id_xpath do
+      ~x"./ResponseMetadata/RequestId/text()"s
+    end
+
+    defp const_to_atom(string) do
+      string |> String.downcase |> String.to_atom
+    end
+
+    defp parse_metadata_json(json) do
+      config = ExAws.Config.new(:cloudformation)
```
benwilson512: So unfortunately this will not work, because the config needs to match whatever was used to make the request. If someone does `|> ExAws.request(json_codec: :foo)` that needs to be honored.

benwilson512: This will require modifying https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/operation/query.ex#L32 to pass in the `config`. 

Probably the least problematic way to do this would be:
```
result = ExAws.Request.request(:post, url(config, operation.path), query, headers, config, operation.service)
parser = operation.parser
cond do
  is_function(parser, 2) ->
    parser.(result, operation.action)
  is_function(parser, 3) ->
    parser.(result, operation.action, config)
  true ->
    result
end
```

At that point you can have `def parse({:ok, %{body: xml}=resp}, :describe_stack_resource, config) do` and pass that config on to your json parsing function.

lib/ex_aws/cloudformation/parsers.ex
```diff
@@ -0,0 +1,60 @@
...
+        resource_type: ~x"./ResourceType/text()"s
+      ]
+    end
+
+    defp request_id_xpath do
+      ~x"./ResponseMetadata/RequestId/text()"s
+    end
+
+    defp const_to_atom(string) do
+      string |> String.downcase |> String.to_atom
```
benwilson512: `String.to_atom` is a generally pretty dangerous function. If the set of possible values `ResourceStatus` has is bounded, make a list of those possibilities in a typespec or something somewhere so that you can use `String.to_existing_atom`

joffotron:
Thanks @benwilson512 for the feedback! Very useful :-)

I think the config thing you mentioned has worked alright, and added list_stacks as a simple call we can make in an integration test without someone needing existing stacks set up.

This has highlighted some difficulty I had with `.to_existing_atom` and using typespecs to pre-seed the atom table.

Only way I can make tests pass at the moment is to have a function called at runtime that loads the set of atoms before we use them.

Having them in the @type spec doesn't seem to be sufficient :-( I know these get set at compile-time, but I'm fairly new at Elixir, so I'm probably fundamentally misunderstanding something here.

The other approach would be to pass these straight through as the strings we get from AWS. Thoughts? 

(The same question arises about the Date/Time types - should we be turning these into something more native, or leaving that up to the consumer to handle?)

benwilson512:
This is looking good! I'm somewhat leaning towards just pushing the status values through as strings instead of atoms. So far we generally just have things like keys as atoms, values are generally left as strings.

joffotron:
Thanks. Yes, I'll change the status values to strings :-)

Still have a fair few actions to implement, but I'll slowly finish them off!

benwilson512:
@joffotron I'm inclined to just merge this as is, and then you can expand on it as necessary. Thoughts?

joffotron:
@benwilson512 Yeah, I've not had much time lately to work on this due to Real Life, so perhaps we can merge what's here for now (maybe with a note about incomplete support for the API).

The Cfn API is a pretty big one, it'll take a while to fully build out. Wondering if there's a better way? ;)

But yeah, let's put it in, and we can slowly chip away at it as we can

andrevdm:
@benwilson512 @joffotron, I see this is approved but not yet merged, is that planned to happen soon?

I want to add [DescribeStacks](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStacks.html) as I need the stack's output values. 

I'll fork this change and add it there so long.

Thanks
Andre

PR #316, opened 2016-12-22
relax hackney dependency

CrowdHailer:
Is this possible? I am getting dependency conflicts when using ex_aws in the same project as httpoison


benwilson512:
Duplicate of https://github.com/CargoSense/ex_aws/pull/316. TL;DR: No, 1.6.3 has known issues. use `override: true` if you want to force it to one version or another.

CrowdHailer:
Ok. 

BTW this is issue 316 so the duplicate link isn't pointing anywhere useful.

benwilson512:
Fixed.

PR #317, opened 2016-12-22
helpful error for missing secret_access_key

CrowdHailer:
It took me a while to debug that I had a typo when setting my environment variables. This would have been very helpful


benwilson512:
Hey thanks for prompting me to handle this! I wanted to handle it in a more generic way for both of the keys that are required, as well as different possible situations. It's handled on master now, thanks again!

PR #319, opened 2016-12-22
Add missing regions from S3 config

Gazler:
The values from http://docs.aws.amazon.com/general/latest/gr/rande.html
were used.


benwilson512:
Thanks!

PR #328, opened 2017-01-13
Properly format SNS MessageAttributes

danielberkompas:
AWS expects MessageAttributes, as a list, to be formatted like so:

```
"MessageAttribute.entry.1.Name" => "value"
```

SNS currently formats attributes without "entry" and with 0 indexing:

```
"MessageAttribute.0.Name" => "value"
```

This results in the following error when you try to publish a message
with MessageAttributes:

```
{:error,
 {:http_error, 400,
  %{code: "MalformedInput", detail: "",
    message: "Top level element may not be treated as a list",
    request_id: "3fc49509-919d-57a3-8147-c5faf4348859", type: "Sender"}}}
```

This has already been fixed elsewhere. See the following examples:

- https://github.com/CargoSense/ex_aws/blob/7082f43acabf283d2e98673ee080e9122c737b97/test/lib/ex_aws/sns_test.exs#L40-L41
- https://github.com/CargoSense/ex_aws/blob/7082f43acabf283d2e98673ee080e9122c737b97/test/lib/ex_aws/sns_test.exs#L208-L213

### DIFF
--- a/lib/ex_aws/sns.ex
+++ b/lib/ex_aws/sns.ex
@@ -104,7 +104,7 @@ defmodule ExAws.SNS do
   end
 
   def build_message_attribute({%{name: name, data_type: data_type, value: {value_type, value}}, i}, params) do
-    param_root = "MessageAttribute.#{i}"
+    param_root = "MessageAttribute.entry.#{i + 1}"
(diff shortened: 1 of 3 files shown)


benwilson512:
Thank you!

Issue #337, opened 2017-01-18
Flow now is separated module


ssbb:
Flow now is not part of `gen_stage` package and available at https://github.com/elixir-lang/flow

benwilson512:
Yeah, honestly Flow could just be replaced with `Task.async_stream` although that would mean I'd have to bump the Elixir requirement to 1.4

benwilson512:
In master we now just use Task.async_stream. releasing today.

PR #339, opened 2017-01-20
switch from gen stage to newly spun-off flow library

raorao:
see: https://elixirforum.com/t/the-future-of-genstage-and-flow/3323 for context

fix for #337


benwilson512:
Hey! I appreciate the PR.

As I allude to in that issue however, I'm now thinking I would rather just use `Task.async_stream` and remove the need for `Flow` altogether. The usage of Flow in this library is incredibly simple, and it doesn't need anything more than what the 1.4 standard library offers.

If you'd like to make a PR to that affect you're more than welcome to!

raorao:
Would you consider merging the PR as a half measure? Happy to own writing a
second PR for async_stream, but current state of the world is preventing my
project from migrating to GenStage 0.11.

benwilson512:
>  but current state of the world is preventing my
project from migrating to GenStage 0.11

I'm not sure I understand. You can use `override: true` on the gen stage entry of your mix.exs file to force it to whatever version you like. This particular S3 function won't work sure, but everything else will. If you are using this S3 function you're best off just using your fork. I'm planning on removing the GenStage dependency prior to the next release of this library anyway.

raorao:
hm, I couldn't get `override: true` to work, since the references to Experimental.Flow prevent the entire ex_aws application from compiling. I'll take a shot at switching ex_aws to Task.async_stream (and also migrating the entire project to 1.4), but I would still a prefer a half-measure.

benwilson512:
Re-opening this for the moment because it's possible Flow might provide superior throughput to Task.async stream, investigating.

benwilson512:
In the end Task.async_stream can be just as fast in this case.

PR #341, opened 2017-01-23
Added some examples for S3.list_objects() including more params.

dmilith:


lib/ex_aws/s3.ex
```diff
@@ -155,6 +155,17 @@ defmodule ExAws.S3 do
...
+
+  ## Examples
+  ```
+  S3.list_objects("my-bucket") |> ExAws.stream!
+
+  S3.list_objects("my-bucket", delimiter: "/", prefix: "backup") |> ExAws.stream!
+
+  S3.list_objects("my-bucket", prefix: "some/inner/location/path") |> ExAws.stream!
+
+  S3.list_objects("my-bucket", max_keys: 5, encoding_type: "url") |> ExAws.stream!
```
benwilson512: Honestly I'd prefer if we only did the extra examples here. There isn't any particular need to include them in the readme as well as the main ExAws module doc. Continuing this pattern to other services would lead to entirely too many examples in the readme.

benwilson512:
Fixed in master.

PR #343, opened 2017-01-24
fix type mismatch in dynamo.ex.

melpon:
These types should be used `map()`.

[Map specification](https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax):

> Notice that the syntactic representation of map() is %{optional(any)
> => any}, not %{}. The notation %{} specifies the singleton type for
> the empty map.


benwilson512:
Thanks!

PR #347, opened 2017-01-30
Add support for AWS Polly service

Gringill:
https://aws.amazon.com/polly/

Added support for all Polly operations
http://docs.aws.amazon.com/polly/latest/dg/API_Operations.html

Added integration and operation-struct-building tests.

Let me know what else I need to do to make this happen :).


lib/ex_aws/operation/polly.ex
```diff
@@ -0,0 +1,57 @@
+defmodule ExAws.Operation.Polly do
```
benwilson512: Operations are not per service, but rather per category of service. Polly is a `RestJSON` service. There is not at the moment a RestJSON operation although looking at what we have here I'm pretty sure you could just re-use the JSON operation.

All of the path logic you have that is custom to polly you could just do in the `defp request` function you have inside `ExAws.Polly`

Gringill:
Hey Ben thanks for looking over this, I appreciate it. Your suggested changes sound good 👍🏼 .

I actually originally went down the path of using the JSON operation but felt I was unable to because of the `synthesize_speech` method, the body it returns is binary and as such the json decode that is attempted in the JSON operation's `parse` function fails. Do you have any suggestions regarding that?

Additionally, the `synthesize_speech` method accepts a param `LexiconNames` that is an array of strings which, when passed, causes us to bump into the issue of `encode_query/1` not being able to encode values that are lists.

Gringill:
I addressed your concerns about the `ExAws.Polly` module.

Take a look at the JSON operation now. Due to the issue of `synthesize_speech` being the only method that should not be json decoded I took a stab at allowing the parser used to be optionally overridden by the operations parser field. Let me know if you would prefer a different approach.

I now realize that correctly using the JSON operation (as opposed to the since deleted polly operation) causes an actual json body to be sent rather than just encoding the params into the url, which causes the aforementioned `encode_query/1` issue to be side-stepped, hooray.

Gringill:
Hey @benwilson512 when you get a chance I would love to make any further changes to this PR that you see fit.

isaiahdw:
Any thoughts on getting this merged? I've been using the `Gringill:feature/polly` fork, but it would be great to have this merged and include in an official release.

benwilson512:
This looks good, thanks!

lib/ex_aws/polly.ex
```diff
@@ -0,0 +1,115 @@
...
+      |> camelize_keys()
+      |> Map.merge(%{
+           "Text" => text,
+           "VoiceId" => voice_id,
+           "OutputFormat" => output_format})
+    request(:synthesize_speech, opts, request_opts)
+  end
+
+  defp request(action, data, opts \\ %{}) do
+    operation =
```
benwilson512: This variable is unused, why is it here?

lib/ex_aws/polly.ex
```diff
@@ -0,0 +1,115 @@
+defmodule ExAws.Polly do
+  @moduledoc """
+  Operations on AWS Polly
+
+  http://docs.aws.amazon.com/polly/latest/dg/API_Reference.html
+  """
+
+  import ExAws.Utils, only: [camelize_keys: 1, identity: 2]
+
+  @namespace "Polly_20160610"
```
benwilson512: This attribute is unused

lib/ex_aws/polly.ex
```diff
@@ -0,0 +1,115 @@
...
+  #############
+
+  @doc "Synthesize speech"
+  @spec synthesize_speech(text :: binary, voice_id :: binary, output_format :: binary) :: ExAws.Operation.JSON.t
+  @spec synthesize_speech(text :: binary, voice_id :: binary, output_format :: binary, opts :: synthesize_speech_opts) :: ExAws.Operation.JSON.t
+  def synthesize_speech(text, voice_id, output_format, opts \\ []) do
+    parser =
+      fn
+        {:error, _} = error, _ -> error
+        {:ok, %{body: body, headers: headers}}, config ->
```
benwilson512: This variable is unused.

lib/ex_aws/polly.ex
```diff
@@ -0,0 +1,115 @@
...
+      opts
+      |> camelize_keys()
+      |> Map.merge(%{
+           "Text" => text,
+           "VoiceId" => voice_id,
+           "OutputFormat" => output_format})
+    request(:synthesize_speech, opts, request_opts)
+  end
+
+  defp request(action, data, opts \\ %{}) do
```
benwilson512: Defaults are never used.

benwilson512:
Additionally, the test suite does not pass.

lib/ex_aws/polly.ex
Gringill: Indeed, switching to `@version "v1"` and actually using it.

lib/ex_aws/polly.ex
Gringill: Holdover from when I thought I was going to need a `"x-amz-target"` header, I was mistaken.

Gringill:
Thanks for looking at this Ben. I addressed all of your concerns and made some improvements on my feature/polly branch, should I make another pull request?

benwilson512:
yes please!

PR #352, opened 2017-02-06
Add us-east-2 region & endpoint to S3 options

jackvnimble:
I checked the list and this seems to be the only one that's missing. Coincidentally, it's the region I was trying to use for my project 😝 

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region


benwilson512:
Thanks!

PR #354, opened 2017-02-09
Fix presigned_url_opts type

bernardd:
The type specification for presigned_url_opts was making dialyzer complain about valid values. This change corrects it.


bernardd:
Sorry, disregard this. I was doing something really dumb on my end.

PR #356, opened 2017-02-10
fix s3 upload

mlankenau:
Hi there,
I had trouble uploading an image to S3 today (eu-central-1). I got a redirect. The bucket has to be at the beginning of the hostname. Changed that, please review.


benwilson512:
The bucket does not need to go at the front if the region is correctly configured. Please open an issue with your config so we can sort out what is wrong.

mlankenau:
Thx you are right. The region fixed it. Weird behavior (on AWS side).
Thx Alot!

Issue #241, opened 2016-10-04
Correct usage of EC2 filters 


thehosepipe:
Im trying to use filters such as describe_instances_filters. Cant find an example anywhere and I have tried what I think should be correct from the documents...

ExAws.EC2.describe_instances([max_results: 5, filter_1: [tag: "atagvalue"] ] ) 

Have I misread the documentation? How should this be done?
Thanks

benwilson512:
That looks correct based on the typespecs. I didn't actually write that bit though so maybe @sashaafm can weigh in?

@sashaafm is there a reason why we need the `_1` suffix on `filter_1` ?

thehosepipe:
It looks as though the filters might be fundamentally wrong, but I'm not sure. The filter_1 is so that the querystring for AWS ends up as &Filter.1.Name=...&Filter.1.Value=...

This is a working query for example
`%ExAws.Operation.RestQuery{http_method: :get, params: %{"Action" => "DescribeInstances", "Filter.1.Name" => "private-dns-name", "Filter.1.Value.1" => _your_DNS_, "Version" => "2015-10-01"}, path: "/", service: :ec2} `

But looking into the codebase I cant see how the filters in the array get turned into those Name and Value pairs

benwilson512:
yeah I agree. The EC2 API is massive and this just seems to be an oversight. @sashaafm any chance you've got some bandwidth to tackle this? If not I'll try to take a look at it later this week.

benwilson512:
By way of update: This project is at this point a product of not just my own efforts but the efforts of many collaborators. I do not have the time to be 100% involved in every aspect of this enormous API. Particularly in situations like these where I am not the original author of the API in question, I am going to defer to the original author or new contributors who want to make improvements.

Specifically in this case that's @sashaafm.

thehosepipe:
Not a problem. For now I have a workaround, furthermore I hope to find time to have a go at this myself. Some ideas but little time.

sashaafm:
Hello @thehosepipe and @benwilson512 yes I wrote both the EC2 and RDS API implementations. I'm going to try to have a look at this during the weekend (I'll try to allocate an hour or two). I was pretty sure everything was working as I tried to test every single function with the various input possibilites (and I also heavily used both services in the past), but like Ben said it might have been an oversight, the EC2 API is truly huge.

kyleaa:
[...]

PR #369, opened 2017-02-28
Fix EC2 filter implementations

kyleaa:
As discussed in https://github.com/CargoSense/ex_aws/issues/241, the filter implementations on the describe methods do not work correctly.

Since I need `describe_tags` I started with that one, and if this looks good I can implement changes in other methods. I borrowed inspiration from JS SDK 
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeTags-property

### DIFF
--- a/test/lib/ex_aws/ec2/integration_test.exs
+++ b/test/lib/ex_aws/ec2/integration_test.exs
@@ -5,4 +5,21 @@ defmodule ExAws.EC2.IntegrationTest do
     assert {:ok, %{body: body}} = ExAws.EC2.describe_instances |> ExAws.request
     assert body |> String.contains?("DescribeInstancesResponse")
   end
+
+  test "#describe_tags" do
+    assert {:ok, %{body: body}} = ExAws.EC2.describe_tags |> ExAws.request
+    assert body |> String.contains?("DescribeTagsResponse")
+  end
+  
+  test "#describe_tags with basic filters" do
+    params = [filters: [{:key, ["a-key", "b-key"]}, {:value, "a-value"}]]
+    assert {:ok, %{body: body}} = ExAws.EC2.describe_tags(params) |> ExAws.request
+    assert body |> String.contains?("DescribeTagsResponse")
+  end
+
+  test "#describe_tags with resource_type filter" do
+    params = [filters: [{:resource_type, [:customer_gateway]}]]
+    assert {:ok, %{body: body}} = ExAws.EC2.describe_tags(params) |> ExAws.request
+    assert body |> String.contains?("DescribeTagsResponse")
+  end
 end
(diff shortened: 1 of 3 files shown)


test/lib/ex_aws/ec2/integration_test.exs
```diff
@@ -5,4 +5,21 @@ defmodule ExAws.EC2.IntegrationTest do
     assert {:ok, %{body: body}} = ExAws.EC2.describe_instances |> ExAws.request
     assert body |> String.contains?("DescribeInstancesResponse")
   end
+
+  test "#describe_tags" do
```
kyleaa: I used these for my own sanity testing, but they could potentially leak data (if the request succeeds but the pattern match fails) through travis CI if they were given describe on a * resource - I'm not sure if those tests are isolated in their own AWS account.

kyleaa:
I moved the filter building into `normalize_opts` since it then fixes all of the filtered endpoints, which are numerous. How detailed do we want the typespecs to be? They currently are incredibly detailed (but don't work), so I was just curious before diving into fixing those.

I also presume all of the other `_1` typespec'd parameters need similar treatment - take a list, automatically apply the `.N` counts.

test/lib/ex_aws/ec2/integration_test.exs
benwilson512: Is it necessary to do actual requests here? Can't you just assert on the request body?

test/lib/ex_aws/ec2_test.exs
```diff
@@ -1609,6 +1609,41 @@ test "delete_key_pair with options" do
...
+          "Filter.1.Name" => "resource-type",
+          "Filter.1.Value.1" => "customer-gateway",
+          "Filter.2.Name" => "resource-type",
+          "Filter.2.Value.1" => "customer-gateway",
+          "Filter.2.Value.2" => "instance",
+        },
+        path: "/",
+        http_method: :get
+      }
+      assert expected == EC2.describe_tags filters: [{:resource_type, [:customer_gateway]},{:resource_type, [:customer_gateway, :instance]}]
```
benwilson512: normal keyword syntax should work fine here right? `[resource_type: [:customer_gateway], etc`

kyleaa: Yes it does, I'll update the tests to reflect.

kyleaa: re: integration tests, yeah they're probably unnecessary - I used them to sanity-check that it was working.

I'll go through and clean up the typespecs and other endpoints.

benwilson512:
Thanks!

PR #380, opened 2017-03-13
extend ExAws.Request.Url to handle stringified port

jbodah:
When the port is provided via an environment variabe it will come in as
a string, e.g.

```ex
config :ex_aws, :s3,
  scheme: {:system, "S3_SCHEME"},
  host: {:system, "S3_HOST"},
  port: {:system, "S3_PORT"}
```

`URI.to_string/1` expects port to be an integer or else it will raise.
This PR makes sure we perform that conversion if necessary to prevent
this exception


benwilson512:
Thanks!

PR #385, opened 2017-03-22
Add support for EC2 Container Service, aka ECS

larskrantz:
https://aws.amazon.com/ecs/

Support added for all API operations:
http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Operations.html

Thanks for a great package!
/Lars


lib/ex_aws/ecs.ex
```diff
@@ -0,0 +1,1091 @@
...
+    |> Map.merge(%{"service" => service})
+    request(:update_service, data)
+  end
+
+
+  ## Private parts
+
+  # Pascalize has room for improvements.. but it seems to work
+
+  defp pascalize_keys(opts) when is_map(opts) do
```
benwilson512: Just to be clear on what this does, it takes
```
[some_option: 1]
```
and makes it
```
[{"someOption", 1}]
```
right?

lib/ex_aws/ecs.ex
```diff
@@ -0,0 +1,1091 @@
...
+    {:container_port, 1..65535} |
+    {:load_balancer_name, binary} |
+    {:target_group_arn, binary}
+  ]
+
+  @type create_service_opts :: [
+    {:client_token, binary} |
+    {:cluster, binary} |
+    {:deployment_configuration, deployment_conf} |
+    {:load_balancers, list(load_balancer)} |
```
benwilson512: All of these uses of `list` would probably be better off as just `[load_balancer]`. It's mostly just a personal preference thing however.

larskrantz: Ok, no problem - I have no preference whatsoever, this is the first time I've used types.
I'll change it.

lib/ex_aws/ecs.ex
larskrantz: Yes, that's correct.

benwilson512: I think we can simplify it significantly:
```
defp pascalize_keys(opts) do
  Map.new(opts, fn {k, v} -> {pascalize_key(k), v} end)
end

defp pascalize_key(key) do
  {first, rest} =
    key
    |> to_string
    |> Macro.camelize
    |> String.split_at(1)

  String.downcase(first) <> rest
end
```

I'd actually like to put this function in the `ExAws.Utils` module because I think it has applicability beyond this service as well.

lib/ex_aws/ecs.ex
```diff
@@ -0,0 +1,1091 @@
...
+  def list_clusters(opts \\ []) do
+    data = opts
+    |> normalize_opts
+    request(:list_clusters, data)
+  end
+
+  @type container_instance_status :: [
+    :ACTIVE |
+    :DRAINING
+  ]
```
benwilson512: If we're going to typespec these I think it would be more idiomatic to have them lower case. This does require that we handle upcasing them.

lib/ex_aws/ecs.ex
```diff
@@ -0,0 +1,1091 @@
...
+  end
+
+
+  @type sort_order :: [
+    :ASC |
+    :DESC
+  ]
+  @type task_definition_status :: [
+    :ACTIVE |
+    :INACTIVE
```
benwilson512: Same with these, `sort_order`, and any other upper cased atoms.

lib/ex_aws/ecs.ex
larskrantz: Sweet, much better. I'll change it in this file and test it. Or would you like me to move it ExAws.Utils?

lib/ex_aws/ecs.ex
larskrantz: Agree, I'm on it.

lib/ex_aws/ecs.ex
larskrantz: Would that apply to `nonempty_list(binary)` transforming to `[binary]` or just `list`?

larskrantz:
Changes:
* list()-types changed to []
* used suggested `pascalize_keys`, but added recursion
* atoms in form of `:UPPER`, `:"json-file"` and `:memberOf` converted to `:upper`, `:json_file`, `:member_of` and converted to expected output

larskrantz:
Not sure if I'm supposed to ping @benwilson512 - so I'll just do it ;-)

lib/ex_aws/ecs.ex
```diff
@@ -1021,16 +1049,48 @@ defmodule ExAws.ECS do
...
+  defp replace_underscore_with_dash(value) when is_binary(value), do: String.replace(value, "_", "-")
+
+  defp pascalize_keys(opts) when is_map(opts) do
+    Map.new(opts, &pascalize_keys/1)
   end
-  defp pascalize_keys([]), do: []
-  defp pascalize_keys([head | tail]) do
-    [pascalize_keys(head)] ++ pascalize_keys(tail)
+  defp pascalize_keys(list) when is_list(list) do
+    list |> Enum.map(&pascalize_keys/1)
```
larskrantz: It's recursive and options can be deeply nested, with both maps and lists of maps etc. For an example of complex options, see [this test](https://github.com/larskrantz/ex_aws/blob/08fbc80655ac266d21e7090c2e8d86b23beb2622/test/lib/ex_aws/ecs_test.exs#L306-L347). We need to traverse the options and handle all of the keys and values deep down.

This is how I think: It starts with a map. The map has keys and values. Keys are pascalized using `pascalize_key/1` and values are recursively passed on to `pascalize_keys/1`. If it's a list, traverse and handle values. If a value is another list or map, recurse. If the value is a plain value, transform it if it's needed (atom to string) or just return value in the catch-all clause.
I can't really see how to use different function names for the list and values when recursing.

I'm sorry if it's confusing, but I'm totally open to suggestions as long as tests pass.

lib/ex_aws/ecs.ex
```diff
@@ -1021,16 +1049,48 @@ defmodule ExAws.ECS do
...
   end
   defp pascalize_keys(bool) when is_boolean(bool), do: bool
   defp pascalize_keys(atom) when is_atom(atom), do: Atom.to_string(atom)
```
larskrantz: Booleans are in fact atoms, and we want `true` and  `false` as values, not `"true"` and  `"false"`.  However, I have not handle the case when the value is `:true` or `:false`, just realizing that they are booleans, and values will be passed on. 
I've never seen booleans as those atoms, but suspect that the json-converters will handle them correctly so we don't need to match on them.

lib/ex_aws/ecs.ex
```diff
@@ -1021,16 +1049,48 @@ defmodule ExAws.ECS do
...
+    map
+    |> Keyword.new
+    |> transform_value(key, fn_transform)
+    |> Map.new
+  end
+  defp transform_value(opts, key, fn_transform) do
     opts
-    |> Enum.map(&pascalize_keys/1)
-    |> Enum.into(%{})
+    |> Keyword.get_and_update(key, fn(value) ->
```
larskrantz: Since the opts transforming are mainly keywords, transform of specific keys are done before normalize.

lib/ex_aws/ecs.ex
```diff
@@ -0,0 +1,1140 @@
...
+  container instance or instances.
+
+  Alternatively, you can use `run_task/2` to place tasks for you. For more information,
+  see [Scheduling Tasks](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/scheduling_tasks.html) in the Amazon EC2 Container Service Developer Guide.
+
+  _The list of container instances to start tasks on is limited to
+  10._
+  """
+  @spec start_task(task_definition :: binary, container_instances :: nonempty_list(binary)) :: ExAws.Operation.JSON.t
+  @spec start_task(task_definition :: binary, container_instances :: nonempty_list(binary), opts :: start_task_opts) :: ExAws.Operation.JSON.t
```
larskrantz: ok, thanks

mayppong:
I'm really looking forward to the ECS support. What was the reason for closing this PR?

larskrantz:
Lack of response. I guess somebody else has been doing it since readme states "ExAws.ECS (COMING SOON)".

benwilson512:
You may be interested in https://github.com/CargoSense/ex_aws/issues/497

mayppong:
@benwilson512, I saw the ticket and I was wondering the same thing with wanting to add RedShift support. I guess the breaking up of `ex_aws` is passed the proposal stage? If so, I'd be happy to start working on RedShift support as a separate project.

benwilson512:
Feedback has been uniformly positive, so I plan on going ahead with the change. I hope to move forward this week with everything, so please feel free to start RedShift as a separate project called `ex_aws_redshift` and I'll look to include it in the coming organization.

larskrantz:
The proposal is a wonderful idea, and I would be happy to do `ex_aws_ecs`. 
I understand your situation completely, `ex_aws` has grown to something very big.

PR #387, opened 2017-03-24
add export module S3.parsers.parse_list_multipart_uploads

kdxu:
function ` ExAws.S3.Parsers.parse_list_multipart_uploads ` cannot access from s3.ex because this function is not exported.
at s3.ex:271


benwilson512:
Thanks!

PR #401, opened 2017-05-06
Add ListIdentities call to SES

protochron:
This PR adds the ListIdentities call to the SES module. It is almost
identical to `list_configuration_sets` only that it has an additonal
optional parameter for specifying `IdentityType`

API Docs: http://docs.aws.amazon.com/ses/latest/APIReference/API_ListIdentities.html


benwilson512:
Thanks!

PR #405, opened 2017-05-11
fix test typo

bfauble:
fix test calling wrong function.

https://travis-ci.org/CargoSense/ex_aws/builds/231123985?utm_source=github_status&utm_medium=notification


benwilson512:
yeah just saw that, thanks for the fix!

Issue #404, opened 2017-05-09
ExAws.S3.delete_multiple_objects/3 does not produce compatible operation struct


DavidAntaramian:
I was looking into an issue where the request phase of `ExAws.S3.delete_multiple_objects/3` fails consistently. The struct `ExAws.Operation.S3DeleteAllObjects` struct is created, but unlike other operation structs it does not contain a `:service` key. The struct is designed as a struct that can produce `ExAws.Operation.S3` structs that handle the interaction with S3. The `ExAws.request/2` function, however, requires that the `:service` key be present on all operations ([see here](https://github.com/CargoSense/ex_aws/blob/fccbd34/lib/ex_aws.ex)).

To be honest, I'm not sure this function belongs in the ExAws library. It's a utility function, but it tries to act like a service function by producing an operation that can be used with the `request` and `stream` functions. Those functions have valid reasons to expect the `:service` key to be present, but `ExAws.Operation.S3DeleteAllObjects` doesn't conform to the standard operation structure anyways. I think it would be cleaner if the caller is responsible for handling calls to `ExAws.S3.delete_mulitple_objects/3` where the number of objects exceeds the 1000 object limit.

evadne:
I was also attempting to use this module, and found that it could possibly be invoked thusly,

`ExAws.Operation.S3DeleteAllObjects{} |> ExAws.Operation.perform`

But it seems that there is a typo preventing it from being used this way.

To be fair, I would rather the library does the chunking for me and handle retries transparently. Doing it in the caller is a lot of work!

benwilson512:
> To be honest, I'm not sure this function belongs in the ExAws library.

Ultimately I don't see this function as meaningfully different than any of the other streaming functions we have. There are a lot of AWS resources that are paginated and are a lot simpler to treat as a stream. This does essentially the same thing for a delete operation.

The particular issue has been fixed via #408 as best I can tell.

ijunaid8989:
Hey @benwilson512, first of all, thanks for all the work you put up for AWS lib, I was kind of going to create a new issue as I found this one, as I can see we have an `ExAws.S3.delete_multiple_objects/3`, is there any way to send multiple files at once? 

```
          S3.do_save("#{upload_path}#{compare_id}.gif", gif_content, [content_type: "image/gif", acl: :public_read])
          S3.do_save("#{upload_path}#{compare_id}.mp4", mp4_content, [acl: :public_read])
[...]

PR #408, opened 2017-05-19
fixes ExAws.Operation.S3DeleteAllObjects

evadne:
- corrects internal nomenclature: `files` → `objects`
- adds missing `:service` key
- addresses #404


benwilson512:
Thanks!

PR #412, opened 2017-05-24
[sqs] Add parsing for error entries in SendMessageBatch response

swaroopch:
- https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html#API_SendMessageBatch_ResponseElements
- https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_BatchResultErrorEntry.html

References:

- https://boto3.readthedocs.io/en/latest/reference/services/sqs.html#SQS.Client.send_message_batch
- https://github.com/boto/boto/blob/master/boto/sqs/batchresults.py
- https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-sqs/source/model/BatchResultErrorEntry.cpp
- https://github.com/aws/aws-sdk-go/blob/0bc99f4aa1cf50245de234b1ce0b33463a3ad5fe/service/sqs/api.go#L3689

### DIFF
--- a/lib/ex_aws/sqs/parsers.ex
+++ b/lib/ex_aws/sqs/parsers.ex
@@ -145,6 +145,7 @@ if Code.ensure_loaded?(SweetXml) do
     end
 
     def parse({:ok, %{body: xml}=resp}, :send_message_batch) do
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

Issue #411, opened 2017-05-24
EC2.describe_snapshots doesn't support more than one snapshot 


occurr:
While using the ES2 function `EC2.describe_snapshots` I noticed the type spec suggests using `[snapshot_id_1: "id", .... ]`. 

Based on the documentation, I would expect the syntax to be something like: 
```
EC2.describe_snapshots([snapshot_id_1: "snap-123", snapshot_id_2: "snap-456"])
|> ExAws.request!
```

Unforchenetly this doesn't work, as `SnapshotId_1:` is an invalid Aws request. _(It expects a list of snapshot `SnapshotId: []`)_. Because of this you can only request one snapshot at a time via `EC2.describe_snapshots` with the syntax `EC2.describe_snapshots([snapshot_id: "snap-123"])`.

I also looked into the test folder and found that no tests for this case. Is this not currently supported, or am I missing something?

benwilson512:
@sashaafm contributed the EC2 module, let's see if we can get him to chime in.

bearleethere:
I saw on the Amazon EC2's documentation if you want to specify more than one snapshot, the naming convention seems to be SnapshotId.N: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSnapshots.html

I was able to query for more than one snapshots with the EC2.describe_snapshots, but I needed to use a quoted atom. The underscore doesn't work. It needed to be a "." followed by a number. Not only that, the atom also had to be quoted. This didn't match the documentation where it looks like this: `{:snapshot_id_1, [binary]}`. 

This is the way where I was able to get more than one specific snapshot: 
```
EC2.describe_snapshots([
  {:"snapshot_id.1", "snap-123"}, 
  {:"snapshot_id.2", "snap-456"}
]) 
|> ExAws.request(region: "earth")
```

bearleethere:
Seems like it was a similar issue as this: https://github.com/CargoSense/ex_aws/pull/369
where the `_1` typespec'd parameters needs to switch to the `.N` counts

occurr:
It doesn't look like this issue is isolated to describe_snapshots. Using the `describe_tags` function with filters doesn't work as expected. Currently the function returns: 
```
iex> EC2.describe_tags filters: [{:key, ["a-key", "b-key"]}, {:resource_id, "id"}]
%ExAws.Operation.RestQuery{
 action: nil, body: "", http_method: :get,
 params: %{"Action" => "DescribeTags",
[...]

PR #413, opened 2017-05-26
fix ec2 list-based parameters

kyleaa:
The change from `:instances` to `:instance_id` in `describe_instances/1` and `describe_instance_status/1` are breaking, but they have never worked correctly in a released version so I think it's better to remove the special handling and standardize them with the AWS API parameter name.

This will support syntax like `instance_id: ["instance1", "instance2"]` but won't work with `instance_id: "instance1", instance_id: "instance2"`. I don't see a way to support the second one without adding a second iteration through the parameter list in `normalize_opts/1`, I'm not sure what the preference is for performance vs convenience (the typespec only shows the supported syntax).

This addresses #411

### DIFF
--- a/lib/ex_aws/ec2.ex
+++ b/lib/ex_aws/ec2.ex
@@ -21,7 +21,7 @@ defmodule ExAws.EC2 do
   ExAws.EC2.describe_instances(filters: [image_id: "ami-1ecae776"])
   ExAws.EC2.describe_instances(filters: ["network-interface.availability-zone": "us-east-1a"])
   ExAws.EC2.describe_instances(filters: ["tag:elasticbeanstalk:environment-name": "demo"])
-  ExAws.EC2.describe_instance_status(instances: ["i-e5974f4c"])
+  ExAws.EC2.describe_instance_status(instance_id: ["i-e5974f4c"])
   ```
   """
 
@@ -179,7 +179,7 @@ defmodule ExAws.EC2 do
   @type describe_instances_opts :: [
     {:dry_run, boolean}                        |
     {:filters, [filter]}                       |
-    {:instances, [binary]}                     |
+    {:instance_id, [binary]}                     |
     {:max_results, integer}                    |
     {:next_token, binary}
   ]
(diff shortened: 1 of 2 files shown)


lib/ex_aws/ec2.ex
```diff
@@ -21,7 +21,7 @@ defmodule ExAws.EC2 do
   ExAws.EC2.describe_instances(filters: [image_id: "ami-1ecae776"])
   ExAws.EC2.describe_instances(filters: ["network-interface.availability-zone": "us-east-1a"])
   ExAws.EC2.describe_instances(filters: ["tag:elasticbeanstalk:environment-name": "demo"])
-  ExAws.EC2.describe_instance_status(instances: ["i-e5974f4c"])
+  ExAws.EC2.describe_instance_status(instance_id: ["i-e5974f4c"])
```
benwilson512: The fact that it's a list here would seem to imply that you can pass multiple to it. If this is correct we need to use `instance_ids`.

kyleaa: Certainly doable, this first stab was naming them to match the API parameters.

Without adding multiple enumerations through the options these could all be handled in the `normalize_opts` reduce with either of these...

  ```@pluralized_parameters [
    "instance_id",
    "block_device_mapping",
    ...
  ]

  @pluralized_parameters
  |> Enum.map(fn param -> defp reduce_list_params({unquote(String.to_existing_atom("#{param}s")), val}, acc) when is_list(val), do: Map.merge(acc, list_builder(val, unquote(String.to_atom(param)), 1, %{})) end)```

or defining each in full
```
defp reduce_list_params({:instance_ids, keys}, acc), do: Map.merge(acc, list_builder(val, :instance_id, 1, %{}))
  defp reduce_list_params({:block_device_mappings, keys}, acc), do: Map.merge(acc, list_builder(val, :block_device_mapping, 1, %{}))
  ```

Do either of those sound reasonable or should they all be handled in the function? This is how I was handling that before:

```
instance_ids = Keyword.get(opts, :instances, [])
    query_params = opts
    |> Keyword.delete(:instances)
    |> normalize_opts
    |> Map.merge(%{
      "Action"  => "DescribeInstanceStatus",
      "Version" => @version
      })
    |> Map.merge(list_builder(instance_ids, "InstanceId", 1, %{}))
```

bearleethere: Just from a quick glance, what you currently have looks good to me. 

@benwilson512, your thoughts?

benwilson512: Hm yeah I think this is fine as is.

benwilson512:
Thanks!

PR #414, opened 2017-05-26
Updated Dynamo create_table example.

AlexMouton:
attribute_type seems to have been replaced by key_type, and the 'email' attribute is required in your attribute definitions for the base table.

HTH
Thanks
A


benwilson512:
Thanks!

PR #416, opened 2017-05-29
use Base.encode16 :case option instead of piping to String.downcase

scmorrison:
Not sure if there is a preference for piping to String.downcase, but Base.encode16 has a :case option for this:

```elixir
Base.encode16(case: :lower)
```


benwilson512:
Thanks!

Issue #420, opened 2017-06-08
ExpiredToken error on credentials from instance metdata


adamkittelson:
Hey Ben, a while back (6 months at least). I mentioned in irc a weird thing that happened on one of our QA servers. We use instance metadata for credentials and on one of the boxes all the SQS requests started returning `%{code: "ExpiredToken", detail: "", message: "The security token included in the request is expired", request_id: "xxx", type: "Sender"}`. I never got around to digging in on handling that because it had never happened before and seemed like kind of a fluke.

Last night it happened again, but this time on one of our production nodes which ended up taking the node down due to exceeding supervisor restart thresholds. So, while this does seem to be super rare (experienced it twice in the last year), it definitely can happen and when it does it's really bad because there's currently no mechanism to prompt ex_aws to refresh the credentials.

I'd like to get a discussion going on how I can ensure that this error is handled when it occurs, and to what degree I should be handling this in my app vs in ex_aws.

osayame:
Hey Ben, 

We have also been experiencing this issue in a production node for the last few weeks. A system restart resolves the issue.

After looking through the `AuthCache` (https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/config/auth_cache.ex#L14)  module, it appears that the expiration logic is not implemented for when using the `:aws_instance_auth` option. Is this assumption correct? If so, it would explain the situation where a token still exists, but is expired.

Thanks,
Osa

benwilson512:
That it is not correct. The logic you're linking to merely says that if no auth exists at all yet, do a manual refresh regardless of any existing timers. Any auth checks will ultimately hit this logic: https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/config/auth_cache.ex#L58-L60 which always schedules a refresh.

Looking at this https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/config/auth_cache.ex#L67 though I'm only refreshing 2 min prior to the expiration. If there's much clock drift this is potentially not enough of a buffer. Is anyone clear on how early I can ask for a refresh?

osayame:
Ben, thanks for clarifying.

The docs recommend checking 5 minutes prior:
https://aws.amazon.com/premiumsupport/knowledge-center/security-token-expired/

adamkittelson:
[...]

PR #421, opened 2017-06-12
Feature/2017.06.12 increase refresh time before expiration

osayame:
### Features
- Increases refresh time before expiration from `2` minutes to `5` minutes per https://github.com/CargoSense/ex_aws/issues/420


benwilson512:
Thanks!

PR #423, opened 2017-06-14
Added some more CloudFormation's API Operations

bearleethere:
API Operations added:
- CreateStack
- DeleteStack
- GetTemplate

Added the parsers for the API Operations added, and also fixed a couple of bugs regarding the params ("RoleArn" should be "RoleARN", just camelizing doesn't work) and cleaned up some other things in the utils.ex.


bearleethere:
@benwilson512 This would be useful for supporting CloudFormation in this library.

PR #424, opened 2017-06-15
Modify ExAws.Dynamo.Encodable.encode for: List to handle encoding an …

rinosamakanata:
…all number List to Dynamo's NS(Number Set) type.

### DIFF
--- a/lib/ex_aws/dynamo/encodable.ex
+++ b/lib/ex_aws/dynamo/encodable.ex
@@ -102,12 +102,21 @@ defimpl ExAws.Dynamo.Encodable, for: List do
 
   @doc """
   Dynamodb offers typed sets and L, a generic list of typed attributes.
+  If all elements in a list are number, the list is encoded to NS type.
   """
   def encode(list, _) do
-    typed_values = for value <- list do
-      Encodable.encode(value, [])
+    if Enum.all?(list, fn(x) -> is_integer(x) || is_float(x) end) do
+      %{"NS" => Enum.map(list, fn(x) ->
+        cond do
+          is_integer(x) -> Integer.to_string(x)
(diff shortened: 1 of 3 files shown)


rinosamakanata:
@benwilson512, Thank you for your advice! I'll implement a protocol for MapSet that has all numbers that are not duplicated.

benwilson512:
Well that's the nice thing about a MapSet, by definition it won't have any duplicates. You obviously will have to ensure they're all numbers.

rinosamakanata:
@benwilson512 Thanks! I reverted my previous commits and newly implemented a protocol for MapSet. How about this?

rinosamakanata:
In addition, I added some code to convert string-only MapSet for Dynamo's SS (StringSet) type.

lib/ex_aws/dynamo/encodable.ex
```diff
@@ -111,3 +111,29 @@ defimpl ExAws.Dynamo.Encodable, for: List do
...
+  def encode(mapset, _) do
+    cond do
+      MapSet.size(mapset) == 0 ->
+        raise "Cannot determine a proper data type for an empty MapSet"
+      Enum.all?(mapset, &is_number/1) ->
+        %{"NS" => number_set_to_list(mapset)}
+      Enum.all?(mapset, &is_binary/1) ->
+        %{"SS" => MapSet.to_list(mapset)}
+      true ->
+        raise "All elements in a MapSet must be only numbers or only strings"
```
benwilson512: I'm not wildly enthusiastic about raising exceptions here, but i'm not sure that the alternatives are much better.

rinosamakanata:
@benwilson512 Yes, I'm not sure about raising exceptions, but I didn't find better solution, too.. Thank you for merging!

PR #425, opened 2017-06-15
Fix incorrectly public functions in Cloudformation module

occurr:
Using `Kernel.apply` in the Cloudformation module look to have cleaned up a lot of boilerplate code for the transforms, but unfortunately also exposed many helper functions to the public interface that **should not be used directly**. This pull request fixes that issue by replacing apply with function patern_matching.  

Also added tests for build_indexed_params.


benwilson512:
Good choice.

bearleethere:
Jesus you're fast.

occurr:
Wow you're fast. Where the heck are you getting your coffee from?

PR #430, opened 2017-06-19
Update deps version {hackney, poison}

kdxu:
Hi. I'm using dialyzer it depends on `ex_aws`.
I found error when getting error building a plt of certifi, 
And It caused by an old version of hackney depends on an  old version of certifi.

## Summery
- Updated hackney latest version.
- Updated poison latest version.

## Backgrounds
- current version of hackney has dialyzer warnings. (https://github.com/benoitc/hackney/issues/409)
- now latest version of hackney, solved dialyzer bugs.


mix.exs
```diff
@@ -36,10 +36,10 @@ defmodule ExAws.Mixfile do
       {:xml_builder, "~> 0.0.6", optional: true},
       {:sweet_xml, "~> 0.6", optional: true},
       {:ex_doc, "~> 0.14", only: :dev},
-      {:hackney, "1.6.3 or 1.6.5 or 1.7.1 or 1.8.6", optional: true},
-      {:poison, ">= 1.2.0", optional: true},
+      {:hackney, "1.8.6", optional: true},
```
benwilson512: I don't understand the point of this. I already permit 1.8.6. I'm not going to force people to upgrade just to fix dialyzer warnings.

mix.exs
```diff
@@ -36,10 +36,10 @@ defmodule ExAws.Mixfile do
...
+      {:poison, "~> 3.1.0", optional: true},
```
benwilson512: I'm not particularly interested in forcing people to upgrade Poison either. It works on 1.2 still so I may as well keep that. Poison doesn't maintain a changelog so it's a bit hard to know what actually changes between versions, so I don't want to for people to upgrade this either.

benwilson512:
Hey! I appreciate the PR, but I'm not really sure I see the need to force people to ugprade. I already support the latest versions of hackney and poison, all this PR does is remove support for earlier versions.

mix.exs
kdxu: Oh sorry it's my fault. I will revert this soon.

benwilson512: Well I mean by the time we revert this and the poison change there aren't really any changes left. We can bump the version of dialyze but I think it'd just be cleaner for me to do that in master.

mix.exs
kdxu: > Poison doesn't maintain a changelog so it's a bit hard to know what actually changes between versions, so I don't want to for people to upgrade this either.

I didn't know such background of Poison :( ...
Thanks for the information, now I think it doesn't need for update :)

benwilson512:
I really do appreciate your desire to help, but I think it will be easier for me to just update dialyze, and leave the other dependencies as is.

kdxu:
Thank you reviewing for this PR.
Now I know these backgrounds of Poison and apologize for my fault.
So, you may close this pull request 🙇

kdxu:
Thanks!!

PR #435, opened 2017-06-23
Cleanup of cloudformation, and related Utils functions. 

occurr:
General cleanup of the cloudformation module, and changed to the relevant utils. Including 

- Additional documentation for cloudformation
- Better overall naming of cloudformation's new functions
- Renamed cloudformation's private `transform` functions to `format_param` 
- Refactored Utils.build_indexed_params to flatten_params. Now recursive, making for more concise usage. 
- Refactored `format_param` functions to use new Utils.flatten_params
- Added `Utils.maybe_format` macro. A very simple macro for calling private `format_param` functions if the param isn't null.
- Added `Utils.maybe_camelize` and `Utils.maybe_stringify` to reduce atom and string checking clutter
- Added `__using__` macro to `Utils` for basic `import + require`

### DIFF
--- a/lib/ex_aws/utils.ex
+++ b/lib/ex_aws/utils.ex
@@ -171,8 +171,8 @@ defmodule ExAws.Utils do
         maybe_camelize: 2, maybe_camelize: 1 
       ]
 
-      def flatten_params(params, kwargs \\ [prefix: ""]), 
-        do: ExAws.Utils.flatten_params(params, unquote(inject))
+      def format(params, kwargs \\ [prefix: ""]), 
+        do: ExAws.Utils.format(params, unquote(inject))
       def camelize_keys(opts, kwargs \\ [deep: false]), 
         do: ExAws.Utils.camelize_keys(opts, unquote(inject))
       def camelize_key(opts, kwargs \\ []), 
--- a/lib/ex_aws/utils.ex
+++ b/lib/ex_aws/utils.ex
@@ -128,12 +128,12 @@ defmodule ExAws.Utils do
     Enum.flat_map(key_templates, fn {key_template, values} -> build_indexed_params(key_template, values) end)
   end
 
-  # a simple macro for calling private format functions if param isn't nul
-  defmacro maybe_format(map, format_name) do
+  # A simple macro for calling private format functions if param isn't nul
+  defmacro maybe_format(params, key) do
     quote do
-      case unquote(argument)[unquote(format_name)] do
+      case unquote(params)[unquote(key)] do
         nil -> []
-        value -> format_request(unquote(format_name), value)
+        value -> format_param(unquote(key), value)
       end
     end
   end
--- a/lib/ex_aws/utils.ex
+++ b/lib/ex_aws/utils.ex
@@ -155,7 +155,14 @@ defmodule ExAws.Utils do
   def maybe_stringify(elem) when is_atom(elem),      do: Atom.to_string(elem)
   def maybe_stringify(elem) when is_bitstring(elem), do: elem
 
-  defmacro __using__(non_standard_keys: non_standard_keys) do
+  defmacro __using__(kwargs) do
+    inject = 
+      quote do 
+        [ type: unquote(kwargs[:format_type] || :xml), 
+          spec: unquote(kwargs[:non_standard_keys] || %{}) ] 
+        ++ [kwargs] 
+      end
+
     quote do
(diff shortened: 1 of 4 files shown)


occurr:
While the addition of `maybe_format` macro may seem a little heavyweight for just the cloudformation module. The usage of the `maybe_format` and `format_param` structure has simplified the cloudformation module significantly. 

In addition, @bearleethere is currently working on the EC2 cleanup and is already using these new/updated utils to reduce the modules overall complexity. They will likely be applicable anywhere indexed params exist and need to be convered to the flattened (`Elem.3.Member`) format.

bearleethere:
So far, after using this in my branch, I can say that all of the custom list builders functions in ec2 can go away. This has worked for every ec2 list params, including TagSpecification, InstanceId, Filter, Tag, etc. It has simplified the module quite a bit.

lib/ex_aws/utils.ex
```diff
@@ -109,23 +116,58 @@ defmodule ExAws.Utils do
...
 
-  def build_indexed_params(key_template, value), do: build_indexed_params(key_template, [value])
+  def maybe_camelize(elem)  when is_atom(elem),     do: camelize_key(elem)
+  def maybe_camelize(elem)  when is_bitstring(elem),do: elem
+  def maybe_stringify(elem) when is_atom(elem),     do: Atom.to_string(elem)
+  def maybe_stringify(elem) when is_bitstring(elem),do: elem
 
-  def build_indexed_params(key_templates) when is_list(key_templates) do
-    Enum.flat_map(key_templates, fn {key_template, values} -> build_indexed_params(key_template, values) end)
+  defmacro maybe_format(argument, format_name) do
```
benwilson512: Can't this just be a function?
```
def maybe_format(argument, format_name) do
  case argument[format_name] do
    nil -> []
    value -> format_param(format_name, value)
  end
end
```

lib/ex_aws/utils.ex
```diff
@@ -109,23 +116,58 @@ defmodule ExAws.Utils do
...
+        add_prefix(key, kv_pairs) |> flatten_params
+        
+      {key, value} -> 
+        [{key, value}]
+    end)
+    |> Enum.to_list
+  end
+  # When only one key_template and value is passed  
+  def flatten_params(key, value), 
+  do: [{maybe_camelize(key), value}]
```
benwilson512: Sorry, but I'm really not a fan of this style. Please either make it on a single line or, if it's going to be multiline, use `do end`

lib/ex_aws/utils.ex
occurr: We can't refactor this macro to a function unless we make all `format_param(:param, opts)` public 

That being said, If we implement a module by module `normalize_opts` that catch non-standard keys, almost every function will boil down to calling `Enum.flat_map(opts, &maybe_format(opts, &1))` and this macro will not be necessary. This would make things both more concise, and make functions very consistent and readable.

lib/ex_aws/utils.ex
occurr: Absolutely, will do

occurr:
I performed a major refactor. There are no macros used other than `use`. 

`use Utils` lets us specify the `format_type` as `:xml` or `:json`. and pass in any number of non-standard keys as a map. 

After this refactor, we removed most of the redundancy in creating xml formatted requests. At this point, it is mostly documentation :)

bearleethere:
@benwilson512 Hey Ben! I just finished a major rework of the EC2 module. I am planning on using these changes in refactor of the EC2 module, so once this is pushed to master, I can update my branch and send a pull request. 

We should also talk about a 1.2 release considering the mounting changes to ExAws. So far, we added a lot of functionalities, fixed a lot of broken operations, and improved the docs for EC2 and for Cloudformation. These changes should help anyone currently using these modules.

benwilson512:
Getting an error on the CI https://travis-ci.org/CargoSense/ex_aws/builds/248821574?utm_source=email&utm_medium=notification

I really gotta figure out how to get the CI to run on PRs.

benwilson512:
I took a look but I'm not sure how to fix this. Let me know when you've got a fix in, gonna release v1.1.3 after.

occurr:
Ok, I found the issue. `:stack_filter: [...]` didn't match the AWS spec so was changed in the new cloudformation module. I reintroduced deprecated support for this param in #438

PR #436, opened 2017-06-29
add sts methods: assume role, decode_authorization_message, get_session_token

kyleaa:
Some services don't allow cross-account permissions without using AssumeRole. 

While I was in there I added two other missing methods - DecodeAuthorizationMessage and GetSessionToken. 

This still leaves two missing methods:
AssumeRoleWithSAML
AssumeRoleWithWebIdentity

I can't see any use for WebIdentity in elixir, as it is designed for mobile app authentication. The SAML method would be very difficult to test, and they don't even provide sample request/response data on that doc. I'd rather err on the side of caution and not include them, but certainly could if that's preferred.

### DIFF
--- a/lib/ex_aws/sts/parsers.ex
+++ b/lib/ex_aws/sts/parsers.ex
@@ -2,6 +2,30 @@ if Code.ensure_loaded?(SweetXml) do
   defmodule ExAws.STS.Parsers do
     import SweetXml, only: [sigil_x: 2]
 
+    def parse({:ok, %{body: xml}=resp}, :assume_role) do
+      parsed_body = xml
+      |> SweetXml.xpath(~x"//AssumeRoleResponse",
+                        access_key_id: ~x"./AssumeRoleResult/Credentials/AccessKeyId/text()"s,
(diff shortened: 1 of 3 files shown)


lib/ex_aws/sts/parsers.ex
```diff
@@ -2,6 +2,30 @@ if Code.ensure_loaded?(SweetXml) do
...
+
+      {:ok, Map.put(resp, :body, parsed_body)}
+    end
+
+    def parse({:ok, %{body: xml}=resp}, :decode_authorization_message) do
+      parsed_body = xml
+      |> SweetXml.xpath(~x"//DecodeAuthorizationMessageResponse",
+                        decoded_message: ~x"./DecodeAuthorizationMessageResult/DecodedMessage/text()"s,
+                        request_id: request_id_xpath())
+      |> Map.update!(:decoded_message, &Poison.decode!/1)
```
benwilson512: So the thing here is that we can't hardcode the JSON codec. Not everyone uses Poison. The configured `json_codec` can be found in the config. You'll need to add that to the arguments of the parse function, which is pretty easy because it's available inside the operation.

benwilson512:
Actually I'll just fix it, trying to release.

PR #438, opened 2017-06-30
Added (deprecated) support for :status_filter param

occurr:
One Cloudformation integration test was failing due to using the `:status_filter` param instead of the updated  `:stack_status_filters`. This was changed because `:status_filter` doesn't match the AWS spec `StackStatusFilter`.  And doesn't match `ExAws.Cloudformation`'s own naming convention of pluralizing params that require lists. (for example `stack_status_filters: [...]`)

In addition to fixing the integration test, support for the old `:status_filter` param has been readded with a deprecated warning. This function should be removed in 1.2.


benwilson512:
Thanks!

PR #440, opened 2017-07-01
Make `use ExAws.Utils` expanded functions private

occurr:
`use ExAws.Utils` expands to four partially applied Util functions with `:non_standard_keys` and `:format_type` automatically applied to every relevant Util function. This is very useful for removing some unnecessary, and hard to read code. 

Unforchenetly our implementation of `Utils.__using__` was adding public functions. We were foreseeing a possible use-case where having `Cloudformation.format` be public would be useful. But we neglected to think about the documentation. 

While there is no downside from an encapsulation standpoint, it clutters the documentation with functions that the end user doesn't need to know about. This pull request changes the functions to be private.

There are other ways of addressing the above-mentioned usecase that don't have this downside. So this is probably a good ide.a

### DIFF
--- a/lib/ex_aws/utils.ex
+++ b/lib/ex_aws/utils.ex
@@ -11,7 +11,7 @@ defmodule ExAws.Utils do
     deep = kwargs[:deep] || false
     spec = kwargs[:spec] || %{}
     do_cmlz_keys(opts, deep: deep, spec: spec)
-  end 
+  end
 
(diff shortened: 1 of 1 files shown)


occurr:
Also, there was some whitespace that got added in a commit somewhere. So I got rid of that as well.

benwilson512:
Looks good, thanks!

PR #441, opened 2017-07-02
Add (beta) tags to readme for cloudformation and EC2

occurr:
Another very small pull request. But I think it's a good idea to be clear that EC2 and Cloudformation are not yet fully fleshed out and with parsers.

Readme now has

- `ExAws.Cloudformation (BETA)`
- `ExAws.EC2 (BETA)`
- `ExAws.ECS (COMING SOON)`


benwilson512:
Thanks!

PR #443, opened 2017-07-03
Gets rid of unused variables warning.

sabondano:
Gets rid of the following warning:
```
warning: variable "val" is unused
  lib/ex_aws/sqs/parsers.ex:273
```


benwilson512:
Thanks!

PR #444, opened 2017-07-04
Add SNS ListEndpointsByPlatformApplication

kanmo:
https://docs.aws.amazon.com/sns/latest/api/API_ListEndpointsByPlatformApplication.html

Support for Lists the endpoints and endpoint attributes for devices in a supported push notification service.


benwilson512:
Thanks!

PR #448, opened 2017-07-06
Fixed a lot of the EC2 API Operations

bearleethere:
-Fixed parameters allowed to pass in for a few of the API operations.
-Updated to use version 2016.11.15 for AWS EC2
-Refactored the code heavily
-Added a bunch more tests
-Added docs for the EC2 Module with examples for each of the different API endpoints
-Made sure the API operations work or the results of query matches AWS' doc.

### DIFF
--- a/lib/ex_aws/cloudformation.ex
+++ b/lib/ex_aws/cloudformation.ex
@@ -24,8 +24,8 @@ defmodule ExAws.Cloudformation do
   http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Operations.html
   """
 
-  use ExAws.Utils, 
-  format_type: :xml, 
+  use ExAws.Utils,
+  format_type: :xml,
   non_standard_keys: %{
     stack_name: "StackName",
     role_arn: "RoleARN",
@@ -40,23 +40,13 @@ defmodule ExAws.Cloudformation do
 
   @version "2010-05-15"
 
-  @type accepted_capability_vals ::
-    :capability_iam |
-    :capability_named_iam
-
-
-  @type on_failure_vals ::
-    :do_nothing |
-    :rollback |
-    :delete
-
   @type parameter :: [
     parameter_key: binary,
     parameter_value: binary,
     use_previous_value: boolean
   ]
 
-  @type tag :: 
+  @type tag ::
     {key :: atom | binary, value :: binary}
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/ex_aws/ec2.ex
```diff
@@ -2,1924 +2,2795 @@ defmodule ExAws.EC2 do
...
-  import ExAws.Utils
-
-  @version "2015-10-01"
-
-  ########################
-  ### Instance Actions ###
-  ########################
+  @version "2016-11-15"
 
-  @type instance_types :: [
```
bearleethere: We shouldn't keep track of the instance_types or any of the type specs that AWS only accepts. This becomes tedious and most likely will not keep up with AWS' changes.

lib/ex_aws/cloudformation.ex
```diff
@@ -40,23 +40,13 @@ defmodule ExAws.Cloudformation do
 
   @version "2010-05-15"
 
-  @type accepted_capability_vals ::
```
bearleethere: I don't see how the type spec for this helps for the following reasons:
1) When AWS makes changes, we would need to update the typespec to reflect those changes. 
2) Easier for the user to pass in the correct value that AWS dictates in their docs instead of counting on some kind of helper function to convert the atom of the accepted values. One thing I noticed about AWS is that the values allowed can be very inconsistent. Safer for people using this library to just pass in the correct value.

bearleethere:
@benwilson512 @occurr 
All of these should work now. Added a bunch of tests and validated that the built request matches with AWS documentation. Also added a bunch of tests and documentations on how to use each of these. 

I also got rid of the type specs used in Cloudformation and EC2. I don't see how the type spec help much for the following reasons:
1) When AWS makes changes, we would need to update the typespec to reflect those changes. This may go unnoticed for a while. 
2) Easier for the user to pass in the correct value that AWS dictates in their docs instead of counting on some kind of helper function to convert the atom of the accepted values. One thing I noticed about AWS is that the values allowed can be very inconsistent and it doesn't make sense to try to convert the atom to the accepted values AWS accepts. 

Examples: For some values, like for the "capabilities" params, the values all need to be upcased. However, for some of the values passed in for filters, some have both a combination of "." and "-". Some are "ID" instead of "Id". Can be a bit annoying to deal with.

benwilson512:
@occurr Do you want to weigh in here? I know you've been looking at both CloudFormation and EC2.

occurr:
@benwilson512 Everything looks good. EC2 is now something like 80% documentation, so that is a major improvement 👍 

The use of the new Utils all looks good. There may be a few instances I can cut out some minor redundancy in their usage, but I wasn't able to find anything significant. I will submit my own pull request if I notice anything that can be cleaned up while writing the parsers. 

Everything looks solid, so I'd say it's ready to pull.

bearleethere:
Woo hoo!!! Thanks!!!

PR #449, opened 2017-07-06
Detect error in instance_meta HTTP GET request

visciang:
For example, if the EC2 instance has no IAM Role attached the response is "404 not found".


benwilson512:
Thanks!

PR #452, opened 2017-07-09
fix @spec for map() type

visciang:
Some fixes to **spec for map()** type.

> the syntactic representation of map() is %{optional(any) => any}, not %{}. The notation %{} specifies the singleton type for the empty map.

PS: I see there is no dialyzer step in your travis build. Currently dialyzer reports about 19 warning. I think could be useful to fix those and add a dialyzer step in travis to find immediately grey spots in new code.


benwilson512:
Thanks! Yeah these specs were added back when `%{}` meant any map.

PR #453, opened 2017-07-12
Make S3.list_objects work with stream! and prefix

ijcd:
I have a fixture bucket that I created in the account I was using. The best approach would be local fixtures, but I can send you the format of the bucket (or you can duplicate from tests). I ran into corner cases that required each of these.

```
# for testing delimiter without pagination
bucket/folders/short/0001/keep
bucket/folders/short/0002/keep
bucket/folders/short/0003/keep

# for testing delimiter with pagination
bucket/folders/long/0001/keep
...
bucket/folders/long/1200/keep

# for testing delimiter without pagination
bucket/objects/short/0001
bucket/objects/short/0002
bucket/objects/short/0003

# for testing delimiter with pagination
bucket/objects/long/0001
...
bucket/objects/long/1200
```

### DIFF
--- a/lib/ex_aws/s3/lazy.ex
+++ b/lib/ex_aws/s3/lazy.ex
@@ -9,22 +9,20 @@ defmodule ExAws.S3.Lazy do
     end
 
     Stream.resource(fn -> {request_fun, []} end, fn
-      :quit -> {:halt, nil}
+      :quit ->
+        {:halt, nil}
 
-      {fun, args} -> case fun.(args) do
+      {fun, args} ->
+        results = fun.(args)
+        contents = Map.get(results, :contents, [])
+        common_prefixes = Map.get(results, :common_prefixes, [])
 
-	      results = %{contents: contents = [_|_], is_truncated: "true"} ->
-          {contents, {fun, [marker: next_marker(results)]}}
-
-	      %{contents: contents = [_|_]} ->
-          {contents, :quit}
-
-        results = %{common_prefixes: common_prefixes = [_|_], is_truncated: "true"} ->
-          {common_prefixes, {fun, [marker: next_marker(results)]}}
-
-        %{common_prefixes: common_prefixes} ->
-          {common_prefixes, :quit}
-      end
+        case results do
+          %{is_truncated: "true"} ->
+            {contents ++ common_prefixes, {fun, [marker: next_marker(results)]}} 
+          _ ->
+            {contents ++ common_prefixes, :quit}
+        end
     end, &(&1))
   end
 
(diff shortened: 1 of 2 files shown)


ijcd:
Ok, this should cover the corner cases. The code reads better too.

New fixtures:

````
buckets/mixed/neither_paginate/{0001..1200}
buckets/mixed/neither_paginate/d{0001..1200}/keep

buckets/mixed/both_paginate/{0001..1200}
buckets/mixed/both_paginate/d{0001..1200}/keep

buckets/mixed/folders_paginate/{0001..0003}
buckets/mixed/folders_paginate/d{0001..1200}/keep

buckets/mixed/objects_paginate/{0001..1200}
buckets/mixed/objects_paginate/d{0001..0003}/keep
```

benwilson512:
Hey! These look like good changes, can you supply more info on the fixture bucket you had setup?

ijcd:
I had it named so it was the first bucket, for a reason I can no longer recall (aaaa-aws). These are all zero-length files:

   https://gist.github.com/ijcd/fec6d90de340e2fd8bc4a9f921366469

I have a tar.gz of it here, which I will leave for awhile:

  https://drive.google.com/open?id=0B5mnv4BFZdhmb0FObFFlN0VhQVk

This is a lot of fixture data. Today I would probably do something with FakeAWS and StreamData.

benwilson512:
Sorry but in looking at this more closely I don't think this is going to work. Right now the stream implementation only emits content results, if we change it to also emitting common prefixes it's going to be a massive breaking change. Everyone's code assumes each item is an object.

If you also need custom prefixes I think the best bet is a custom lazy iterator.

benwilson512:
Hey @ijcd,

What blows up right now?

Here's what I mean about how this is a breaking change. Right now the following code works:

```
iex(8)> (
  ExAws.S3.list_objects(bucket, delimiter: "/", prefix: "foo/")
  |> ExAws.stream!
  |> Stream.map(&(&1.key))
  |> Enum.to_list
)
```

It works because every item in the stream has a consistent shape, so if we want to get all the keys we can just ` |> Stream.map(&(&1.key))`.

After this PR however this code no longer works:

```
** (KeyError) key :key not found in: %{prefix: "..."}
    (stdlib) erl_eval.erl:668: :erl_eval.do_apply/6
    (elixir) lib/stream.ex:516: anonymous fn/4 in Stream.map/2
    (elixir) lib/enum.ex:3161: Enumerable.List.reduce/3
    (elixir) lib/stream.ex:1312: Stream.do_list_resource/6
    (elixir) lib/stream.ex:1433: Enumerable.Stream.do_each/4
    (elixir) lib/enum.ex:1823: Enum.reverse/1
    (elixir) lib/enum.ex:2581: Enum.to_list/1
```

ijcd:
Is been a few months. Let me look again. That wasn't my recollection.

--Ian

PR #455, opened 2017-07-14
Add us-gov-west-1 to config defaults for S3

mikeastock:


benwilson512:
Is this a publicly available region?

mikeastock:
Yes it is. https://aws.amazon.com/govcloud-us/

benwilson512:
Thanks!

PR #457, opened 2017-07-24
Add support for Polly Service

Gringill:
https://aws.amazon.com/polly/

All operations supported: http://docs.aws.amazon.com/polly/latest/dg/API_Operations.html


benwilson512:
Hey! I would love to include the polly service in the ExAws organization. Per https://elixirforum.com/t/proposal-exaws-2-0/9269 ExAws has split each service into its own project to ease the maintenance process.

To continue with the PR process please see: https://github.com/ex-aws/ex_aws/blob/master/CONTRIBUTING.md

PR #458, opened 2017-07-24
fix phoneNumber key name for opt_in_phone_number request

bfauble:
Mistakenly had up-cased key name for the request, should be lowercase per the docs.

http://docs.aws.amazon.com/sns/latest/api/API_OptInPhoneNumber.html


benwilson512:
Thanks!

PR #459, opened 2017-07-26
Use String.trim and String.trim_trailing to fix warnings

jayjun:


benwilson512:
Already fixed in master, sorry I didn't see this earlier, thanks!

PR #461, opened 2017-07-31
Support hackney versions >= 1.6 but < 2.0

aptinio:
This simplifies the required hackney version, at the same time,
adds support for the newly released 1.9.0 version.


benwilson512:
Hackney releases a string of patch releases that broke ExAws and other libraries. I will consider permitting an additional ~> 1.9 clause but only the enumerated 1.6 versions are known to be safe.

benwilson512:
Hackney released a string of patch releases that broke ExAws and other libraries. I will consider permitting an additional ~> 1.9 clause but only the enumerated 1.6 versions are known to be safe.

aptinio:
Oh! Got it. Let me create another PR. Thanks!

PR #463, opened 2017-08-01
fixed CustomUserData if it’s nil

iwarshak:
I have run into a strange and confusing situation with SNS and ex_aws. Specifically when registering a device with `SNS.create_platform_endpoint`. Before this PR, a nil value for `custom_user_data` gets converted to an empty string `””` on AWS side.

Additionally, a device registered with another client (aws cli for example) that truly does have no custom user data incorrectly shows up with the custom_user_data as `””` with ex_aws.

This causes a problem because when re-registering a device (which you are allowed to do), and the custom user data does not match you will get an AWS error.

This PR prevents a nil `custom_user_data` param from getting sent to AWS (which causes the empty string to be set). It also removes the custom_user_data param when parsing a `GetEndpointAttributesResponse` response if there is no custom_user_data (instead of setting it to an empty string).

### DIFF
--- a/lib/ex_aws/sns.ex
+++ b/lib/ex_aws/sns.ex
@@ -157,11 +157,18 @@ defmodule ExAws.SNS do
                                    token :: binary,
                                    custom_user_data :: binary) :: ExAws.Operation.Query.t
   def create_platform_endpoint(platform_application_arn, token, custom_user_data \\ nil) do
-    request(:create_platform_endpoint, %{
+    attrs = %{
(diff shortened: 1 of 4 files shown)


benwilson512:
Hey Ian, sorry I took so long to get back to you. You may be interested in https://elixirforum.com/t/proposal-exaws-2-0/9269. I have included your changes in the new repo https://github.com/ex-aws/ex_aws_sns

PR #464, opened 2017-08-02
Tests: SQS.receive_message wrong argument name

Hinidu:
Hello!
I found the wrong name of an argument while using tests as a source of SQS query samples - `wait_timeout` but it should be `wait_time_seconds`.
Thank you for your work!


benwilson512:
Thanks!

PR #470, opened 2017-08-17
Fixed Elixir 1.5 deprecation warnings

AndrewDryga:
```
==> ex_aws
Compiling 61 files (.ex)
warning: String.rstrip/1 is deprecated, use String.trim_trailing/1
  lib/ex_aws/auth.ex:137

warning: String.strip/1 is deprecated, use String.trim/1
  lib/ex_aws/auth.ex:185

Generated ex_aws app
```


benwilson512:
Thanks!

PR #475, opened 2017-08-31
Add support for query params in S3 key names

mikeastock:
AWS supports ? in key names however the way we were previously encoding
the URL to be signed caused these query params to be dropped.

For more info on what S3 supports in keys see:
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys


mikeastock:
@bruce @benwilson512 Are you guys open to this?

PR #480, opened 2017-09-14
Implement put_bucket_notification

evax:


evax:
@benwilson512 this is an adaptation of #131 to the new ExAws design, integrating your original feedback.
I took the approach of letting AWS validate.
The user is responsible for providing meaningful data, but the format is pretty flexible.
A string will just be passed through and structured data will be converted to XML.

benwilson512:
Hey! Per https://elixirforum.com/t/proposal-exaws-2-0/9269 the ExAws project is splitting each service into its own module to make maintenance easier. Do you mind re-opening this on https://github.com/ex-aws/ex_aws_s3 ?

PR #482, opened 2017-09-18
Cloudwatch PutMetric

danj3:
The PR includes PutMetric for Cloudwatch. Since a single PutMetric can include many individual metrics and dimensions, the approach is compositional and is expected to be used in a pipeline. This PR is primarily to find out if this (or another) approach is palatable and to solicit any comments on alternatives.

An example:

ExAws.Cloudwatch.start_metric_data("Support", "out", 50, "Count", source: "generator")
|> ExAws.Cloudwatch.add_metric_data("in", 47, "Count", source: "network") 
|> ExAws.Cloudwatch.put_metric_data
|> ExAws.request

### DIFF
--- a/lib/ex_aws/cloudwatch.ex
+++ b/lib/ex_aws/cloudwatch.ex
@@ -30,19 +30,29 @@ defmodule ExAws.Cloudwatch do
 
   @type put_metric_t :: { metric_counter :: integer, metrics :: list() }
   
-  @doc "Start a metric request"
-  @spec start_metric_data(namespace :: binary,
-    name :: binary,
-    value :: (binary | integer | float),
-    unit :: binary,
-    dimensions :: [ { dimension_name :: binary, dimension_value :: binary } ]
-  ) :: put_metric_t
+  @doc ~S"""
+  Start a metric request
+
+  iex> ExAws.Cloudwatch.start_metric_data("Support")
+  { 1, [namespace: "Support"] }
+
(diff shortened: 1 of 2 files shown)


benwilson512:
Hey! I really really like this overall approach.

As a first question, I'm a bit unclear about the difference between `start_metric_data` and `add_metric_data`, is that an AWS distinction or one that you have?

danj3:
The start_metric_data establishes the namespace (support in this case) and allows for one metric at the same time, Add just takes more metric data, but can't alter the namespace. The start_ could take _only_ the namespace and then Add just do the metrics, might be clearer. I'm glad it seems generally ok, the AWS api itself is such a mix I'm never quite sure if I have the model quite right.

danj3:
Changed start_metric_data to only take one param (the namespace). Added doctest

Hinidu:
@danj3 Great contribution! We also need this functionality. @benwilson512 what stops you from merging it?

benwilson512:
Hey! I'd be happy to include these changes. As you may have seen via https://elixirforum.com/t/proposal-exaws-2-0/9269 the project is splitting up each service into its own package to improve the maintenance process.

Please re-open this PR on https://github.com/ex-aws/ex_aws_cloudwatch

danj3:
Move to https://github.com/ex-aws/ex_aws_cloudwatch/pull/2

PR #483, opened 2017-09-19
Do not pass negative ms to send_after in AuthCache

antipax:
Currently, if refresh_in determines that the credentials should have already been refreshed, the auth_cache process crashes. While I believe that (functionally) the code after this change has the same effect, the error is no longer logged.


benwilson512:
Perfect, thanks!

PR #493, opened 2017-10-07
Adding support for ExternalId parameter on sts:AssumeRole

courtneycouch:
ExternalId is supported in the type spec but wasn't supported in the option parser. Just added the line in so :external_id can now be used.


benwilson512:
Hey! The ExAws project has split each service into its own project per https://elixirforum.com/t/proposal-exaws-2-0/9269 to ease the maintenance process. Please re-open this PR on https://github.com/ex-aws/ex_aws_sts

PR #496, opened 2017-10-11
Fix double slash on instance metadata queries

greenboxal:
When building paths to retrieve instance metadata, you would end up with double slashes: http://169.254.169.254/latest/meta-data*//*iam/security-credentials/

AWS instance metadata doesn't seem to care about the double slash, but when using services like kube2iam in order to expose IAM profiles to Kubernetes pods, sometimes they do care and won't work.

After merging, can you release a new version? I depend on this and I don't want to be using deps from git.


benwilson512:
Thanks!

PR #502, opened 2017-10-17
Support Fetching Credentials with AssumeRoleCredentials based on role_arn in awscli config

bundacia:
## Description
AWS CLI supports specifying a `role_arn` and `source_profile` in `~/.aws/config` and will issue an `AssumeRoleCredentials` request to STS to get the `secret_access_key` and `access_key_id`. This PR adds the same functionality to this library.

See https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#using-aws-iam-roles for more details.

We were unable to find a good way to add tests for this behavior. It seems that this library doesn't use Mock but instead actually issues requests to live AWS APIs in the tests. This makes it hard to test something like this that requires specific things to be setup in the service (roles, users, permissions, etc). We'd also need a way to test the code with a variety of different `~/.aws/config` setups, which again is hard to do the way the code is structured now without mocks. 

At this point the code that handles `:awscli`-derived credentials is somewhat untested (presumably because of some of the same difficulties). We'd be happy to get some tests around this if you feel it's necessary, but we may need some direction on what approach the approach to testing this sort of thing the maintainers want to take.

Thanks,
@bundacia & @mlagutko


benwilson512:
Hey! This introduces some complications per https://github.com/ex-aws/ex_aws/issues/497

In short, I had been planning to break out STS out as a separate service just like every other service, which means we can't have core functionality like the auth cache using it. I think there's a way around that by just building the operation struct by hand, but I want to look at what our options are AFTER v2.0 lands.

Re: tests. I do try to have all services do at least one actual request to AWS because that ensures the default config actually works. However tests are by no means limited to actual requests, IIRC S3 has several test cases using Bypass. I do hope to make use of bypass more comprehensively in v2.0

bundacia:
Thanks for the quick reply @benwilson512! I guess for v2.0 one option would be to check for the STS service and just skip this behavior if it's not already loaded. What's the timeline for v2.0?

benwilson512:
This week, I'll keep you posted.

bundacia:
Oh, fantastic. I'll hold tight and then rework this for the new version then.

bundacia:
@benwilson512, I've merged in the 2.0 changes to this PR and gotten it working again. It's ready to be re-reviewed. Let me know what you think.

arnomi:
@bundacia @benwilson512 Is there any updated on the state of this pull request? Is it on hold for some reason?

benwilson512:
Yes, this is intrinsically a bit harder after the breakup of ExAws. `Code.ensure_loaded?(ExAws.STS) ` is not a function that should be used at runtime since it's really very slow, and also serializes requests through the code server. Unfortunately however it can't be run at compile time here because then there'd be a circular compilation dependency.

I'm not entirely sure what the right option is here at the moment.

arnomi:
What exactly is the purpose of the `Code.ensure_loaded()` at that point. If it is just to provide a nicer error message in case of misconfiguration maybe you could simply assume that the code is loaded and do a try rescue instead. This was a change that at some point was [introduced into Ecto](https://github.com/elixir-ecto/ecto/commit/46674c3d31f55480e8cedba1e1d2e651cf2ae941)

benwilson512:
If you don't do an ensure loaded check and you don't have `:sts` as a package added, `ex_aws` will always compile with xref errors, since lines like this https://github.com/ex-aws/ex_aws/pull/502/files#diff-66192424a851f96ff0959e4c6da6a224R20 won't point to any valid modules.

arnomi:
True, but shouldn't that just be a compiler warning? And, if I am not mistaken, if you call the module not _directly_ via `ExAws.STS.assume_role(role_arn, role_name, opts)` but instead do `apply(ExAws.Sts, :assume_role, [role_arn, role_name, opts])` you should also not get the compiler warning.

benwilson512:
Yes it's a compiler warning, but I'm disinclined to ship code that contains compiler warnings by default. You can indeed hack it via `apply` but again, I'd prefer a design that could avoid the hacks. Hopefully I'll get a couple hours soon to see if I can think up one at some point soon, but I'm happy to hear alternatives in the meantime.

bundacia:
Do we still get the compiler waring if we list `ExAws.STS` as an *optional* dependency? If not, maybe this would work:

```elixir
if !:erlang.function_exported(ExAws.STS, :assume_role, 3) do
  raise "ExAws.STS required to use `role_arn` in the aws config"
end
```

Though I don't know weather `apply` within a `rescue` is more or less of a hack than this solution.

benwilson512:
It can't be listed as an optional dependency because ExAws.STS already has ExAws as a dependency, and you can't have circular dependencies.

I suppose what I'm really trying to get at is: Any solution here won't be able to have any STS code within the core ExAws package at all.

arnomi:
What about the following approach?

The implementation of `refresh_awscli_config` is made configurable along the following lines

```
def refresh_awscli_config(profile, expiration, ets) do
	refresh_handler().refresh_awscli_config(profile, expiration, ets)
end

defp refresh_handler do
	Application.get_env(ExAws, :refresh_handler, ExAws.AuthCache.RefreshHandler)
end
```

The default implementation of `ExAws.AuthCache.RefreshHandler` is what it is currently.

On `ExAws.STS` we would provide an additional implementation of the RefreshHandler which would be the one from the pull request.

Would that make sense?

arnomi:
@benwilson512 I could look into this change but it would be great if you could give an indication whether the above outline would be a solution you would accept in principle.

benwilson512:
I think that will be ok.

PR #505, opened 2017-10-20
Fix: parsing config causes unused processes to accumulate

vietkungfu:
When `:aws_instance_auth` & `:awscli` config are expired,  `ExAws.Config.AuthCache.get` is called for loading config at runtime. And every time called, some processes have generated and still alive through app running.
That is caused by `StringIO.open(config_string)` is used for reading string then return result without closing it (`StringIO.close(pid)`).
It is fixed and available in `configparser_ex` 2.0.1 version now.

[https://github.com/easco/configparser_ex/releases/tag/2.0.1](https://github.com/easco/configparser_ex/releases/tag/2.0.1)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ This means it will try to resolve credentials in order
 AWS CLI config files are supported, but require an additional dependency:
 
 ```elixir
-{:configparser_ex, "~> 0.2.1"}
+{:configparser_ex, "~> 2.0"}
 ```
 
 You can then add `{:awscli, "profile_name", timeout}` to the above config and it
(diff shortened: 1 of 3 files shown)


rinosamakanata:
This PR is very IMPORTANT. WIth configparser_ex 2.0.0, Erlang processes are increasing unlimitedly as time goes by. That's very serious especially in production operation.

bruce:
Thanks @vietkungfu, @rinosamakanata, et al; I'm sure @benwilson512 will take a look at this in the morning and address it quickly.

rinosamakanata:
Thanks @bruce, say hello to @benwilson512 👍

benwilson512:
@rinosamakanata This PR looks fine, and I will merge. Do know that you can always override dependency versions if you need to do so immediately via:

```
{:configparser_ex, "~> 2.0", override: true}
```

I will also probably backport this to a final 1.1.x release.

rinosamakanata:
@benwilson512 Thanks for your quick merge and letting me know abut `override` option. Backport to 1.1.x will be very welcome. Thank you, too!

PR #512, opened 2017-11-10
Global endpoint config

benwilson512:


lib/ex_aws/config/builder.ex
```diff
@@ -0,0 +1,14 @@
+defmodule ExAws.Config.Builder do
+
+  def get_partitions() do
+    :ex_aws
+    |> :code.priv_dir()
+    |> Path.join("endpoints.json")
```
benwilson512: With this file ExAws now supports every endpoint on every service.

Issue #359, opened 2017-02-13
Handle 30x redirects gracefully


joshprice:
If a bucket is in a different region than configured a 301 redirect is returned. For example:

```
iex> ExAws.S3.get_object("bucket", "filename") |> ExAws.request
** (CaseClauseError) no case clause matching: {:ok, %{body: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>PermanentRedirect</Code>...snip...</Error>", headers: [...], status_code: 301}}
    (ex_aws) lib/ex_aws/request.ex:36: ExAws.Request.request_and_retry/7
    (ex_aws) lib/ex_aws/operation/s3.ex:40: ExAws.Operation.ExAws.Operation.S3.perform/2
```

It'd be nice to handle this more gracefully and perhaps provide some tips on how to configure the defaults to avoid this.

This was raised in #300 but was unrelated to that issue.

joshprice:
Couple of notes now that I've configured my region and it's now working:

- adding a note in the README about default region config would be handy (there wasn't one I could see)
- the redirect should be matched in https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/request.ex#L36-L53 and return a helpful error reminding the user to check their region config
- perhaps trying to follow the redirect as was suggested in #300 but issuing a warning describing that the region is misconfigured?

Happy to put together a PR to achieve the above once the approach is agreed

benwilson512:
> adding a note in the README about default region config would be handy (there wasn't one I could see - maybe I missed something?)
👍 
> the redirect should be matched in https://github.com/CargoSense/ex_aws/blob/master/lib/ex_aws/request.ex#L36-L53 and return a helpful error reminding the user to check their region config
👍 
> perhaps trying to follow the redirect as was suggested in #300 but issuing a warning describing that the region is misconfigured?
👎  Redirects only work for certain AWS resources anyway and are never what you want to be using in production. Fixing the region is like one line of config and solves a whole host of issues.

benwilson512:
In this particular case following redirects would be the wrong choice, closing.

PR #513, opened 2017-11-12
Region of service and common config is ignored

Joe-noh:
Hi, thank you for great product ex_aws.

In `ExAws.Config.build_base/2` region string is passed to `ExAws.Config.Defaults.get/2`, however this comes from `overrides` and falls back to "us-east-1". This behaviour ignores region configuration in `common_config` or `service_config`. leads to getting an error such as reported in #359


benwilson512:
Thanks!

PR #515, opened 2017-11-14
Fix config crash with custom region

elfenlaid:
Description
-----

When `Config` used with custom region, in my case unsupported by Amazon, it crashes with error:

```
ExAws.Config.new(:s3, [host: %{"nyc3" => "nyc3.digitaloceanspaces.com"}, region: "nyc3"])
** (MatchError) no match of right hand side value: nil
    (ex_aws) lib/ex_aws/config/defaults.ex:47: ExAws.Config.Defaults.host/2
    (ex_aws) lib/ex_aws/config/defaults.ex:37: ExAws.Config.Defaults.get/2
    (ex_aws) lib/ex_aws/config.ex:38: ExAws.Config.build_base/2
    (ex_aws) lib/ex_aws/config.ex:29: ExAws.Config.new/2
```

So, now defaults in `Config` should skip host creation in case if missing predefined aws partition.

### Reasoning behind fix
There are couple of aws-api compatible cloud storages out there, for example [DigitalOcean Spaces](https://www.digitalocean.com/products/object-storage/). It would be nice if `ex_aws` were able to work with them.

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -44,11 +44,20 @@ defmodule ExAws.Config.Defaults do
   ]
 
   def host(service, region) do
-    {_, partition} = Enum.find(@partitions, fn {regex, _} ->
-      Regex.run(regex, region)
-    end)
+    if partition = find_partition(region) do
+      do_host(partition, service, region)
+    end
+  end
 
-    do_host(partition, service, region)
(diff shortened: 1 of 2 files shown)


benwilson512:
Hey, thanks for the PR! I took the test case and then did things slightly differently in the config itself: efc83e9

PR #517, opened 2017-11-16
Fixing typo in import statement for README.md

marklee22:
The `:sweet_xml` import is missing an arrow in the docs, which makes it an invalid config setting when copy/pasted. Fixing the typo.


benwilson512:
Thanks!

PR #523, opened 2017-12-09
Add a hint that you can set the region in `ExAws.request/2`

axelson:
This is to help users who are not in `us-east-1` have a hint on how to get up and running faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ end
 With these deps you can use `ExAws` precisely as you're used to:
 
 ```
-# make a request
+# make a request (with the default region)
+ExAws.S3.list_objects("my-bucket") |> ExAws.request
+# or specify the region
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ With these deps you can use `ExAws` precisely as you're used to:
 
 ```
 # make a request
-ExAws.S3.list_objects("my-bucket") |> ExAws.request
+ExAws.S3.list_objects("my-bucket") |> ExAws.request(region: "us-west-1")
```
benwilson512: Let's have both lines here, I don't want folks to think that the region is required.

axelson: Good point. Adding the other line back in.

axelson:
Great, thanks!

PR #531, opened 2018-01-17
add paris region

AlexKovalevych:


benwilson512:
Handled on master, thank you!

Issue #528, opened 2018-01-11
Prefix ex-aws mix tasks with aws.


rmoorman:
Hello,

first of all thank you for this great effort to offer aws integration for elixir! I am just getting started with ex_aws but I really enjoy it already :-)

While reading through the readme I noticed, that there is a `kinesis.tail` mix task made available together with this package. I am aware that you just had a breaking change and I don't know how much of an impact it would have, but wouldn't it be much cleaner to instead have the task named `aws.kinesis.tail` and likewise, put all other (potential) ex-aws tasks inside `aws.`?

Best regards and keep up the great work!

benwilson512:
PR welcome!

rmoorman:
@benwilson512 here is the PR for the prefix but I just noticed that the task doesn't actually do anything but is being advertised within the readme. How would you like to go about that?

PR #534, opened 2018-02-06
Rename kinesis.tail mix task to aws.kinesis.tail

rmoorman:
This pull request prefixes the kinesis.tail mix task with `aws` in order to establish a common prefix for mix tasks of `ex_aws`.

For further reference see #528


benwilson512:
Thanks!

PR #539, opened 2018-02-28
Fix SignatureDoesNotMatch error when retrying a request

frahugo:
Do not mutate URL so recursion works when the URL contains a space.  When retrying, the header was recomputed using a "modified" URL and was at the root of a `SignatureDoesNotMatch` error from S3.


benwilson512:
This looks good! Please fix merge issues, then I'll be happy to merge.

frahugo:
@benwilson512 here we go... Sorry, my change was based on the v2.0.2 tag.  Rebased with master.

benwilson512:
Thanks!

PR #540, opened 2018-03-02
Make headers signed to prevent tampering

jerel:
Currently if you sign an upload URL with ex_aws_s3 that includes query params the query params will be passed in the url but not signed. This means that the actor that receives the url could modify anything in the query params and S3 will accept them. Example:

```
{:ok, url} = ExAws.S3.presigned_url(config, :put, "my_bucket", "foo.jpg", [query_params: [{"content-type", "image/jpeg"}]])
curl -I -XPUT -H "Content-Type: image/png" -T './foo.png' url
200 OK
```

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -52,10 +52,10 @@ defmodule ExAws.Auth do
   def presigned_url(http_method, url, service, datetime, config, expires, query_params \\ [], body \\ nil) do
     with {:ok, config} <- validate_config(config) do
       service = service_name(service)
-      headers = presigned_url_headers(url)
+      signed_headers = presigned_url_headers(url, query_params)
 
       org_query_params = query_params |> Enum.map(fn({k, v}) -> {to_string(k), v} end)
-      amz_query_params = build_amz_query_params(service, datetime, config, expires)
+      amz_query_params = build_amz_query_params(service, datetime, config, expires, signed_headers)
       [org_query, amz_query] = [org_query_params, amz_query_params] |> Enum.map(&canonical_query_params/1)
       query_to_sign = org_query_params ++ amz_query_params |> canonical_query_params
       query_for_url = if Enum.any?(org_query_params), do: org_query <> "&" <> amz_query, else: amz_query
@@ -70,7 +70,7 @@ defmodule ExAws.Auth do
 
       path = uri_encode(path)
 
-      signature = signature(http_method, path, query_to_sign, headers, body, service, datetime, config)
+      signature = signature(http_method, path, query_to_sign, signed_headers, body, service, datetime, config)
       {:ok, "#{uri.scheme}://#{uri.authority}#{path}?#{query_for_url}&X-Amz-Signature=#{signature}"}
     end
   end
@@ -195,18 +195,18 @@ defmodule ExAws.Auth do
     |> Enum.sort(fn {k1, _}, {k2, _} -> k1 < k2 end)
   end
 
-  defp presigned_url_headers(url) do
+  defp presigned_url_headers(url, query_params) do
     uri = URI.parse(url)
-    [{"host", uri.authority}]
+    canonical_headers([{"host", uri.authority} | query_params])
   end
 
-  defp build_amz_query_params(service, datetime, config, expires) do
+  defp build_amz_query_params(service, datetime, config, expires, signed_headers) do
     [
       {"X-Amz-Algorithm",     "AWS4-HMAC-SHA256"},
       {"X-Amz-Credential",    Credentials.generate_credential_v4(service, config, datetime)},
       {"X-Amz-Date",          amz_date(datetime)},
       {"X-Amz-Expires",       expires},
-      {"X-Amz-SignedHeaders", "host"},
+      {"X-Amz-SignedHeaders", Keyword.keys(signed_headers) |> Enum.join(";")},
     ] ++
     if config[:security_token] do
       [{"X-Amz-Security-Token", config[:security_token]}]


benwilson512:
Thanks!

evadne:
I don’t think this is correct and in fact this has caused a regression in our testing because query params are not part of headers, see [documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html).

> `X-Amz-SignedHeaders`: Lists the headers that you used to calculate the signature. The following headers are required in the signature calculations:
> •  The HTTP host header
> • Any x-amz-* headers that you plan to add to the request.

Query Params should not be passed there, and in fact this appears to break URIs that have a custom query param for `content-disposition`, for example.

Queries are already part of the Canonical Request which ExAWS uses for signature computation.

fahchen:
As @evadne says in https://github.com/ex-aws/ex_aws/pull/540#issuecomment-408441438, this pr breaks query params for `content-disposition` and others.

```elixir
{:ok, url} = ExAws.S3.presigned_url(config, :put, "my_bucket", "foo.jpg", [query_params: [{"response-content-disposition", "attachment; filename=hello.txt"}]])
```
When I visit the `url`, it returns `AccessDenied` error (with message `There were headers present in the request which were not signed`).

evadne:
@benwilson512 do you think we can roll this back?

jerel:
If you're getting `AccessDenied` then your client is trying to upload something that does not exactly match the signed url. Here's a note from the docs you linked:

> __Note__
>
> For added security, you should sign all the request headers that you plan to include in your request.

It appears that boto works this way, signing all query params: https://github.com/boto/boto3/issues/578

However, `presigned_url` could potentially be made more flexible by accepting `X-Amz-SignedHeaders` as a query param [on this line](https://github.com/ex-aws/ex_aws/pull/540/commits/db9a62990a05d65562d48e625e015da5f02007da#diff-eafe6fd43d0b53d25332654962f10b61R209) and only falling back to `host` + `headers` if it was not provided. That would allow a dev to choose to sign some and not others.

I don't think reverting this change is the answer as that means all presigned urls are open to query param tampering. If the current behavior is undesirable for your use case I don't believe there's anything preventing you from signing the basic url and then manually concatenating the extra (insecure) query params that you want added before using it.

fahchen:
Sorry for typo, I used it to generate `get` method (not `put` method) presigned_url.
Here is my actual case:
I used `ExAws.S3.presigned_url/5` in `ex_aws_s3`(https://github.com/ex-aws/ex_aws_s3/blob/master/lib/ex_aws/s3.ex#L870-L883) to generate the url, and got `AccessDenied` error when I visited it.

### generated by ex_aws_s3:
first/b1c0eeed-8aa6-4f42-b50a-fd9ca68cf67e/importing/022510fc-f902-4f57-9e78-ac5cef2ac77c_users-importing-blank-data.xlsx?response-content-disposition=attachment%3B%20filename%3Da.xlsx&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minio%2F20180807%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20180807T032249Z&X-Amz-Expires=300&**X-Amz-SignedHeaders=host%3Bresponse-content-disposition**&X-Amz-Signature=2fbd058c8fd75ff456f3faf999e6f3d3fb0b7aa58cf665b1006bd578a280f8e0

### generated by aws-sdk-ruby which works:
/first/b1c0eeed-8aa6-4f42-b50a-fd9ca68cf67e/importing/022510fc-f902-4f57-9e78-ac5cef2ac77c_users-importing-blank-data.xlsx?response-content-disposition=attachment%3B%20filename%3Dhello.xlsx&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minio%2F20180804%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20180804T142758Z&X-Amz-Expires=900&**X-Amz-SignedHeaders=host**&X-Amz-Signature=a6f74f4659c2491be29eac87c8b6c2c57b0f9583a69d7b176b54a26458f7cac0

evadne:
@jerel as I mentioned earlier:

> Queries are already part of the Canonical Request which ExAWS uses for signature computation.

boto/boto3#578 as you have linked to is an Issue, not a Pull Request, and it is dated 2016. Furthermore, boto3 is not the project that handles signature generation, which is handled by botocore in `auth.py`:

```
        # Note that we're not including X-Amz-Signature.
        # From the docs: "The Canonical Query String must include all the query
        # parameters from the preceding table except for X-Amz-Signature.
        signed_headers = self.signed_headers(self.headers_to_sign(request))
```

```
    def headers_to_sign(self, request):
        """
        Select the headers from the request that need to be included
        in the StringToSign.
        """
        header_map = HTTPHeaders()
        for name, value in request.headers.items():
            lname = name.lower()
            if lname not in SIGNED_HEADERS_BLACKLIST:
                header_map[lname] = value
        if 'host' not in header_map:
            header_map['host'] = self._canonical_host(request.url)
        return header_map
```

See https://github.com/boto/botocore/blob/develop/botocore/auth.py

As it currently stands this project breaks builds at tip.

lukeses:
@benwilson512 @evadne 
Should we create an issue with this? I'm having the same issue as: @fahchen when trying to get presigned url with `response-content-disposition`.

I am not sure, is it possible to somehow pass: "X-Amz-SignedHeaders" to `presigned_url` function? For now, I plan to use a version, just before this change.

benwilson512:
@evadne @fahchen to make sure I understand correctly, aws sdk ruby as well as boto only do the host header?

archSeer:
(Sorry for the comment chain)

I think a confusion comes from the original example:

```
{:ok, url} = ExAws.S3.presigned_url(config, :put, "my_bucket", "foo.jpg", [query_params: [{"content-type", "image/jpeg"}]])
curl -I -XPUT -H "Content-Type: image/png" -T './foo.png' url
200 OK
```

`-H` in this case is specifying a header, **not** a query parameter; `url?content-type=image/jpeg` would have been the query parameter.  A possible fix would be to have `presigned_url` accept both headers and query parameters. Query parameters would be signed as a part of the URL, and headers would get signed and appended into `X-Amz-SignedHeaders`.

archSeer:
Ok, so I went ahead and fixed it, presigned_url now supports specifying both headers and query parameters.

```elixir
{:ok, url} = ExAws.S3.presigned_url(config, :put, "my_bucket", "foo.jpg", [
  headers: [{"content-type", "image/jpeg"}],
  query_params: [partNumber: 1, uploadId: "my-id"]
])
```

https://github.com/ex-aws/ex_aws/pull/593
https://github.com/ex-aws/ex_aws_s3/pull/39

benwilson512:
Is this a breaking change or will existing code still worK?

jerel:
> A possible fix would be to have presigned_url accept both headers and query parameters.

From my current understanding I believe that would be the best solution. So using the original example:

``` elixir
{:ok, url} = ExAws.S3.presigned_url(config, :put, "my_bucket", "foo.jpg", [
  query_params: [
    {"uploadId", "ABC"},
    {"partNumber", "1"}],
  headers: [
    {"content-type", "image/jpeg"}
  ]])
```
Does that make sense to everybody involved?

archSeer:
@benwilson512 I've added `headers` as an extra parameter onto `presigned_url` with a default of `[]`, so the signature isn't ideal, but it's backwards compatible.

```elixir
presigned_url( http_method, url, service, datetime, config, expires, query_params \\ [], body \\ nil, headers \\ [])
```

(ideally it would be this)

```elixir
presigned_url( http_method, url, service, datetime, config, expires, query_params \\ [], headers \\ [], body \\ nil)
```

On the ex_aws_s3 side, the function now takes a new option key `headers`, and defaults to `[]` so it's also backwards compatible.

I've tested the code, and it fixes ex-aws/ex_aws_s3#36

PR #548, opened 2018-03-28
handle 301 return from AWS

andyleclair:
Hi! 

I'm not sure how you feel about adding `Mock` as a dependency, but I find it easiest to use for mocking things like HTTP requests to AWS. I'm more than happy to change it if you're not pleased.


benwilson512:
Hey Andy I appreciate the PR. I definitely agree that the test suite should be expanded, but I am significantly opposed to mocking as conducted by the `Mock` library. http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/ sums up my stance on it pretty well.

andyleclair:
@benwilson512 how do you feel about Mox? I could see mocking the `ExAws.Request.HttpClient` behavior working out decently, given the update at the top of the "Mocks and Explicit Contracts" article

andyleclair:
I did start off writing the tests by writing a `Test.HttpClient` but `Mock` was the lazy man's option

benwilson512:
Yup understood, I think `Mox` would work great with the http client behaviour, I think that's be a great way to go

andyleclair:
Great, I'll update this PR

andyleclair:
@benwilson512 how's that look to you?

benwilson512:
Looks great, thanks!

PR #554, opened 2018-04-27
Remove redundant uppercasing

yawaramin:
The called function already uppercases.


benwilson512:
Thanks, but the function may still be called directly, and it needs to work in such a case.

yawaramin:
To clarify, do you mean that `method_string` needs to work if called directly? If so, that is guaranteed because it does a string uppercase as its last operation already: https://github.com/ex-aws/ex_aws/blob/e96660f3f613bddea0597dfb267c5f8c9c63589f/lib/ex_aws/auth/utils.ex#L35-L39

I am proposing just to not do it again in `build_canonical_request`.

PR #560, opened 2018-05-21
Explicitly use fully qualified domain names

discordianfish:
The DNS names here are all fully qualified names but without the
trailing dot, they are eligible to search path expansion.

While most systems attempt to resolve a name as fully qualified first if
it contains at least one dot, this isn't always the case.

Changing the name to include a trailing dot will always resolve the name
without adding a search path.

Kubernetes for example sets ndots to 5 by default, causing every ex-aws
request to resolve 5 names:

```
Trying "firehose.us-east-1.amazonaws.com.master.svc.int.example.com"
Trying "firehose.us-east-1.amazonaws.com.svc.int.example.com"
Trying "firehose.us-east-1.amazonaws.com.int.example.com"
Trying "firehose.us-east-1.amazonaws.com.ec2.internal"
Trying "firehose.us-east-1.amazonaws.com"
```


benwilson512:
Instead of modifying the endpoints.exs file I think it would be best to make this change via the code that does the HTTP requests. I'm trying to keep the endpoints.exs file as a very simple Elixir-ification of the endpoints.json file from the go sdk. PR would definitely be welcome for this latter change, thank you!

PR #566, opened 2018-06-14
Change .to_existing_atom to .to_atom method for parse_ini_file

luizfonseca:
I am having an issue with this method, that always return an empty Map even when I have a `.aws/config` file present.

Given the following file:
```
[profile test]
source_profile=credential_name
role_arn=arn:aws:iam:::role/
mfa_serial=arn:aws:iam::::/
```

`ExAws.CredentialsIni.security_credentials("test")` will always return `%{}` when using `String.to_existing_atom`

Changing it to `String.to_atom` fixes the problem and the output is now:

```elixir
%{
    mfa_serial: "arn:aws:iam::::/",
    role_arn: "arn:aws:iam:::role/",
    source_profile: "credential_name"
}
```

### DIFF
--- a/lib/ex_aws/credentials_ini.ex
+++ b/lib/ex_aws/credentials_ini.ex
@@ -33,7 +33,7 @@ if Code.ensure_loaded?(ConfigParser) do
         updated_key =
           key
           |> String.replace_leading("aws_", "")
-          |> String.to_existing_atom
+          |> String.to_atom
 
         {updated_key, val}
       end)


luizfonseca:
Any updates on this @benwilson512 ?

benwilson512:
I definitely want to fix the issue you're seeing, but I'm loath to introduce a `to_atom` call. `to_existing_atom` should raise when there isn't an atom, I'm surprised that you're getting nil instead of an exception. Maybe on of the config refactoring PRs made that change. I would consider a PR to remove the whole `to_atom` notion entirely and instead just build a map of keys we care about.

luizfonseca:
Ah, ok. I get your point. So wouldn't be the case to have use the `~a()` sigil instead of the word list? Specifically here, in `ExAws.CredentialsIni`. That way we can still keep using `.to_existing_atom`?

```elixir
  @valid_config_keys ~w(
    aws_access_key_id aws_secret_access_key aws_session_token region
    role_arn source_profile credential_source external_id mfa_serial
    role_session_name
  )
```

benwilson512:
On further review, the possibility space here is already constrained by the Map.take, so the `to_atom` is safe.

PR #571, opened 2018-06-29
Include profile in ETS key used for :awscli auth cache

jwarlander:
This enables per-profile caching, instead of re-using the credentials for the
previous profile that was used (even if a *different profile* is passed to
`ExAws.request` at runtime).


bernardd:
Hey @jwarlander - first, thanks for the contribution and sorry it got left on the pile for so long. At a glance this looks like a perfectly reasonable change - I've taken the liberty of resolving the conflicts.

PR #573, opened 2018-07-02
Add secretsmanager to endpoints.exs

mhodgson:


nimf:
But it is already there: https://github.com/ex-aws/ex_aws/blob/fb56b347c506d180012bdbd2c131061f39a358e0/priv/endpoints.exs#L1562

benwilson512:
It didn't used to be I think, it got added via a different PR.

PR #576, opened 2018-07-05
Removed xml_builder dependency from ex_aws

thonydam:
Related to this issue: https://github.com/ex-aws/ex_aws/issues/575


benwilson512:
Thanks!

PR #585, opened 2018-09-27
Allow :region to be retrieved at runtime

smaximov:
This will allow the value of `:region` to be retrieved at runtime when set using the `{:system}` and/or `{:awscli}` config styles:

``` elixir
config :ex_aws,
  # ...
  region: [
    {:system, "AWS_REGION"},
    {:awscli, "default", 30}
  ]
```


smaximov:
Actually, the `{:awscli}` style doesn't work because  the corresponding `retrieve_runtime_value` clause returns a map instead of a single value.

aks:
Please see PR https://github.com/ex-aws/ex_aws/pull/651 for comparison

PR #588, opened 2018-10-11
Adapter based AuthCache and including caches in project supervision tree

kingoftheknoll:
I'm working on a project where we are continually harvesting infrastructure data from AWS across hundreds of accounts. This is certainly not the typical use case since most projects USE Amazon, we interrogate it. So part of making requests is fetching and decrypting credentials. If we only used standard access/secret credentials this wouldn't be as much of an issue because I can pass them to `ExAws.request` as options but AssumedRole and STSAssumed role are very much used.

While my specific use case with lots of accounts is rare, I think there's some shared commonality with those fetching credentials from Vault, etcd or some other secret store. For me, I want to spin up a process for account under a dynamic supervisor with the child_spec matching the credential type of the technique I'm using. Then when calling `ExAws.request` I can pass in an option for the pid or MFA to call to retrieve and merge the credentials. The pid I would look up from my registry.

I think this pattern has the added benefit of breaking up the current AuthCache into more specific modules. A user would add `ExAws.Config.ApiKeys` or `ExAws.Config.STSAssumedRole` or `ExAws.Config.CLIProfile` to their supervision tree depending on their credential type. The community could then contribute adapters for third party services. You could also get the same clean interface with `ExAws.request` if your cache is given a standard name via `GenServer.start_link` which would be implicitly called for the standard user. Having a user add the cache to their supervision tree would allow ExAws to eventually just ship as a library rather than an application. But I do think this can be done without breaking changes.

I threw together a quick example of my idea for illustrative purposes. I'm not a big fan of the MFA call but I'm not sure of a better way at the moment since there's not a single module to act as an interface at the call site.

Thanks so much for your hard work on this great project and I look forward to your feedback!

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -52,6 +52,13 @@ defmodule ExAws.Config do
     |> Map.merge(overrides)
   end
 
+  def retrieve_runtime_config(%{auth_cache: [m, f, a]} = config) do
+    cache_config = Kernel.apply(m, f, a)
+    |> Map.take(@common_config)
+
+    Map.merge(config, cache_config)
+  end
+
   def retrieve_runtime_config(config) do
     Enum.reduce(config, config, fn
       {:host, host}, config ->
--- a/lib/ex_aws/config/adapter.ex
+++ b/lib/ex_aws/config/adapter.ex
@@ -1,0 +1,18 @@
+defmodule ExAws.Config.Adapter do
+  @moduledoc false
+
+  @doc "Retrieve config options"
+  @callback get(pid(), config :: map()) :: map()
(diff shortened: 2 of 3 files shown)


hosh:
Well rats, I wished this got developed. I have a need for AssumeRoleWithWebIdentity and requires querying STSAssumeRole. Chances are, AWS will continue to come up with more auth methods.

bernardd:
Hey - since this doesn't seem to have progressed in a few years, I'm going to close it. @kingoftheknoll thanks for the thoughtful contribution - I'd definitely be interested in seeing it developed, and would be happy to help out if you wanted to progress it.

PR #590, opened 2018-10-16
Added inets HTTP client (:httpc the Erlang built-in HTTP client)

edescourtis:
* Uses a profile called `:ex_aws`

NOTE: Switch to `:httpc` like this:

```
config :ex_aws,
  http_client: ExAws.Request.Httpc
  ...
```

### DIFF
--- a/lib/ex_aws/request/httpc.ex
+++ b/lib/ex_aws/request/httpc.ex
@@ -1,0 +1,66 @@
+defmodule ExAws.Request.Httpc do
+  @behaviour ExAws.Request.HttpClient
+
+  @ex_aws_profile :ex_aws
+
+  @moduledoc """
+  Configuration for :httpc
+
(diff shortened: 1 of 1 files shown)


edescourtis:
By the way, I get much better performance under load on DynamoDB and CloudWatch with `httpc` for some reason.

benwilson512:
Hey @edescourtis, thanks for the PR!

Unfortunately I'm very hesitant to provide an Httpc adapter within ExAws, because `:httpc` does not validate ssl, and is thus completely insecure. Data transmission to AWS is critical, and I don't think it's wise to hand people an adapter that does that transmission insecurely by default.

Hackney can perform well if the pool and keep alive is enabled. The pool is disabled by default because earlier versions of hackney could end up with deadlocks when using the pool. Hackney has undergone quite a few changes since I last looked at it though, so perhaps the pool is back to working properly.

PR #593, opened 2018-10-26
Add an additional parameter to ExAws.Auth for specifying headers.

archseer:
Addresses ex-aws/ex_aws_s3#36, see my comments on #540 for the rationale.

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -64,11 +64,12 @@ defmodule ExAws.Auth do
         config,
         expires,
         query_params \\ [],
-        body \\ nil
+        body \\ nil,
+        headers \\ []
       ) do
     with {:ok, config} <- validate_config(config) do
       service = service_name(service)
-      signed_headers = presigned_url_headers(url, query_params)
+      signed_headers = presigned_url_headers(url, headers)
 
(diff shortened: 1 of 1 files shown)


jerel:
I believe this is a breaking change in the sense that if someone today is using `query_params` to pass things like `content-type` in the URLs they generate, after they upgrade S3 will cease verifying the value of the `Content-Type` header at upload time until they add it to the new `headers` option.

I think it's a good fix regardless... not sure if there'd be a way to make it fully backward compatible but I can't think of any good ways. One ugly solution might be to temporarily emit a warning if `query_params` is provided and `headers` is not.

archSeer:
Well I mean, 540 was a breaking change as well so this is just bringing ex_aws back to where it was. Given that the URLs it generates still work, and that the original documentation never mentioned headers, I think it's fine. -- I'll add some tests later this week and document the new option, just so the documentation is a bit clearer.

tjshipe:
Was this merged in a different PR? I found this one: https://github.com/ex-aws/ex_aws_s3/pull/39 but it's also open. I'm looking to set a `content-disposition` header to set a filename on download. I think one of these PRs would solve my issue. Is there anything I can help with to get these merged in?

benwilson512:
Thanks, this seems like the least intrusive solution.

PR #595, opened 2018-10-31
Defaults for SageMaker Runtime service

dideler:
Prepares ex_aws to be used with the SageMaker Runtime service.

It's one of those exceptional services - it's like a subservice of SageMaker.

- Adds an alias for the service name to make it easier to work with
- Override base defaults with `sagemaker` service defaults

Resolves https://github.com/ex-aws/ex_aws/issues/594

### DIFF
--- a/priv/endpoints.exs
+++ b/priv/endpoints.exs
@@ -1066,6 +1066,7 @@
           }
         },
         "runtime.sagemaker" => %{
+          "defaults" => %{"credentialScope" => %{"service" => "sagemaker"}},
           "endpoints" => %{
             "ap-northeast-1" => %{},
             "eu-west-1" => %{},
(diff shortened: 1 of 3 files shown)


priv/endpoints.exs
```diff
@@ -1066,6 +1066,7 @@
           }
         },
         "runtime.sagemaker" => %{
+          "defaults" => %{"credentialScope" => %{"service" => "sagemaker"}},
```
dideler: This service uses its "parent" service in the credential scope for the authorisation header.

Note that ex_aws is ignoring this value right now, it's only here for documentation purposes.

The only default that has an effect right now is `%{"defaults" => %{"hostname" => hostname}}`, see `ExAws.Config.Defaults.do_host/3`. A good next step would be to honour other defaults that are specified here.

benwilson512: This file is autogenerated, so editing it isn't really viable. This kind of mapping will have to go elsewhere.

dideler: I'll remove this change as the service override in `ExAws.Config.Defaults.defaults(:sagemaker_runtime)` solves the problem of having a different service name in the credential scope.

Good to know that this file is autogenerated. I couldn't find anything in the docs to suggest that. Is this the source? https://github.com/aws/aws-sdk-go/blob/master/models/endpoints/endpoints.json

I'll try moving this specific change upstream.

PR #601, opened 2018-11-16
Support content-encoding in ExAws.Operation.Query

Hinidu:
Hello!
I need this functionality for CloudWatch PutMetricData API. It has 40KB limit on the payload size but allows GZIP compression to fit bigger payloads.
PR to `ex_aws_cloudwatch` is coming.

### DIFF
--- a/lib/ex_aws/operation/query.ex
+++ b/lib/ex_aws/operation/query.ex
@@ -10,6 +10,7 @@ defmodule ExAws.Operation.Query do
 
   defstruct path: "/",
             params: %{},
+            content_encoding: "identity",
             service: nil,
             action: nil,
             parser: &ExAws.Utils.identity/2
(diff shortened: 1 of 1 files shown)


lib/ex_aws/operation/query.ex
```diff
@@ -10,6 +10,7 @@ defmodule ExAws.Operation.Query do
 
   defstruct path: "/",
             params: %{},
+            content_encoding: "identity",
```
benwilson512: I think it'd be better to just provide a generic `headers` key here.

Hinidu: And look for `content_encoding` in these headers in `perform` to decide whether we should compress `data` or not?
If it's ok for you then I can do that.

Hinidu: @benwilson512 could you please answer my question above? I'm already using this code for more than a month in production and I would be glad to update this PR so it could be merged.

Hinidu:
Thank you for merging!

PR #603, opened 2018-11-23
Accept sign_headers

cschilbe:
Issue: https://github.com/ex-aws/ex_aws/issues/602

Depends on: https://github.com/ex-aws/ex_aws_s3/pull/43

Accept `sign_headers` in addition to `query_params` to allow for query params that will not also be sent in the headers of the request.

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -64,11 +64,12 @@ defmodule ExAws.Auth do
         config,
         expires,
         query_params \\ [],
+        sign_headers \\ [],
         body \\ nil
       ) do
     with {:ok, config} <- validate_config(config) do
       service = service_name(service)
-      signed_headers = presigned_url_headers(url, query_params)
+      signed_headers = presigned_url_headers(url, sign_headers)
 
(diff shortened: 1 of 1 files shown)


tjshipe:
I tested this out and it works for my use case!

I needed the ability to set a `response-content-disposition` query param on the presigned url to change the filename on download. This PR allows that functionality.

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html#RESTObjectGET-requests

benwilson512:
Per the comment on https://github.com/ex-aws/ex_aws/pull/555, we can't change the order of the parameters to this function because it introduces a major breaking change. Are you able to comment on what differences it has from https://github.com/ex-aws/ex_aws/pull/555 ?

cschilbe:
Maybe we could introduce another function that takes this parameter? #555 looked like it was not the best way to handle it.

netflakes:
I would like to send a header `response-content-disposition: inline` to ensure images are displayed in the browser rather than downloading them - I take this ticket needs to be resolved before that is possible?

PR #604, opened 2018-11-26
Properly set defaults for iot_data service

mattlorey:
Currently trying to do an operation with the `iot_data` service causes the following runtime error:
`(RuntimeError) iot_data not found in partition aws`

This properly maps `iot_data` to `data.iot`  in the endpoints file and overrides the service to use `iotdata`.

Followed the conventions for `dynamodb_streams` and `sagemaker_runtime`


benwilson512:
Thanks:!

PR #605, opened 2018-11-30
Generalize json RPC error parsing

Hinidu:
Support errors with `__type` without `#` and `Message` with capital `M`
i.e.
```
{
  "__type": "InvalidRequestException",
  "Message": "Matchmaking ticket not found ...."
}
```

The actual method is taken from official Go AWS library (split `__type`
up to 2 parts and take the last one, also Go json parser is case
insensitive).

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -100,12 +100,15 @@ defmodule ExAws.Request do
 
   def client_error(%{status_code: status, body: body} = error, json_codec) do
     case json_codec.decode(body) do
-      {:ok, %{"__type" => error_type, "message" => message} = err} ->
-        error_type
-        |> String.split("#")
-        |> case do
-          [_, type] -> handle_aws_error(type, message)
-          _ -> {:error, {:http_error, status, err}}
+      {:ok, %{"__type" => error_type} = err} ->
+        type =
+          error_type
+          |> String.split("#", parts: 2)
+          |> List.last
+
+        case Map.get(err, "message") || Map.get(err, "Message") do
+          nil -> {:error, {:http_error, status, err}}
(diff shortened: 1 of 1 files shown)


Hinidu:
I've just found that AWS sometimes returns errors in format `%{"__type" => "InvalidRequestException"}` without the message or any other fields except `__type`. I think it would be handier to handle them like errors with empty messages (again just like the official Go SDK does) than ugly `{:http_error, 400, %{"__type" => "InvalidRequestException"}}`. I'm going to update this PR to handle these errors as soon as I get a little bit of time to do that.

benwilson512:
This is largely a holdover from when I translated this library from erlcloud. The solution I'd like to have hands off all errors to the operation struct so that each service can handle those errors.

Hinidu:
Why? I don't think that your desired solution is reasonable because errors from all services are very similar and even official AWS libraries parse them that way.

benwilson512:
@Hinidu you're confusing operation structs with services. Each operation struct represents a whole category of services, there's only like 4 or 5 total. This is exactly the right place to capture error conventions since the error types are definitely different for a JSON service vs a REST vs xml

Hinidu:
Ah, I see. I've misunderstood "so that each service can handle those errors."

benwilson512:
My language was definitely inconsistent, sorry!

bernardd:
I believe that the original issue this PR handled (errors with a capital-M `Message` field) is now resolved in `master` by https://github.com/ex-aws/ex_aws/commit/22e2494351e0e620602f1c613fb10f3cf4be192e.
I'm going to close this, but @Hinidu if you still need a change to be made to handle errors with no message field at all, as you suggest, please do feel free to open a fresh PR.

PR #607, opened 2018-12-05
"request" function return value

AsharDweedar:
the return value from request function is the same regardless the `config_overrides` value


benwilson512:
Thanks!

PR #608, opened 2018-12-10
Describe direct usage

angelikatyborska:
I don't have much experience using the AWS HTTP API directly except for figuring out how to run an ECS task with this library, so I would appreciate a thorough review. The contributing guidelines about adding new service modules were quite helpful.

I found four different implementations of `ExAWS.Operation`: `JSON`, `Query`, `RestQuery`, and `S3`. The example of `JSON` is from my own app, the example of `Query` is from the contributing guidelines.

I didn't mention `S3` because there is a service module for S3.

I didnt' mention `RestQuery` because I do not know for which AWS service it could be used. In https://github.com/aws/aws-sdk-go/tree/master/models/apis (mentioned in the contributing guidelines), I only found protocols "query", "json", and "rest-json" - not "rest-query". If somebody gives me a hint about this, I will update the PR.

The PR also includes unrelated fixes of typos I found while reading the `ExAWS.Operation` module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,64 @@ config :ex_aws,
 
 ExAws can also be used directly without any specific service module.
 
---TODO--
+You need to figure out how the API of the specific AWS service works, in particular:
+- Protocol (JSON or query).
+- Path (depends on the service and the specific operation, usually "/").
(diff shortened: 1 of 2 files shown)


benwilson512:
Thanks!

PR #609, opened 2018-12-17
Log error and details of the failed request

ishikawa:
Currently, ExAWS reports errors with `Logger.warn/2`, but we want to know what the request failed. So this PR adds request information if `config[:debug_requests]` enabled.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -32,9 +32,11 @@ defmodule ExAws.Request do
       safe_url = replace_spaces(url)
 
       if config[:debug_requests] do
-        Logger.debug("Request URL: #{inspect(safe_url)} ATTEMPT: #{attempt}")
-        Logger.debug("Request HEADERS: #{inspect(full_headers)}")
-        Logger.debug("Request BODY: #{inspect(req_body)}")
+        Logger.debug(
+          "ExAws: Request URL: #{inspect(safe_url)} HEADERS: #{inspect(full_headers)} BODY: #{
+            inspect(req_body)
+          } ATTEMPT: #{attempt}"
+        )
       end
 
       case config[:http_client].request(
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -86,9 +86,7 @@ defmodule ExAws.Request do
 
         {:error, %{reason: reason}} ->
           Logger.warn(
-            "ExAws: HTTP ERROR: #{inspect(reason)} for URL: #{inspect(safe_url)} HEADERS: #{
-              inspect(full_headers)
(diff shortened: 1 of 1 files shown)


pmenhart:
Can you log also `attempt`, please? Knowing whether the same request failed several times would be very useful.

ishikawa:
@pmenhart Thanks for your suggestion! Committed 68846bb 😉

Hinidu:
Good job! I also thought about adding something like this. But why do you add these things to the log only when `debug_requests` is enabled? It looks for me like the vital information in production environments too.
BTW I would also like to change the way `debug_requests` performs logging now - it's quite strange to log one request using 3 calls to `Logger` in such highly concurrent environment like erlang VM, these logs from different requests are quite often interleaving.

ishikawa:
@Hinidu https://github.com/ex-aws/ex_aws/pull/609#issuecomment-450322713

>Good job! I also thought about adding something like this. But why do you add these things to the log only when debug_requests is enabled? It looks for me like the vital information in production environments too.

Maybe I had been too conservative to change the current behavior. So I changed my mind 🙂 bc8030843a213ffb20644c5ec5de568ede8649d4

> BTW I would also like to change the way debug_requests performs logging now - it's quite strange to log one request using 3 calls to Logger in such highly concurrent environment like erlang VM, these logs from different requests are quite often interleaving.

Agreed. I've encountered such a situation in my production environment. 953e7d5daf042d9628ad9c538b02677d62c6ffe0

Please look at the changes if you have a time. Thanks! 😉

Hinidu:
@ishikawa looks great to me! I'm not the maintainer of this repo but I hope that @benwilson512 will give you more feedback or merge this PR as it is now.

lib/ex_aws/request.ex
```diff
@@ -83,7 +85,11 @@ defmodule ExAws.Request do
           )
 
         {:error, %{reason: reason}} ->
-          Logger.warn("ExAws: HTTP ERROR: #{inspect(reason)}")
+          Logger.warn(
+            "ExAws: HTTP ERROR: #{inspect(reason)} for URL: #{inspect(safe_url)} HEADERS: #{
+              inspect(full_headers)
+            } BODY: #{inspect(req_body)} ATTEMPT: #{attempt}"
```
benwilson512: Inspecting the body can be problematic here as it can be arbitrarily large. We can maybe print the byte size of the body, but `inspect(req_body)` is guaranteed to cause major issues.

ishikawa: @benwilson512 Thanks! You're right. I also thought about request headers which can contain sensitive information (e.g. Authorization header). So I just removed both contents from logging. bd058c0

benwilson512:
Thanks!

PR #618, opened 2019-02-18
Add eu-north-1

daveli:
Add eu-north-1 based on https://docs.aws.amazon.com/general/latest/gr/rande.html.


benwilson512:
Thanks!

PR #621, opened 2019-03-04
Add a behaviour for the toplevel functions

the-mikedavis:
I found [this discussion](https://github.com/ex-aws/ex_aws_s3/issues/27) in the s3 repo, but it looks like nothing ever came of it. I think it's pretty important to be able to Mox pretty much anything that one would do with s3 for testing. Thanks to how you've set up the library, it's such a small change to make that possible!


benwilson512:
Thanks!

PR #625, opened 2019-04-01
exponential backoff for InstanceMeta rate limit

paulswartz:
Periodically, our ECS instances will crash with the following error:
```
** (RuntimeError) Instance Meta Error: HTTP response status code 429

Please check AWS EC2 IAM role.
```

The roles are fine, but something has made too many requests to the instance metadata endpoint. This PR re-uses the retry logic from ExAws.Request to retry these InstanceMeta requests as well.

### DIFF
--- a/lib/ex_aws/instance_meta.ex
+++ b/lib/ex_aws/instance_meta.ex
@@ -11,19 +11,29 @@ defmodule ExAws.InstanceMeta do
   # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html
   @task_role_root "http://169.254.170.2"
 
+  alias ExAws.Request
+
   def request(config, url) do
+    do_request(config, url, 1)
+  end
+
+  defp do_request(config, url, attempt) do
     case config.http_client.request(:get, url, "", [], []) do
       {:ok, %{status_code: 200, body: body}} ->
         body
 
-      {:ok, %{status_code: status_code}} ->
-        raise """
-        Instance Meta Error: HTTP response status code #{inspect(status_code)}
+      {:ok, %{status_code: 429}} ->
+        case Request.attempt_again?(attempt, :rate_limit, config) do
+          {:attempt, attempt} ->
+            do_request(config, url, attempt)
+          {:error, _} ->
+            raise_status_error(429)
+        end
 
-        Please check AWS EC2 IAM role.
-        """
+      {{:ok, %{status_code: status_code}}, _} ->
+        raise_status_error(status_code)
 
-      error ->
+      {error, _} ->
         raise """
         Instance Meta Error: #{inspect(error)}
 
@@ -41,6 +51,14 @@ defmodule ExAws.InstanceMeta do
         ```
         """
     end
+  end
+
+  defp raise_status_error(status_code) do
+    raise """
+    Instance Meta Error: HTTP response status code #{inspect(status_code)}
+
+    Please check AWS EC2 IAM role.
+    """
   end
 
   def instance_role(config) do
--- a/test/ex_aws/instance_meta_test.exs
+++ b/test/ex_aws/instance_meta_test.exs
@@ -1,5 +1,5 @@
 defmodule ExAws.InstanceMetaTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
 
   import Mox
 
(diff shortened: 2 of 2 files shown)


benwilson512:
Hm, I'm surprised to see this is happening. Are you explicitly calling these functions in your application code? ExAws itself should not be calling instance metadata very frequently, just to update tokens.

paulswartz:
You are not the only one surprised :) Our only ExAWS call is `ExAws.S3.put_object`, but we're calling that very frequently (~4x a second). We don't see the crash very frequently, either: only 7 times in the last 3 months.

benwilson512:
@paulswartz Definitely weird. Regular ExAws calls shouldn't even hit the instance metadata more than once every 15 minutes to refresh tokens regardless of how many requests to AWS you're making. How is ExAws configured?

paulswartz:
From the crash message:
```
Last message: {:refresh_config, %{access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :instance_role], host: "s3.amazonaws.com", http_client: ExAws.Request.Hackney, json_codec: Jason, port: 443, region: "us-east-1", retries: [max_attempts: 10, base_backoff_in_ms: 10, max_backoff_in_ms: 10000], scheme: "https://", secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, :instance_role]}}
```
The environment does not have `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY`.

benwilson512:
Ah, I have a theory. If there are many concurrent requests then when the tokens expire each of those will send a message to the auth cache to refresh. The auth cache will run a refresh for each of those messages, when it really should be doing it just once and then ignoring the rest. Let me make a PR.

benwilson512:
The flaw in this theory is that the AuthCache, if functioning properly, should be automatically renewing the tokens 5 minutes before the expiration, which should mean individual processes making the requests should never need to call a manual refresh. Maybe there's something to do with Fargate and clocks here?

Ultimately, I'm hesitant to simply place a back off here because the auth cache blocking will start to block every other process that's waiting on the auth to refresh. The root issue here is making too many instance metadata requests, and that's the thing I want to try to figure out.

paulswartz:
That's fair: I wanted to open the PR to at least start a discussion and get some feedback.

I'm running this branch internally, so if we run into other issues (or continue to see crashes) I'll be sure to let you know.

ryosan-470:
I have met the same issue on AWS Fargate. I've contacted AWS support why this event happened, however they didn't tell me the reasons. So we cannot detect this root issue.

BlueHotDog:
Hey guys, any news regarding this? just encountered the same issue in prod

rraub:
I think this change is still worth making, it add some safeguards for issues like this which is generally a best practice even if we don't have a root cause yet. Making this library more tolerant of network issues (even local network) is an improvement by itself.

aseigo:
> The flaw in this theory is that the AuthCache, if functioning properly, should be automatically renewing the tokens 5 minutes 
> before the expiration, which should mean individual processes making the requests should never need to call a manual 
> refresh.

I think your theory is correct. Looking at the code, here are two possible way to trigger this:

* several processes call AuthCache.get(config) before the config exists
* due to a race to the return from `GenServer.call` with `{:refresh_config, config}`, they will *all* see the config as missing and *all* issue `refresh_config` messages ... 
* those will all end up calling `refresh_config` which will itself will then also push a `refresh_config` message creating a small storm of requests every (expiration - 5 minutes)

or:

* manual calls to AuthCache.refresh_config are being made which conflict with the regularly scheduled calls, however ...
*  ... in that case this will likely cause other problems as it will send a message to whatever process made the call (`self()`), which probably isn't waiting for that message.

(though I don't see where this could be happening .. perhaps from client code?)

some observations/questions:

*  `handle_call` taking `{:refresh_config, config}` should do an ets lookup for the config and only send a request if it does not exist in the ets table *or* if the expiration is less than 5 minutes in the future.
* AuthCache.refresh_config/2 should be private. nothing should be calling it except AuthCache
* messages should be sent to `__MODULE__` to prevent accidentally sending messages to the wrong process due to where it is called from. Alternatively, a `GenServer.cast(__MODULE__, :schedule_refresh)` could be called which schedules the call, ensuring that the message is *always* created in (and delivered to) the AuthCache process
* calls `iso_z_to_secs` and `now_ins_seconds` could be replaced with: 

```
expiration
|> NaiveDateTime.from_iso8601!()
|> NaiveDateTime.diff(NaiveDateTime.utc_now())
```

if one is concerned that `expiration` may hold an invalid datetime string, there is of course a non `!` version that returns an :ok/:error tuple.

bernardd:
@aseigo 's alternative patch in #737 to resolve this issue has now been merged into `master`. I'd appreciate if anyone seeing this problem is able to try out the fix and let me know how they get on.

bernardd:
I'm going to close this since an alternative fix for the issue has been merged.

PR #631, opened 2019-04-26
add support for lex models and runtime on service overrides

techgaun:
I am working on lex runtime and models services for ex_aws. `models.lex` or `runtime.lex` is prefixed in url and `lex` is used in credential scope. It would nice to get it in the ex_aws.

Also, are we doing any new releases for changes like these? Looks like sagemaker change like this is not in the hex yet.

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -23,6 +23,16 @@ defmodule ExAws.Config.Defaults do
     |> Map.merge(defaults(:dynamodb))
   end
 
+  def defaults(:lex_runtime) do
+    %{service_override: :lex}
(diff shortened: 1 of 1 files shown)


bjunc:
Lex would be great to support out-of-the-box.  We're currently doing something like this to use Runtime Lex:

```elixir
ExAws.request(%ExAws.Operation.JSON{
  service: "runtime.lex",
  http_method: :post,
  path: "/bot/#{bot_name}/alias/#{bot_alias}/user/#{user_id}/#{operation_name}",
  headers: [{"content-type", "application/json"}],
  data: Poison.encode!(data)        
}, [service_override: :lex])
```

benwilson512:
Thanks! I'll get a hex release out today

PR #633, opened 2019-05-09
Support reading the profile from env variable

saverio-kantox:
`aws-cli` supports choosing the profile from the AWS_PROFILE variable. With this change the profile will be resolved on runtime so the same build can be executed with any local profile.

Old config (hardcoded profile name):
```ex
config :ex_aws,
  access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, {:awscli, "default", 30}, :instance_role]
```

dynamic profile selection (from env):
```ex
config :ex_aws,
  access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, {:awscli, :system, 30}, :instance_role]
```

### DIFF
--- a/lib/ex_aws/credentials_ini.ex
+++ b/lib/ex_aws/credentials_ini.ex
@@ -6,14 +6,14 @@ if Code.ensure_loaded?(ConfigParser) do
       role_arn source_profile credential_source external_id mfa_serial role_session_name
     )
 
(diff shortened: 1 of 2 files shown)


bernardd:
Hi Saverio - this looks great. Would you be able to add a note on its availability to the README, please? Then I'll be happy to merge it.

PR #637, opened 2019-05-28
Update HttpClient example to follow @callback

JoeARO:
The HttpClient example was misleading not only for me but I can see in the Issues another user, it led users of ExAws to believe 
that ExAws.Request.HttpClient.request/5 only took /4. Have also added
some extra description to help with the conforming of the Client.request
firing to give a bit more clarity.

### DIFF
--- a/lib/ex_aws/request/http_client.ex
+++ b/lib/ex_aws/request/http_client.ex
@@ -20,11 +20,21 @@ defmodule ExAws.Request.HttpClient do
   ```elixir
   defmodule ExAws.Request.HTTPotion do
     @behaviour ExAws.Request.HttpClient
-    def request(method, url, body, headers) do
+    def request(method, url, body, headers, http_opts) do
       {:ok, HTTPotion.request(method, url, [body: body, headers: headers, ibrowse: [headers_as_is: true]])}
     end
   end
   ```
+
+  When conforming your selected HTTP Client take note of a few things:
+    - The module name doesn't need to follow the same styling as this module it is simply your own 'HTTP Client', i.e. #{project_name}.HttpClient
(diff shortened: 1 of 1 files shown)


benwilson512:
Thanks!

JoeARO:
@benwilson512, worlds quickest merge! Thanks 👍

lib/ex_aws/request/http_client.ex
```diff
@@ -20,11 +20,21 @@ defmodule ExAws.Request.HttpClient do
...
     @behaviour ExAws.Request.HttpClient
-    def request(method, url, body, headers) do
+    def request(method, url, body, headers, http_opts) do
       {:ok, HTTPotion.request(method, url, [body: body, headers: headers, ibrowse: [headers_as_is: true]])}
     end
   end
   ```
+
+  When conforming your selected HTTP Client take note of a few things:
+    - The module name doesn't need to follow the same styling as this module it is simply your own 'HTTP Client', i.e. #{project_name}.HttpClient
```
safwanolaimat: @benwilson512  compile error project_name does not exists

JoeARO: `#{project_name}` is string interpolation so you would replace that with whatever your project is called. 
At the top level of your project you usually have a module with the same name as your project:
```elixir
defmodule MyApp do
end
``` 
You could then have a folder structure similar to this:
my_app.ex
clients
|-> http_client.ex
inside your http_client.ex you would namespace it with your app name:
```elixir
defmodule MyApp.HttpClient do
end
```
In this case `MyApp` is what replaces the `#{project_name}` here.

JoeARO: Sorry it took so long for a reply!

PR #638, opened 2019-05-30
Adding ap-east-1 region

ottolin:
reference: https://docs.aws.amazon.com/general/latest/gr/rande.html


hermeslive:
Can you please update this ticket? Are there anything need to be changed?
We are forced to use a private fork in order to connect to ap-east-1 currently.
Thanks

PR #640, opened 2019-06-06
Allow turning off normalize_path

mgwidmann:
Very old s3 objects can have a leading forward slash and normalizing makes those objects inaccessible. Fixes https://github.com/ex-aws/ex_aws_s3/issues/58


mgwidmann:
Any update on this? Would like to see this or an equivalent PR merged

PR #643, opened 2019-06-19
Allow setting http_opts for use with HTTPotion instead of hackney

mgwidmann:
The example in the documentation does not use the `http_opts` and since it is not part of the common config it is not picked up when set in config.


benwilson512:
Hey @mgwidmann thanks for the PR. It seems like however that your HTTPotion adapter should do the same thing that the `hackney` adapter does and grab those options directly: https://github.com/ex-aws/ex_aws/blob/master/lib/ex_aws/request/hackney.ex#L18

PR #648, opened 2019-08-19
Fix for issue where + are replaced with spaces, even though the + is …

mmarini:
…valid

Fix for issue with trying to access files on s3 that have `+`'s in them: https://github.com/ex-aws/ex_aws/issues/647


benwilson512:
Handled in #660

PR #650, opened 2019-08-28
Fix the json_codec decode method reference when there is a client error

samfrench:
The other places in this codebase use "decode!" but in the `client_error` method of request it calls `decode` which seems to not exist.

```
** (UndefinedFunctionError) function Eljiffy.decode/1 is undefined or private
        (eljiffy) Eljiffy.decode("{\"Message\":\"...\"}")
```

This PR changes the method to be the same as the others and calls `decode!`.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -105,7 +105,7 @@ defmodule ExAws.Request do
   end
 
   def client_error(%{status_code: status, body: body} = error, json_codec) do
-    case json_codec.decode(body) do
+    case json_codec.decode!(body) do
       {:ok, %{"__type" => error_type, "message" => message} = err} ->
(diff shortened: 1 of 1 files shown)


woodyblah:
This doesn't seem right, Poison does have `decode` method ([here](https://github.com/devinus/poison/blob/master/lib/poison.ex#L61)), and that is the whole reason this call is made in a `case` statement matching the `{:ok, _}`

samfrench:
Thank you @pmenhart and @benwilson512 for the feedback. I had only read the Poison readme and the methods of `encode` and `decode` didn't seem to be specified. Looking at the code they were there.

As the methods are present, it makes sense to add the methods to the library we're using or make a wrapper in our own application.

PR #651, opened 2019-09-13
enable system envar spec for region config

aks:
This PR enables the use of an envar `AWS_REGION` to externally control the region configurations of `ex_aws`  usage.  

The PR provides:
- a simple change to the code that determines the region
- two new tests to confirm that plain-text and system envar configurations both work as expected.


benwilson512:
Looks great, thanks!

Issue #632, opened 2019-05-08
Missing / incorrectly configured `json_codec` can cause AWS credential leak in error report


madlep:
### Environment

* Elixir & Erlang versions
```
Erlang/OTP 21 [erts-10.3.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.4 (compiled with Erlang/OTP 21)
```
* ExAws version `mix deps |grep ex_aws`
```
* ex_aws 2.1.1 (Hex package) (mix)
  locked at 2.1.1 (ex_aws) 1e4de210
* ex_aws_s3 2.0.1 (Hex package) (mix)
  locked at 2.0.1 (ex_aws_s3) 9e09366e
```
* HTTP client version. IE for hackney do `mix deps | grep hackney`
```
* hackney 1.15.1 (Hex package) (rebar3)
  locked at 1.15.1 (hackney) 9f8f471c
```

### Current behavior

If `json_codec` hasn't been configured, and `Poison` isn't available, JSON encode/decode calls will fail.

When these failures happen, it can be when decoding responses from AWS with sensitive credentials included - such as a string containing instance role details.

This will then cause an error that will be logged / sent to error tracking service etc.

```
exit:

exited in: GenServer.call(ExAws.Config.AuthCache, {:refresh_config, %{access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, :instance_role], host: "s3.amazonaws.com", http_client: ExAws.Request.Hackney, json_codec: Poison, port: 443, region: "us-east-1", retries: [max_attempts: 10, base_backoff_in_ms: 10, max_backoff_in_ms: 10000], scheme: "https://", secret_access_key: [{:system, "AWS_SECRET_ACCESS_KEY"}, :instance_role]}}, 30000)
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function Poison.decode!/1 is undefined (module Poison is not available)
[...]

PR #657, opened 2019-10-10
Add json_codec configuration example

cmatheson:
I was seeing the same Poison-related errors as #632 but it wasn't immediately clear what to do about it.  Thanks!


benwilson512:
Thank you!

PR #658, opened 2019-10-10
Use Jason as json encoder/decoder

maartenvanvliet:
See https://github.com/ex-aws/ex_aws/issues/632
As Jason is now the more common json encoder, it can be made the default


benwilson512:
Makes sense, thanks!

ssorallen:
This change broke our build process because we didn't expect a default codec change in a patch release and our library does not currently list Jason as a dependency. This would have been a good reason to do a minor version change.

The documentation also still says that Poison is the default: https://github.com/ex-aws/ex_aws#json-codec-configuration

In case anyone else comes across this thread, we enforced the previous Poison encoder in our config.exs:

```
config :ex_aws,
  json_codec: Poison
```

I do appreciate the great library, thank you for your work.

PR #660, opened 2019-11-08
Fix URL encoding for S3 objects with special characters

dbernheisel:
# Problem

I was getting errors when trying to download certain files with special characters from S3 such as `+` and `#`

# Solution

The root issue is that URLs were being parsed and losing some information from the URL before signing and requesting the URL. For example, if an end-user uploads a file to S3 with this filename: 
`test hello+#3.txt`, it's not URL-encoded correctly and would fail a normal web request, saying either the key didn't exist, or a 403. We need that URL-encoded to `test%20hello%2B%233.txt`. 

When parsing the URL, we currently use Elixir's built-in URI functions:

```elixir
iex(1)> URI.parse("https://examplebucket.s3.amazonaws.com/test hello+#3.txt")
%URI{
  authority: "examplebucket.s3.amazonaws.com",
  fragment: "3.txt",
  host: "examplebucket.s3.amazonaws.com",
  path: "/test hello+",
  port: 443,
  query: nil,
  scheme: "https",
  userinfo: nil
}
```

Notice the path is mangled to `/test hello+` now missing the rest of the URL `#3.txt`. This is correct to the RFC spec which Elixir follows, but we need to URL-encode it first (but without `URI.parse` so it doesn't drop information), but not the entire URL (we don't want `http%3A//`)

This causes two issues:

1. The request path is wrong
2. The signature is generated on the wrong path; so even if the request path is corrected, the signature won't match. 

What I did to help prevent it

- Consolidate URI functions into `ExAws.Request.Url`. Some was happening in `ExAws.Auth.Utils`, and some in `ExAws.Request`. I figured these were all about the URL, so I moved it to that module. I kept the public functions, but passing them through to the new location. It seemed like the URL wrestling was happening in a couple places, so I moved it closer to the signature generation.
- Implemented a crude (help me out here!) way to get the path of the URL, without using `URI.parse` to keep the s3 bucket object path.
- The `+` character was being transformed to a space character too early. The character in filenames needs to be transformed to `%2B`. 
- This new code path is limited to the S3 service, so it shouldn't affect other services (they'll continue to use `URI.parse`).

# Open Questions:

1. Without changing the users' filenames, is there an expectation for a ExAws user (me) to URL-encode things before using ExAws?
2. I found that `service` was passed into these functions in different ways. I saw `"s3"` and `:s3` in the test suite. Are there other variations?
3. I did change some functions and move them, but don't want to break semver. Can someone verify if this breaks semver?
4. I didn't find any AWS docs, examples, or SDKs that had a test around these scenarios. If you find any, that would be really helpful to ensure this is signing correctly (test suite is passing though). [There is confirmation from AWS that they're weird :)](https://forums.aws.amazon.com/thread.jspa?threadID=55746). [There is documentation saying that S3 is treated special for URI encoding when generating the Authorization signature](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) (screenshot below).
5. The new tests that cover these scenarios are trusted with the fact that I have another app using this branch of ExAws and it worked. I'm using these functions: 

```elixir
ExAws.S3.Download.get_chunk(
  ExAws.S3.download_file("us-east-1", "my bad file+name#3.csv", tmp_file),
  %{start_byte: 0, end_byte: 1024},
  ExAws.Config.new(:s3)
)

# and

"us-east-1"
|> ExAws.S3.download_file("my bad file+name#3.csv", tmp_file)
|> ExAws.request()

# and 

ExAws.S3.presigned_url(
  ExAws.Config.new(:s3),
  :get,
  "us-east-1",
  "my bad file+name#3.csv"
)
```

image
 

### Lessons for other developers
Hey, you (and future me!) don't let users upload crazy stuff. Cleanse your user uploads' filenames into normal ascii url-encodeable ways (or error it) before upload to S3 so you avoid this problem. Do something like this (when you launch the feature, not after since this would break older uploaded files)

```elixir
@unallowed_chars ~r|[^a-zA-Z0-9.\-_~!$'()*+,;=:@]|
uploaded_filename = ~s|test hello$-_.+!*'()~"<>#%{}\|\^[]`.txt|
cleansed_filename = String.replace(uploaded_filename, @unallowed_chars, "_")
#=> "test_hello$-_.+!*'()~____________.txt"
```

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -93,8 +93,6 @@ defmodule ExAws.Auth do
           uri.path
         end
 
-      path = uri_encode(path)
-
       signature =
         signature(
           http_method,
@@ -120,14 +118,12 @@ defmodule ExAws.Auth do
 
   defp auth_header(http_method, url, headers, body, service, datetime, config) do
     uri = URI.parse(url)
-    path = uri_encode(uri.path)
-
     query =
       if uri.query,
         do: uri.query |> URI.decode_query() |> Enum.to_list() |> canonical_query_params,
         else: ""
 
-    signature = signature(http_method, path, query, headers, body, service, datetime, config)
+    signature = signature(http_method, url, query, headers, body, service, datetime, config)
 
     [
       "AWS4-HMAC-SHA256 Credential=",
@@ -142,7 +138,8 @@ defmodule ExAws.Auth do
     |> IO.iodata_to_binary()
   end
 
-  defp signature(http_method, path, query, headers, body, service, datetime, config) do
+  defp signature(http_method, url, query, headers, body, service, datetime, config) do
+    path = url |> get_path() |> uri_encode()
     request = build_canonical_request(http_method, path, query, headers, body)
     string_to_sign = string_to_sign(request, service, datetime, config)
     Signatures.generate_signature_v4(service, config, datetime, string_to_sign)
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -87,17 +87,18 @@ defmodule ExAws.Auth do
 
       uri = URI.parse(url)
 
+      path = url |> Url.get_path(service) |> Url.uri_encode
       path =
         if uri.query do
-          uri.path <> "?" <> uri.query
+          path <> "?" <> uri.query
         else
-          uri.path
+          path
         end
 
       signature =
         signature(
           http_method,
-          path,
+          url,
           query_to_sign,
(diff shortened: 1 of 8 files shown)


dbernheisel:
related #648

rogelio2k:
thanks @dbernheisel ! I hope this get merged soon, without it, ex_aws_ s3 fails horribly!

benwilson512:
Thanks!

rogelio2k:
Awesome! thanks guys! 

/cc @dbernheisel

PR #666, opened 2019-11-22
Delete extraneous file

nathanl:
I think .tool-versions-e was created mistakenly - see https://github.com/asdf-vm/asdf/pull/250


benwilson512:
Thanks!

PR #667, opened 2019-11-23
Update available SES endpoints.

daveli:
Update the available SES endpoints based on [https://docs.aws.amazon.com/ses/latest/DeveloperGuide/regions.html](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/regions.html).


benwilson512:
Thanks!

PR #668, opened 2019-11-26
Unsignable request headers

james-bowers:
Specifies a list of unsignable headers that are not used for the signature in the authorisation header. This is required as the `X-Amzn-Trace-Id` header can be sent, but should not be used in the Signature in the request's `Authorization` header. This is confirmed by the [js-sdk](https://github.com/aws/aws-sdk-js/blob/master/lib/signers/v4.js#L191-L199).

I am not too convinced by having the `unsignable_headers` list specify both uppercase and lowercase formats. I did originally have another iteration over the headers after the current `Enum.map` where the keys would always be downcased, perhaps this is better? Happy to discuss/change this!

I am aware that not everyone will be using the unsignable headers such as the `X-Amzn-Trace-Id` and therefore have tried to make as small impact on performance as I can.

This PR is related to https://github.com/ex-aws/ex_aws_lambda/pull/3

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -7,7 +7,8 @@ defmodule ExAws.Auth do
 
   @moduledoc false
 
-  @unsignable_headers ["X-Amzn-Trace-Id", "x-amzn-trace-id"]
+  @unsignable_headers ["x-amzn-trace-id"]
+  @unsignable_headers_multi_case ["x-amzn-trace-id", "X-Amzn-Trace-Id"]
 
   def validate_config(config) do
     with :ok <- get_key(config, :secret_access_key),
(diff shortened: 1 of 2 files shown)


lib/ex_aws/auth.ex
```diff
@@ -207,6 +209,7 @@ defmodule ExAws.Auth do
   defp signed_headers(headers) do
     headers
     |> Enum.map(fn {k, _} -> String.downcase(k) end)
+    |> Kernel.--(@unsignable_headers)
```
benwilson512: Because of the downcase in the previous line you shouldn't need multiple cases.

james-bowers: 👍 Hopefully I've addressed this now, sorry about the delay

MattLongCode: @benwilson512 Would you possibly be able to review? Thanks

PR #675, opened 2020-02-12
Updating available Textract endpoints.

collinmurray:
Updated the available Textract endpoints based on https://docs.aws.amazon.com/general/latest/gr/textract.html


benwilson512:
Thanks!

PR #679, opened 2020-02-25
Use common config for runtime region value

oskarkook:
The current implementation can end up with an issue where the `json_codec` value is not present and fetching region at runtime fails.


oskarkook:
nevermind, region can't be pulled from instance meta anyway with the current implementation

PR #685, opened 2020-03-18
Relax Jason dependency version

nifoc:
With the release of Jason 1.2, having a dependecy on `~> 1.1.0` becomes
a problem. Example:

```
Failed to use "jason" because
  [...]
  ex_aws (version 2.1.2) requires ~> 1.1.0
```

This can be resolved by simply depending on `~> 1.1`.


benwilson512:
Thanks!

PR #688, opened 2020-04-23
High concurrency on ets can cause dirty cache to leak empty/wrong auth

tjsousa:
In a high throughput application using the `awscli_auth_adapter` option we were seeing a small percent of errors that seem to correlate with the `AuthCache` refresh period:

`{:error, "Required key: :secret_access_key is nil in config!"}`

This can happen as the auth config loaded from the configuration file (that can be used for the adapter but is not the final credentials to be used for AWS requests - as it may not have the same permissions or can even be empty when using ENV vars) is temporarily stored in `:ets` that can be concurrently read.

The attached unit test was able to reproduce the issue and is currently passing with the proposed fix.

### DIFF
--- a/lib/ex_aws/config/auth_cache.ex
+++ b/lib/ex_aws/config/auth_cache.ex
@@ -25,8 +25,11 @@ defmodule ExAws.Config.AuthCache do
 
   def get(profile, expiration) do
     case :ets.lookup(__MODULE__, :awscli) do
-      [{:awscli, auth_config}] -> auth_config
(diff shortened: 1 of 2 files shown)


benwilson512:
Good catch, thank you!

dnlserrano:
hey @benwilson512 thanks so much for this library ❤️ when can we expect a release that includes this fix? thanks in advance! keep up the great work. 🙇

ylebedeva:
@benwilson512 we'd greatly appreciate if this fix made it into a released version. Encountered it on our team, and were very happy to find a fix already in place - shows how great the library and the community is. Thank you!

benwilson512:
Released.

PR #691, opened 2020-05-14
normalization configuration info

danj3:
I discovered this feature only by digging through the code. Added info on this option to the README.md


danj3:
A question has arisen regarding normalization: When it's off, S3.list_objects with no arguments doesn't show the root of the bucket contents, with our without a delimiter. In looking at the s3 code, I saw this:

>     request(:get, bucket, "/", [params: params, headers: opts[:headers]],

The "/" generates a double // in the REST url and that breaks things. I changed it to "":

>     request(:get, bucket, "", [params: params, headers: opts[:headers]],

 and that seems to work with normalization on or off for all the S3 request types I've tried so far. Does this change make sense or no?

bernardd:
Thanks Dan - If you'd be kind enough to remove the extra spaces that seem to have snuck in to the last line of the README on this change, I'll go ahead and merge it. 

I don't have a good answer to your question right now - I've just picked up this project and am still learning my way around it.

benwilson512:
@bernardd FYI github lets you edit files from within the UI for PRs now which can sometimes be useful for things like that.

bernardd:
@benwilson512 Ah brilliant, thanks. Didn't know that.

PR #694, opened 2020-06-05
Add eu-west-3 (Paris) to firehose

DanBunea:
We are trying to use firehose for eu-west-3 (Paris) but without being added we get an exception.


bernardd:
Thanks Dan!

PR #699, opened 2020-06-30
Fixed email endpoint regions

alg:
Added ca-central-1, sa-east-1 and replaced ap-southeast-1 with ap-southeast-2


bernardd:
Hi Aleksey,
It looks like ap-southeast-1 is (at least now) a valid email zone - if you're able to add that back in, I'll go ahead and merge this. Cheers.

PR #697, opened 2020-06-24
added chime support

marinakr:
Example:
```
   %ExAws.Operation.JSON{
          service: :chime,  
          http_method: :post,  
          path: "/meetings", 
          data: %{
             "ClientRequestToken" => "client_token_foo", 
             "ExternalMeetingId" => "test_id",
             "MediaRegion" => "us-east-2",
             "MeetingHostId" => "bar.localhost"
          }
}
|>  ExAws.request()
```
Response:
```
{:ok,
 %{
   "Meeting" => %{
     "ExternalMeetingId" => "test_id",
     "MediaPlacement" => %{
       "AudioFallbackUrl" => "wss://haxrp.m1.ue2.app.chime.aws:443/calls/11111111-9ae4-4f07-8ced-a7b0016b3339",
       "AudioHostUrl" => "c00000000e8df2fe0fd97e91dce65b04.k.m1.ue2.app.chime.aws:3478",
       "ScreenDataUrl" => "wss://bitpw.m1.ue2.app.chime.aws:443/v2/screen/11111111-9ae4-4f07-8ced-a7b0016b3339",
       "ScreenSharingUrl" => "wss://bitpw.m1.ue2.app.chime.aws:443/v2/screen/11111111-9ae4-4f07-8ced-a7b0016b3339",
       "ScreenViewingUrl" => "wss://bitpw.m1.ue2.app.chime.aws:443/ws/connect?passcode=null&viewer_uuid=null&X-BitHub-Call-Id=11111111-9ae4-4f07-8ced-a7b0016b3339",
       "SignalingUrl" => "wss://signal.m1.ue2.app.chime.aws/control/11111111-9ae4-4f07-8ced-a7b0016b3339",
       "TurnControlUrl" => "https://ccp.cp.ue1.app.chime.aws/v2/turn_sessions"
     },
     "MediaRegion" => "us-east-2",
     "MeetingId" => "11111111-9ae4-4f07-8ced-a7b0016b3339"
   }
 }}
```

PR #700, opened 2020-07-07
Add Chime service data

bernardd:
This adds details for the Chime conferencing service (https://aws.amazon.com/chime/).

I'm just in the very early stages of putting together an `ex_aws_chime` implementation. More to come soon hopefully.


bernardd:
Arg dammit, I should check existing PRs. #697 beat me to it and in what looks like a more comprehensive way :)

PR #704, opened 2020-07-21
Fix issue where query parameters for S3 are double encoded

kerryjj:
When using `ExAws.S3.list_objects_v2` or `ExAws.S3.list_objects` and a key that has a space or unicode chars in it, the query params passed to S3 are double encoded.

The following command fails because the URL signatures don't match. This is because the query prefix is encoded to replace " " with +. Then the sanitize method replacces + with %2B.

This PR changes the sanitize method so that it only sanitised the path and not the query params again and the output should be as follows. There are tests as part of the PR that show the new behaviour.

### DIFF
--- a/lib/ex_aws/request/url.ex
+++ b/lib/ex_aws/request/url.ex
@@ -63,10 +63,9 @@ defmodule ExAws.Request.Url do
     url
     |> URI.parse()
     |> Map.put(:fragment, nil)
-    |> Map.put(:path, "/" <> new_path)
-    |> Map.put(:query, query)
+    |> Map.put(:path, "/" <> String.replace(new_path, "+", "%2B"))
+    |> Map.put(:query, query_replace_spaces(query))
     |> URI.to_string()
-    |> String.replace("+", "%2B")
   end
 
   def sanitize(url, _), do: String.replace(url, "+", "%20")
@@ -105,6 +104,9 @@ defmodule ExAws.Request.Url do
   def get_path(url, _), do: URI.parse(url).path
 
   def uri_encode(url), do: URI.encode(url, &valid_path_char?/1)
+
+  defp query_replace_spaces(nil), do: nil
+  defp query_replace_spaces(query), do: String.replace(query, "+", "%20")
 
   # Space character
   defp valid_path_char?(?\ ), do: false
--- a/test/ex_aws/request/url_test.exs
+++ b/test/ex_aws/request/url_test.exs
@@ -18,58 +18,74 @@ defmodule ExAws.Request.UrlTest do
     {:ok, %{query: query, config: config}}
   end
 
-  test "it build urls for query operation", %{query: query, config: config} do
-    assert Url.build(query, config) == "https://example.com/path?foo=bar"
-  end
+  describe "build" do
+    test "it build urls for query operation", %{query: query, config: config} do
+      assert Url.build(query, config) == "https://example.com/path?foo=bar"
+    end
 
-  test "it allows setting custom port", %{query: query, config: config} do
-    config = config |> Map.put(:port, 4430)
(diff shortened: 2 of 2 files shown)


lib/ex_aws/request/url.ex
```diff
@@ -106,6 +105,9 @@ defmodule ExAws.Request.Url do
 
   def uri_encode(url), do: URI.encode(url, &valid_path_char?/1)
 
+  defp query_replace_spaces(nil), do: nil
+  defp query_replace_spaces(query), do: String.replace(query, "+", "%20")
```
benwilson512, line 109: It's `%2B` in one place and `%20` in another. This seems wrong?

kerryjj: Actually, that whole `String.replace(new_path, "+", "%2B")` isn't needed and has no effect. The `new_path` gets uri_encoded a little higher up.

In the S3 operation code path, the `url` that gets passed to sanize has already had the query params encoded, but the reset of the URL hasn't been. So if `new_path` in this case contains a +, then it is actually a + and not a space.

s3.ex:29 calls `ExAws.Request.Url.build(config)` which is what encodes the query and not the path. 

It feels a little bit inconsistent to encode one earlier and the other later. Maybe it would be better to shift the encoding to one place in the code. And perhaps it's best to encode both the path and the query in Url.sanitize() because ExAws.Operation.{JSON,Query,RestQuery} also call ExAws.Request.Url.build and therefore have their query encoded if they include any.

kerryjj: I've pushed a quick change taking out that replace. Let me know if you think it'd be a good idea for me to try to move the uri_encoding out of `ExAws.Request.Url.build(config)` and to encode both the path and query in ExAws.Request.Url.sanitize instead.

kerryjj:
Yeah good question about the tests. I have a much simpler implementation of the encoding now that encodes both in the build() method. It works for the tests currently in url_test.ex but I'm a little nervous about pushing it as it's not a comprehensive set of tests.

I'll see what I can find for more test coverage,  but yeah if anyone can point out a good suite to copy that would help.

kerryjj:
Sorry I haven't forgotten about trying to get better test coverage in here. I've not forgotten about it, just ended up with crunch time at work.

I'll hopefully get a chance to add some more over the weekend.

I have had a look through a few of the amazon provided libraries in python and ruby to see if there's an obvious set of tests to copy from there and haven't found anything comprehensive yet. I'll keep looking, but if anyone else finds a set of tests that are worth reviewing and perhaps copying and can point me in the right direction I'm happy to add them.

I am using this patch in production and it's fixed the original issue I had and the test cases in this PR at least cover that case. i.e. "When using ExAws.S3.list_objects_v2 or ExAws.S3.list_objects and a key that has a space or unicode chars in it, the query params passed to S3 are double encoded."

The encoding in the code here can definitely be simplified.

lib/ex_aws/request/url.ex
aseigo: `query_replace_spaces` appears to actually be replacing +'s rather than spaces .. and I think that could more simply use `URI.encode_www_form/1` instead of `URI.encode`.

On a related note, the query string parse is redundant with the `URI.parse/1` call which also separates out the query string.

It would be nice to see this fix get into the repo, but perhaps with a bit more cleanup?

eniac314:
I think I am having this issue as well on an old (2 years) project. `list_bucket` used to work but I seem to have problems with bucket names containing spaces now. Are there any plans to include the fix in the main repo?

bernardd:
@jotaviobiondo No, no updates. If you'd like to have a stab at adding some broader encoding tests and tidying up the PR, I'd be more than happy to accept an updated version.

PR #712, opened 2020-09-18
Elixir 1.11 compatibility

arjan:
Keyword.keys() cannot be used for non-keyword lists from [Elixir 1.11](https://github.com/elixir-lang/elixir/blob/v1.11/CHANGELOG.md#elixir-1); it was used here with a `[{String.t, String.t}]` instead. This fix concatenates signed headers using dedicated function.


benwilson512:
Released in v2.1.5, thanks!

PR #720, opened 2020-10-17
Fix Elixir 1.11 compiler warning

oohnoitz:
This addresses the the following compiler warning when using Elixir 1.11:

```
warning: Supervisor.Spec.worker/2 is deprecated. Use the new child specifications outlined in the Supervisor module instead
  lib/ex_aws.ex:93: ExAws.start/2
```

Also, fixed the `unused variables` warnings when running `mix test` as well. The min version was bumped to `1.5` since that was when the new Supervisor changes was added in.

### DIFF
--- a/lib/ex_aws.ex
+++ b/lib/ex_aws.ex
@@ -87,10 +87,8 @@ defmodule ExAws do
   @doc false
   @impl Application
   def start(_type, _args) do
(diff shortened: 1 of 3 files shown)


bernardd:
Thanks @oohnoitz

PR #721, opened 2020-10-18
Misc HTML doc generation changes

kianmeng:
List of changes:

- Use common source url
- Add new maintainer
- Remove duplicate page title
- Word wrapping
- Badges and more badges!
- Update example code styling


bernardd:
Nice one @kianmeng, thanks!

PR #726, opened 2020-11-03
Use S3 virtual hosted style urls

spatrik:
I've been running into an issue where I get a redirect error when trying to do a S3 PutObject:

```
# [warn] ExAws: Received redirect, did you specify the correct region?
# {:error, {:http_error, 301, "redirected"}}
```

After digging around on the internet for a while, I realized that if I update ExAws to use "virtual host-style" instead of "path style", everything works fine. I'm actually not really sure what the actual problem was, but I do think this is a change that is ok.

I'm not entirely sure about the implementation though. Making the Url module aware of S3 operations doesn't seem like the most awesome design. If someone has any idea how to solve this differently I'll be happy to try it out. 

Also, I haven't run the tests other than the ones for the Url module. If this seems like an ok way forward, I'll try to get the tests to run, but I wanted some feedback on the code f

More information can be found at [1] and according to the latest docs [2], all
S3 requests should use the virtual host style urls.

[1] https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
[2] https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax

### DIFF
--- a/lib/ex_aws/operation/s3.ex
+++ b/lib/ex_aws/operation/s3.ex
@@ -24,7 +24,6 @@ defmodule ExAws.Operation.S3 do
 
       url =
         operation
-        |> add_bucket_to_path
         |> add_resource_to_params
         |> ExAws.Request.Url.build(config)
 
@@ -48,11 +47,6 @@ defmodule ExAws.Operation.S3 do
 
     def stream!(%{stream_builder: fun}, config) do
       fun.(config)
-    end
-
-    def add_bucket_to_path(operation) do
-      path = "/#{operation.bucket}/#{operation.path}" |> String.trim_leading("//")
-      operation |> Map.put(:path, path)
     end
 
     def add_resource_to_params(operation) do
--- a/lib/ex_aws/request/url.ex
+++ b/lib/ex_aws/request/url.ex
@@ -7,6 +7,7 @@ defmodule ExAws.Request.Url do
   def build(operation, config) do
     config
     |> Map.take([:scheme, :host, :port])
+    |> set_host(operation)
     |> Map.put(:query, query(operation))
     |> Map.put(:path, operation.path)
     |> normalize_scheme
@@ -15,6 +16,16 @@ defmodule ExAws.Request.Url do
     |> (&struct(URI, &1)).()
     |> URI.to_string()
     |> String.trim_trailing("?")
+  end
+
+  defp set_host(config, operation) do
+    case operation do
+      %ExAws.Operation.S3{} ->
+        %{config | host: "#{operation.bucket}.s3.amazonaws.com"}
+
+      _ ->
+        %{config | host: config.host}
+    end
   end
 
   defp query(operation) do


bernardd:
On the face of it, it looks alright to me - if I recall correctly the path style was the "old" way of doing things, and at this point everything should work properly with the host style. I'd be grateful if you're able to get as many tests going and passing as possible, since the S3 module is pretty core to everything and this is a change right at the heart of the way it interacts with AWS.

lib/ex_aws/request/url.ex
```diff
@@ -17,6 +18,16 @@ defmodule ExAws.Request.Url do
...
 
+  defp set_host(config, operation) do
+    case operation do
+      %ExAws.Operation.S3{} ->
+        %{config | host: "#{operation.bucket}.s3.amazonaws.com"}
+
+      _ ->
+        %{config | host: config.host}
+    end
+  end
```
bernardd, line 29: For a function that's just one `case` operation, I think it's preferable to divide it into two pattern matches like this:

```elixir
defp set_host(config, %ExAws.Operation.S3{bucket: bucket}), do: %{config | host: "#{bucket}.s3.amazonaws.com"}

defp set_host(config, _operation), do: config
```

bernardd:
Actually looking at the change further, I think the s3 host should be set in `config` in the same place that the path was previously - it's an S3 specific thing so it doesn't make sense to set it in the common ex_aws code.

spatrik:
It turns out that my issue was not actually related to this and instead me being unable to spell my config var name correctly, making the bucket name empty. 

Are you still interested in this change? I believe it would solve ex-aws/ex_aws_s3#95? If so I'll try to take a look at it this weekend and implement what you suggest.

bernardd:
Yep, more than happy for you to go ahead with the tweaks and to get this merged. Cheers.

lib/ex_aws/operation/s3.ex
```diff
@@ -50,11 +49,6 @@ defmodule ExAws.Operation.S3 do
       fun.(config)
     end
 
-    def add_bucket_to_path(operation) do
```
carrascoacd, line 53: I think we should keep this function and make it optional by configuration, so this is not a breaking change.

```elixir
    def maybe_add_bucket_to_path(operation, true = _virtual_host), do: operation

    def maybe_add_bucket_to_path(operation, false = _virtual_host) do
      path = "/#{operation.bucket}/#{operation.path}" |> String.trim_leading("//")
      operation |> Map.put(:path, path)
    end
```

fetching the `virtual_host` from configuration and setting it by default to `true`

lib/ex_aws/request/url.ex
```diff
@@ -17,6 +18,16 @@ defmodule ExAws.Request.Url do
     |> String.trim_trailing("?")
   end
 
+  defp set_host(config, operation) do
+    case operation do
+      %ExAws.Operation.S3{} ->
```
carrascoacd, line 23: IMHO this is coupling this module with the S3 operation, why not doing this stuff in the `S3.ex` and passing it by a parameter in the `config`?

carrascoacd (commented):
I think this PR will be very useful, I left some comments, please let me know if you need help. I'll be glad to lend a hand

bernardd:
> I think this PR will be very useful, I left some comments, please let me know if you need help. I'll be glad to lend a hand

I'm sure it will too - what I'd really like to do is get a bit more automated testing in place first (which is something I'm planning to work on over the holidays) so that I can be a bit more confident about merging in stuff like this because a) I'm not familiar enough with the code yet to just go merging all the changes I think look good, and b) automated testing is a good thing to have anyway :)

I've got a reasonable idea how I want to go about it, it's just a matter of doing it :)

lib/ex_aws/operation/s3.ex
aseigo: There is no reason to keep the non-virtual-host style. Instead, `add_bucket_to_path/1` should be kept *as is* except instead of adding the bucket to the *path* it should be adding the bucket to the *host*. 

That would still be a breaking change, however, as existing configurations that currently work would no longer work (including my own apps using this library, so ... :/).

lib/ex_aws/operation/s3.ex
```diff
@@ -24,7 +24,6 @@ defmodule ExAws.Operation.S3 do
 
       url =
         operation
-        |> add_bucket_to_path
```
aseigo, line 27: Doesn't this break non-virtual-host configus, such that the bucket will never be added to the path now? IOW, this looks like it would work *only* with virtual host setups?

lib/ex_aws/request/url.ex
```diff
@@ -17,6 +18,16 @@ defmodule ExAws.Request.Url do
...
+        %{config | host: "#{operation.bucket}.s3.amazonaws.com"}
```
aseigo, line 24: Assuming the host suffix is `s3.amazonaws.com` is invalid. In an app I have the host suffix is `s3.nl-ams.scw.cloud`.

This should not be done here anyways, imho, but back in `operations/s3.ex` where it used to be.

aseigo (commented):
I have been hit by this same thing, where I'm using a virtual host and including the bucket ended up giving me an extra folder named the same as the bucket in the top level... obviously not what is wanted. So this change looks desirable, but the implementation looks to need some work.

I agree that it should be done in a backwards-compat way (no breaking changes to public API),  and that it should also not break non-virtual-host configs. So I recommend that instead a new configuration variable is added to the s3 configuration: `virtual_host_base: String.t()`. 

`add_bucket_to_path/1` could be expanded with an optional 2nd map parameter that the configuration is passed into, and if there is a *virtual_host_base* defined then add the bucket to that and set it as the host for the operation, otherwise the existing code path that adds it to the path is retained.

This should, at least imho, not touch `request/url.ex` at all.

aseigo:
I have submitted an alternate implementation in #735 .. I invite those interested to look! :)

bernardd:
I think #735 is a preferable implemenation. Thanks for taking the time on this one @spatrik, but I'm going to close it in favour of the other approach.

spatrik:
I agree that 735 is preferable. Sorry for disappearing from this, work and holidays got in the way.

bernardd:
All good, I know the feeling :)

PR #728, opened 2020-11-30
Support for af-south-1

Ivor:
This adds support for the Cape Town region. 

The region has not been added to some, but not all the services that are available in Cape Town. 

The endpoints.exs file seems to have a striking resemblance to this file, 
https://github.com/boto/botocore/blob/master/botocore/data/endpoints.json, as mentioned in this post: 
https://www.transposit.com/blog/2019.10.21-aws-regions-api/

Can I update the contents of priv/endpoints.exs with the contents of the endpoints.json file from Boto?

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -77,7 +77,7 @@ defmodule ExAws.Config.Defaults do
   end
 
   @partitions [
-    {~r/^(us|eu|ap|sa|ca)\-\w+\-\d+$/, "aws"},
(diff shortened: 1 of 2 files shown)


bernardd:
Thanks Ivor!

> Can I update the contents of priv/endpoints.exs with the contents of the endpoints.json file from Boto?

I think that's almost certainly the case. I can't swear to it, since I didn't write this initially, but given the resemblance, and the fact that boto is an official AWS  I'd say by all means go ahead and update based on it. Do you want to make those changes in this PR?

bernardd:
Oh, huh, I didn't notice the changes had already been submitted here. Let's get this merged then.

PR #731, opened 2020-12-18
Add us-east-2 as allowable SES email endpoint

ascandella:
SES is supported in `us-east-2`. I've tested this fork in AWS and received an email from my domain.

image


bernardd:
Thanks @ascandella!

PR #732, opened 2020-12-23
Use :crypto.mac/4 instead of :crypto.hmac/3

IvanIvanoff:
:crypto.hmac/{3,4} is deprecated and will be removed in Erlang/OTP 24.
:crypto.mac/4 is used instead. It is available since Erlang/OTP 22.1, so some availability checks must be performed.

### DIFF
--- a/lib/ex_aws/auth/utils.ex
+++ b/lib/ex_aws/auth/utils.ex
@@ -9,8 +9,18 @@ defmodule ExAws.Auth.Utils do
     |> bytes_to_hex
   end
 
-  def hmac_sha256(key, data) do
-    :crypto.hmac(:sha256, key, data)
+  # :crypto.mac/4 is introduced in Erlang/OTP 22.1 and :crypto.hmac/3 is removed
+  # in Erlang/OTP 24. The check is needed for backwards compatibility.
+  # The Code.ensure_loaded/1 call is executed so function_expored?/3 can be used
+  # to determine which function to use.
+  Code.ensure_loaded?(:crypto) || IO.warn(":crypto module failed to load")
+
+  case function_exported?(:crypto, :mac, 4) do
+    true ->
+      def hmac_sha256(key, data), do: :crypto.mac(:hmac, :sha256, key, data)
+
+    false ->
+      def hmac_sha256(key, data), do: :crypto.hmac(:sha256, key, data)
   end
 
   def bytes_to_hex(bytes) do


bernardd:
Thanks for the patch, Ivan! Could I suggest using `function_exported?(:crypto, :hmac, 3)` to decide whether to use the new version instead? It seems like a clearer, more direct way to figure it out than parsing the OTP version.

aboroska:
Make sure you call `Code.ensure_loaded(:crypto)` before calling `function_exported?/3` as the latter returns `false` when the module is not loaded, which will be the case very likely during compilation.

bernardd:
That shouldn't matter though, should it? The `function_exported?/3` call is made at run-time, not compile time.

aboroska:
@bernardd It is called at compile time.

Compile the following example:
```
defmodule Ver do
  IO.puts("load...")
  Code.ensure_loaded(:crypto)
  if function_exported?(:crypto, :mac, 4) do
    IO.puts("???? mac")
    def m(), do: :mac
  else
    IO.puts("???? hmac")
    def m(), do: :hmac
  end
end
```

```
$ elixirc ver.ex

load...
???? mac
```

```
$ iex
Erlang/OTP 22 [erts-10.7.1]
Interactive Elixir (1.11.1) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Ver.m()
:mac
```

Comment out `Code.ensure_loaded(:crypto)` recompile, and the result will be different.
```
iex(1)> Ver.m()
:hmac
```

bernardd:
Well yes, if you put it outside a function, sure. But that's not what we're doing here. Try this module:

```
defmodule Test do
  def test() do
    if function_exported?(:crypto, :mac, 4) do
      IO.puts("HAS mac")
      true
    else
      IO.puts("DOES NOT HAVE mac")
      false
    end
  end
end
```

Compile it once with `elixirc test.ex` then run. You get:

```
$ iex
Erlang/OTP 23 [erts-11.0.4] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [hipe]

Interactive Elixir (1.10.4) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Test.test()
DOES NOT HAVE mac
false
iex(2)> Code.ensure_loaded(:crypto)
{:module, :crypto}
iex(3)> Test.test()
HAS mac
true
```

If the conditional was evaluated at compile time you couldn't possibly get two different results like that.

aboroska:
@bernardd note that both @IvanIvanoff 's PR and my example does the comparison at **compile** time.

Also, note that you will get different results at **compile** time depending on whether the crypto module is loaded or not. I highlighted that in my example. By default the crypto module is not loaded at compile time, therefore we need to explicitly load it, which is a small inconvenience.

A solution could be to combine all 3 ideas presented here. @IvanIvanoff would you mind altering your PR to incorprorate both @bernardd 's and my suggestions? 

Similar to this:

    Code.ensure_loaded(:crypto)
    case function_exported?(:crypto, :mac, 4) do
      true ->
        def hmac_sha256(key, data), do: :crypto.mac(:hmac, :sha256, key, data)

      false ->
        def hmac_sha256(key, data), do: :crypto.hmac(:sha256, key, data)
    end

bernardd:
Ah dammit, you're right, sorry, I missed that the PR had the check at the top level of the module rather than in a function. My bad. And now that I look again, I see why we *can't* just put it inside a function like I was thinking (because it will fail to compile with a missing function error). I like @aboroska's suggestion, if you're happy to do it that way @IvanIvanoff.

IvanIvanoff:
I've force pushed the requested changes along with some description.

Out there `Code.ensure_loaded?` is used mostly in constructs like this one (with checking the result):
```elixir
if Code.ensure_loaded?(module) do
   defmodule SomeModule do
   end
end
```

I am just wondering if it would be better to raise a compile time warning if the load fails like this:
```elixir
Code.ensure_loaded?(:crypto) || IO.warn(":crypto module failed to load")
```

bernardd:
`:crypto` is one of the standard modules, so if it's missing then something is pretty badly awry (or someone's cooked up their own custom distro). Having said that, there's no harm in including a warning and you never know - it could help someone out.

bernardd:
Fantastic, thanks very much @IvanIvanoff and @aboroska

Issue #733, opened 2021-01-05
Signing URLs with query params appears to be broken in PR#660


aseigo:
### Environment

* Elixir & Erlang versions: Elixir 1.11.2 on Erlang/OTP 23 [erts-11.1.5]
* ExAws 2.1.6
* hackney 1.17.0

### Current behavior

I noticed this when running the `ex_aws_s3` test suite which now fails at https://github.com/ex-aws/ex_aws_s3/blob/2482ca3827f67ff50f238e07a567b7b769833c7f/test/lib/s3_test.exs#L371

Prior to PR#660 the path was being `uri_encode`d, which meant that query string components were being added to the path. Now that the path no longer is encoded, it is possible to get URLs with *two* query strings due to:

```
path <> "?" <> uri.query
```

followed by

```
"#{uri.scheme}://#{uri.authority}#{path}?#{query_for_url}&X-Amz-Signature=#{signature}"
```

I'm not entirely sure what the correct solution ought to be here: should a query string passed in with the URL be added to the path as a URI-encoded suffix, as the test in `ex_aws_3` expects? Or should it be added to the query string along-side the items such as X-Amz-Signature?

I'm happy to do a PR for this if someone knows *which* is the correct approach for AWS.

bernardd:
My problem at the moment is that I only just took over this project and one of the main things Ben told me as he disappeared into the sunset was "by the way, the URL signing is all over the place and every time someone provides a PR to fix it, it breaks it for someone else". Looks like he wasn't joking.
My plan at the moment is to actually get some decent tests up and running to exercise all this stuff so that whatever changes get made in the future can be done so with some confidence that we're not breaking behaviour others are relying on. The framework for the test stuff is about 90% done, I just need to do the remaining 90%.
Like you, right now I don't have my head around this stuff quite enough to know what's the "right" solution here, so I'm more than happy to listen to anyone else who has opinions.

aseigo:
Re: tests -> that's a great idea. There seem to be more tests that exercise this code in the ex_aws_s3 repo (which is how I found this issue: by running its test suite!), so maybe that could be a starting point worth looking at.

I located what appears to be the official documentation on URL signing: https://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html

Not hard to see how tricky this can be to implement. It is well documented, but using natural language and with loooots of steps and details. Fair enough .... 

[...]

PR #734, opened 2021-01-05
Attempt to fix presigned urls with query strings

aseigo:
This is a possible fix for #733 in which the parameters passed as part of the URL as included with the rest of the URL params, avoiding broken urls with things like "?foo=bar?X-Amz-Signature.

I am not 100% sure if this is the correct fix, but it looks so according to the documentation. Missing test suites for this code do not help, and should really be added in order to determine correctness. The best I can say is that the test suite that exists works, that  a correct test suite in `ex_aws_s3` now passes, and a local app I have written which uses `ex_aws_s3` still works as well though it is not hitting all of the affect code paths. 

Therefore, I consider this a moderately risky merge, though it does address what looks to be an obvious error?

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -75,28 +75,27 @@ defmodule ExAws.Auth do
       service = service_name(service)
       signed_headers = presigned_url_headers(url, headers)
 
-      org_query_params = query_params |> Enum.map(fn {k, v} -> {to_string(k), v} end)
+      uri = URI.parse(url)
+      uri_query = query_from_parsed_uri(uri)
+
+      org_query_params =
+        Enum.reduce(query_params, uri_query, fn {k, v}, acc -> [{to_string(k), v} | acc] end)
 
       amz_query_params =
         build_amz_query_params(service, datetime, config, expires, signed_headers)
 
-      [org_query, amz_query] =
-        [org_query_params, amz_query_params] |> Enum.map(&canonical_query_params/1)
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks for this - for now I think the fix for the test supplied in https://github.com/ex-aws/ex_aws_s3/pull/100 (supplying the params in a separate structure) will suffice as a workaround until I can get some tests that actually hit AWS itself up and running. At that point I'll definitely come back and check this and assuming it's all good we'll get it merged.

aseigo:
Personally, I'd leave the test failing until the code is fixed.  The point of tests is to exercise and expose problems in the code, which the unit test in question is doing. Changing the test so it no longer traverses buggy code paths sort of defeats the purpose of having tests, no? If anything, there should be both tests: pass the queries in as a parameter to the function, and in the path directly.

Either that, or this patch could become quite a bit simpler and simply reject any paths that have query params on them, though that would be a (potentially significant) breaking change and so is probably a non-starter.

bernardd:
Yeah, on reflection you're absolutely right. I kind of wonder if it was ever intended to work the way the test implied though. The fact that there's a separate 'params' field kind of suggests it maybe wasn't. Having said that, I'm willing to be there's a bunch of people out there using it exactly that way, so we doubtless do want to fix it :) I'm gonna merge this and reinstate the test into ex_aws_s3 and then everyone is happy :)

aseigo:
:heart:

PR #735, opened 2021-01-06
Support virtual host style buckets

aseigo:
This is a proposed alternative approach to #726 

This is done in a backwards compatible manner to preserve existing configurations. In the next major release, consider switching the default to virtual host style as that is the current recommendation from S3 API providers.

Test coverage and typespecs for the functions were added, and I have also tested in a real-world application that uses virtual host style hostnames to differentiate buckets.

### DIFF
--- a/lib/ex_aws/operation/s3.ex
+++ b/lib/ex_aws/operation/s3.ex
@@ -22,10 +22,11 @@ defmodule ExAws.Operation.S3 do
       headers = operation.headers
       http_method = operation.http_method
 
+      {operation, config} = add_bucket_to_path(operation, config)
+
       url =
         operation
-        |> add_bucket_to_path
-        |> add_resource_to_params
+        |> add_resource_to_params()
         |> ExAws.Request.Url.build(config)
 
       hashed_payload = ExAws.Auth.Utils.hash_sha256(body)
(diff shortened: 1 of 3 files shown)


lib/ex_aws/operation/s3.ex
```diff
@@ -40,21 +41,31 @@ defmodule ExAws.Operation.S3 do
...
+
     defp put_content_length_header(headers, "", :get), do: headers
 
     defp put_content_length_header(headers, body, _) do
       Map.put(headers, "content-length", byte_size(body) |> Integer.to_string())
     end
 
-    def stream!(%{stream_builder: fun}, config) do
-      fun.(config)
+    @spec add_bucket_to_path(operation :: ExAws.Operation.S3.t(), config :: map) ::
```
carrascoacd, line 54: What about defining the `operation` type in line 18, so you can reuse it and make the function single-line?

lib/ex_aws/operation/s3.ex
```diff
@@ -40,21 +41,31 @@ defmodule ExAws.Operation.S3 do
       |> operation.parser.()
     end
 
+    def stream!(%{stream_builder: fun}, config) do
```
carrascoacd, line 44: You can do it single line.

```elixir
def stream!(%{stream_builder: fun}, config), do: fun.(config)
```

carrascoacd (commented):
Thanks for taking care, I'd prefer this approach since the logic resides in the `S3` module.

I left a couple of minor comments.

lib/ex_aws/operation/s3.ex
aseigo: indeed; it was already multi-line, so i left it as-is. may as well tighten it up while i'm in there, though, i supose.

lib/ex_aws/operation/s3.ex
aseigo: Not sure what you mean by this?

Adding the bucket ends up modifying either the host or the path, so it needs to take both and return both. Otherwise, the logic of which needs to be modified ends up outside this function, which is even worse.

carrascoacd: Sorry let me explain it better. What I mean is:

```elixir
#...

@type t :: %__MODULE__{}
@type operation :: ExAws.Operation.S3.t()
@type config :: map()
#...

@spec add_bucket_to_path(operation, config) :: {operation, config}
def add_bucket_to_path(operation, config \\ %{})

#...
```

aseigo: aaah! yes, i see what you mean. personally, i prefer to see it spelled out fully so i don't have to go chasing definitions (even with the LS it's not great). even though it is more verbose, at a glance one can see exactly what it is doing.

bernardd:
Apart from the minor tweak to the test file, this looks good - I much prefer not breaking compatability and using a config param that's already used in ex-aws-s3.  Subject to that tweak, I'm happy to go ahead and merge this. Thanks!

aseigo:
> Apart from the minor tweak to the test file, this looks good

Which tweak?

bernardd:
> Which tweak?
The one in the review I forgot to press 'submit' on :)

aseigo:
> The one in the review I forgot to press 'submit' on :)

lol :) no worries, it's in now...

bernardd:
Perfect, thanks.

PR #736, opened 2021-01-06
Provide fake secrets for testing purposes

aseigo:
After the Nth failure to remember to setup the env vars expected for the API keys while testing, I added fake credentials to the test configuration. One less thing to remember, get wrong, and/or risk accessing actual data with real secrets floating in the ENV, this makes testing a bit easier.


bernardd:
Excellent.

PR #737, opened 2021-01-06
Refactor auth cache refreshing

aseigo:
This PR attempts to address a few different issues highlighted as a result of PR #625 including:

* enforce the singleton status of the AuthCache module
* always check returned auth and re-fetch if non-existent or expired
* only perform a scheduled refetch if the cached auth is close to expiry
* replace the hand-rolled time offset code with NaiveDateTime

It is separated into 3 different commits so each may be reviewed and easily only take parts. If this approach is acceptable and works as desired, a similar change should be made for the `:refresh_awscli_config` paths as well. 

I have very limited ability to test this thoroughly, so testing by those following PR #625 would be great welcomed.

### DIFF
--- a/lib/ex_aws/config/auth_cache.ex
+++ b/lib/ex_aws/config/auth_cache.ex
@@ -13,7 +13,7 @@ defmodule ExAws.Config.AuthCache do
   end
 
   def start_link(opts \\ []) do
-    GenServer.start_link(__MODULE__, :ok, opts)
+    GenServer.start_link(__MODULE__, :ok, Keyword.put(opts, :name, __MODULE__))
   end
 
   def get(config) do
--- a/lib/ex_aws/config/auth_cache.ex
+++ b/lib/ex_aws/config/auth_cache.ex
@@ -5,6 +5,9 @@ defmodule ExAws.Config.AuthCache do
 
   # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
 
+  @refresh_lead_time 300_000
+  @instance_auth_key :aws_instance_auth
+
   defmodule AuthConfigAdapter do
     @moduledoc false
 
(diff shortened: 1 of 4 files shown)


bernardd:
Hi @aseigo - I'm not ignoring this, I just haven't had the time yet to sit down and go through it (which is time that's going to have to include learning about the system that's there at the moment :)). Thanks for PR and the detailed description - I promise I'll get to it soon.

aseigo:
No worries, I understand how it can be :) If you need anything else from me, just holler!

lib/ex_aws/config/auth_cache.ex
```diff
@@ -79,19 +77,64 @@ defmodule ExAws.Config.AuthCache do
...
+      max(expires_in_ms, expires_in_ms - refresh_lead_time())
+    rescue
+      _e -> 0
+    end
   end
+
+  defp next_refresh_in(_), do: 0
+
+  defp refresh_lead_time(), do: 300_000
+  defp instance_auth_key(), do: :aws_instance_auth
```
bernardd, line 139: Just a small stylistic thing - I think these would probably be better as module attributes at the top of the file, ie:
```elixir
@refresh_lead_time 300_000
```
rather than private functions.

bernardd:
Right, I've got my head around it, and I think your explanation over in #625 is probably spot-on, and this looks like an excellent fix for it. Happy to merge pending the little change request. Thanks heaps for looking into this.

lib/ex_aws/config/auth_cache.ex
aseigo: Done :) Was easier to develop this with them as functions (better visibility on where/when it was being used), and I thought at one point it might make sense to be configurable, but as it turned out .. nah  :) I agree, better as it is now to be attrs. Cheers.

bernardd:
Perfect, thanks :)

PR #739, opened 2021-01-22
Support v2 of EC2 Instances Metadata Service (IMDSv2)

justincjohnson:
I'm using AWS's Security Hub service, and it recently created a finding for my environment, specifically `[EC2.8] EC2 instances should use IMDSv2`. The remediation details can be found at https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-8-remediation, but in short, the new version of the service requires an extra API call to generate a token, and then subsequent API calls need to pass that token in a header.

I'm opening this draft PR to raise awareness of the need to make these changes and to see if anyone has initial feedback for me.  As you'll see from my changes, I'm still working through the details of how to make this configurable and how to test.

Thanks in advance for any feedback.

### DIFF
--- a/lib/ex_aws/instance_meta.ex
+++ b/lib/ex_aws/instance_meta.ex
@@ -12,7 +12,10 @@ defmodule ExAws.InstanceMeta do
   @task_role_root "http://169.254.170.2"
 
   def request(config, url) do
-    case config.http_client.request(:get, url, "", [], http_opts()) do
+    # If we're using IMDSv2, we will need to pass in session token headers.
+    headers = get_request_headers(config)
+
+    case config.http_client.request(:get, url, "", headers, http_opts()) do
       {:ok, %{status_code: 200, body: body}} ->
         body
 
@@ -41,6 +44,11 @@ defmodule ExAws.InstanceMeta do
         ```
         """
     end
+  end
+
+  def get_request_headers(config) do
+    # TODO: Make this conditional on something in config
+    ExAws.InstanceMetaTokenProvider.get_headers(config)
   end
 
   def instance_role(config) do
--- a/lib/ex_aws/instance_meta_token_provider.ex
+++ b/lib/ex_aws/instance_meta_token_provider.ex
@@ -1,0 +1,103 @@
+defmodule ExAws.InstanceMetaTokenProvider do
+  @moduledoc """
+  For use with IMDSv2, this module retrieves the metadata session token and refreshes it before expiration.
+  """
+
+  # 6 hours
+  @metadata_token_ttl_seconds 6 * 60 * 60
+  @genserver_call_timeout_seconds 30
+  @metadata_token_api_url "http://169.254.169.254/latest/api/token"
+  # The header we pass to control the token's time to live
+  @metadata_token_ttl_header_name "x-aws-ec2-metadata-token-ttl-seconds"
+  # The header we use to pass the token along to all other metadata calls
+  @metadata_token_header_name "x-aws-ec2-metadata-token"
+
+  use GenServer
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, :ok, opts)
+  end
+
+  def get(config) do
+    case :ets.lookup(__MODULE__, :aws_metadata_token) do
+      [{:aws_metadata_token, token}] ->
+        token
+
+      [] ->
+        GenServer.call(
+          __MODULE__,
+          {:refresh_token, config},
+          @genserver_call_timeout_seconds * 1_000
+        )
+    end
+  end
+
+  def get_headers(config) do
+    [{@metadata_token_header_name, get(config)}]
+  end
+
+  ## Callbacks
+
+  def init(:ok) do
+    ets = :ets.new(__MODULE__, [:named_table, read_concurrency: true])
+    {:ok, ets}
+  end
+
+  def handle_call({:refresh_token, config}, _from, ets) do
+    token = refresh_token(config, ets)
+    {:reply, token, ets}
+  end
+
+  def handle_info({:refresh_token, config}, ets) do
+    refresh_token(config, ets)
+    {:noreply, ets}
+  end
+
+  def refresh_token(config, ets) do
+    token = request_token(config)
+    :ets.insert(ets, {:aws_metadata_token, token})
+    Process.send_after(self(), {:refresh_token, config}, refresh_in(config))
+    token
+  end
+
+  # TODO: Allow overriding defaults in config
+  def refresh_in(_config) do
+    # Check five minutes prior to expiration, or now, which ever is later.
+    refresh_in = @metadata_token_ttl_seconds - 5 * 60
+    max(0, refresh_in * 1_000)
+  end
+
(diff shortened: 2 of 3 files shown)


lib/ex_aws/instance_meta.ex
```diff
@@ -12,9 +12,13 @@ defmodule ExAws.InstanceMeta do
   @task_role_root "http://169.254.170.2"
 
   def request(config, url) do
-    # Certain solutions like kube2iam will redirect instance meta requests,
-    # we need to follow those redirects.
-    case config.http_client.request(:get, url, "", [], follow_redirect: true) do
+    # Certain solutions like kube2iam will redirect instance meta requests, we need to follow those redirects.
```
justincjohnson, line 15: Code reorg'd slightly to make it easier for me to comment on the additional options.

lib/ex_aws/instance_meta.ex
```diff
@@ -45,6 +49,11 @@ defmodule ExAws.InstanceMeta do
     end
   end
 
+  def get_request_headers(config) do
+    # TODO: Make this conditional on something in config
```
justincjohnson, line 53: I'm working on this.  If the host isn't configured to enable IMDSv2 (and thus the metadata token API), this will fail.  I'm assuming we want to add a config option in ExAws to control whether we want to attempt to get a token.

lib/ex_aws/instance_meta_token_provider.ex
```diff
@@ -0,0 +1,103 @@
+defmodule ExAws.InstanceMetaTokenProvider do
+  @moduledoc """
+  For use with IMDSv2, this module retrieves the metadata session token and refreshes it before expiration.
+  """
+
+  # 6 hours
+  @metadata_token_ttl_seconds 6 * 60 * 60
```
justincjohnson, line 7: I need to make these configurable, for testing if nothing else.

lib/ex_aws/instance_meta_token_provider.ex
```diff
@@ -0,0 +1,103 @@
...
+  def refresh_token(config, ets) do
+    token = request_token(config)
+    :ets.insert(ets, {:aws_metadata_token, token})
+    Process.send_after(self(), {:refresh_token, config}, refresh_in(config))
+    token
+  end
+
+  # TODO: Allow overriding defaults in config
+  def refresh_in(_config) do
+    # Check five minutes prior to expiration, or now, which ever is later.
```
justincjohnson, line 65: Same early checking timing used in `AuthCache` and also in https://github.com/aws-beam/aws_credentials.

lib/ex_aws/instance_meta_token_provider.ex
```diff
@@ -0,0 +1,103 @@
...
+    refresh_in = @metadata_token_ttl_seconds - 5 * 60
+    max(0, refresh_in * 1_000)
```
justincjohnson, line 67: It feels like this should have more robust checking, if we allow the timeouts to be configurable.

lib/ex_aws/instance_meta_token_provider.ex
```diff
@@ -0,0 +1,103 @@
...
+        You tried to access the AWS EC2 Instance Metadata token API, but it could not be reached.
+
+        Please check AWS EC2 Instance Metadata Service configuration to make sure the service is enabled.
+        """
+    end
+  end
+
+  # TODO: Allow overriding defaults in config
+  defp token_ttl_seconds_headers(_config) do
+    [{@metadata_token_ttl_header_name, @metadata_token_ttl_seconds}]
```
justincjohnson, line 101: This is the header format used by Hackney, the default client.  I haven't checked what other clients use for header format.  Given that the http client is configurable, I suppose this opens us up to some dependency on the header format for the configured http client.  Perhaps we should have some callback in the http client adapter behaviour to shape headers into the expected shape for the client?  Maybe I'm overthinking this, but thought I'd point out my thought process here in case someone has feedback.

lib/ex_aws/instance_meta.ex
bernardd: That seems reasonable.

lib/ex_aws/instance_meta_token_provider.ex
bernardd: Possibly, but I'm not sure what kind of thing we'd want to check for, nor what to do if we found it.

lib/ex_aws/instance_meta_token_provider.ex
bernardd: I think it's reasonable to use the same header approach as everywhere else in the code (which is what you have here). If we want to support a different client that has some alternative arrangement, the transform can be handled by a shim layer for that client, rather than trying to come up with some "always works for every client ever" system at the point where we're actually setting the headers.

bernardd:
Thanks for looking at this - I can't say I'm actually across the system in question, but I'm more than happy to consider anything that improves functionality.

GavinJoyce:
@justincjohnson thanks for opening the PR. I too would like to lock down my AWS instances to use IMDSv2. 

Do you think there is much to do to get this PR into a place where it could be merged? If there are concrete outstanding tasks, or if you need help testing this in production, please let me know. I'd be happy to help get this over the line

justincjohnson:
@GavinJoyce I just noticed your comment.  I introduced the required changes into https://github.com/aws-beam/aws_credentials and was switching over to https://github.com/aws-beam/aws-elixir soon, which uses aws_credentials.

Perhaps I should just wrap up this PR though so it's available in both.  I'll try to carve out some time.

GavinJoyce:
Thanks @justincjohnson, I wasn't aware of https://github.com/aws-beam/aws-elixir. I'll look into switching too

avivek:
Hi, 
Is there a tentative time that we can expect this to be merged??
Thanks.

avivek:
Hi @bernardd ,
Can we please release this if all good.

bernardd:
Hi @avivek - currently this PR is causing test failures, and the last word from the author is that it is still a work in progress. I'd welcome anyone who wants to finish it up and get it across the line.

avivek:
@justincjohnson Do u think you will be able to take this to closure?
Would be highly helpful. Thanks

justincjohnson:
Hi @avivek.  I can't make any promises but I may try to carve out the time to complete it.  If you feel like taking over my changes though, that is fine by me.  I don't need credit for the work.  :)

avivek:
Thanks for the quick reply @justincjohnson 
This is so outside what am able to do, am gonna have to wait.

brysonjl:
Hey @justincjohnson, is there any progress on getting this completed by chance?

justincjohnson:
Hey @brysonjl.  FWIW, the code on this branch has been working for me.  I haven't taken the time to clean up the remaining items in the PR though, which I believe are mainly making it more configurable and fixing the mocks for tests.  Feel free to create your own branch off of mine for your use or to wrap this up.

justincjohnson:
I'm not using ex_aws any more, so I closed my PR.  Thanks.

PR #744, opened 2021-02-10
Enable backblaze

keatz55:
ex_aws has AWS-specific region validation  in `ExAws.Config.Defaults`. E.g., `~r/^(us|eu|ap|sa|ca)\-\w+\-\d+$/` matches and validates against `us-west-2`. Unfortunately Backblaze has very similar region structure, `us-west-000`, which causes ex_aws to error out when validating that specific region. This PR updates the region regex to skip our specific backblaze region. I would just open a PR to enable backblaze specific regions, however, I am unsure how much region digit overlap there could be between AWS and Backblaze. Also, ex_aws admins are explicit about lib only caring about AWS-specific use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+v2.1.7
+
+- Various documentation updates
+- Add `comprehend` endpoint
+- Support firehose in region ca-central-1
+- More documentation fixes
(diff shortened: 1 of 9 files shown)


keatz55:
Wrong place

PR #747, opened 2021-03-04
Allow STS credentials to be injected by configuration

carrascoacd:
## Problem
In order to make `ex_aws` fully production-ready, we need to take into account the containerized applications that run, for example, over Kubernetes or Nomad. A typical configuration of secrets for this kind of application is by injection using environment variables. It is very common to restrict write permissions on the file system to avoid security problems also.

The aforementioned facts enter in conflict with the design of the STS module which resides on a file system file in order to assume a role for example.

## Proposed solution
I bring a proposal with implementation where I try to allow the injection of auth credentials by configuration, allowing the user to decide if using a file in the file system or not using the current design.

I also created the concept of `CredentialsIni.Provider` as I considered that it could be helpful for code injection and further features.

In order to make it works, I'll make use of the new function `ExAws.Config.awscli_auth_credentials/1` in the downstream `ex_aws_sts` library in this PR https://github.com/ex-aws/ex_aws_sts/pull/22.

The configuration could be like this:
```elixir
config :ex_aws,
  access_key_id: [{:awscli, "default", 30}],
  secret_access_key: [{:awscli, "default", 30}],
  awscli_auth_adapter: ExAws.STS.AuthCache.AssumeRoleCredentialsAdapter,
  awscli_credentials: %{
    "default" => %{
      role_arn: "",
      access_key_id: "",
      secret_access_key: "",
      source_profile: "default"
    }
  }
```
Please, let me know if it works for you or you have any improvement so we can release this and improve the library! Thanks.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -122,8 +122,19 @@ defmodule ExAws.Config do
 
   def parse_host_for_region(config), do: config
 
-  def awscli_auth_adapter do
-    Application.get_env(:ex_aws, :awscli_auth_adapter, nil)
+  def awscli_auth_adapter, do: Application.get_env(:ex_aws, :awscli_auth_adapter, nil)
+
+  def awscli_auth_credentials(profile, credentials_ini_provider \\ ExAws.CredentialsIni.File) do
+    case Application.get_env(:ex_aws, :awscli_credentials, nil) do
+      nil ->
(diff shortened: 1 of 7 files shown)


carrascoacd:
cc/ @bernardd

bernardd:
Hi @carrascoacd - sorry for not getting to this sooner. This looks really cool - thanks very much!

PR #748, opened 2021-03-05
Normalize path when using virtual host

VasileCimpean:
Have encountered some issues after switching to virtual host style paths, namely this:

```
** (ArgumentError) :path in URI must be nil or an absolute path if :host or :authority are given, got: %URI{authority: nil, fragment: nil, host: "some-bucket-name.s3.amazonaws.com", path: "some-key", port: 443, query: "", scheme: "https", userinfo: nil}
    (elixir 1.10.2) lib/uri.ex:662: String.Chars.URI.to_string/1
    (ex_aws 2.1.7) lib/ex_aws/request/url.ex:16: ExAws.Request.Url.build/2
    (ex_aws 2.1.7) lib/ex_aws/operation/s3.ex:30: ExAws.Operation.ExAws.Operation.S3.perform/2
```

Traced it back to `ExAws.Operation.S3.add_bucket_to_path`. When using non-virtual host bucket paths, `add_bucket_to_path` will automatically prepend a forward slash to the path to make it absolute. This does not happen when using virtual host bucket paths.

### DIFF
--- a/lib/ex_aws/operation/s3.ex
+++ b/lib/ex_aws/operation/s3.ex
@@ -26,6 +26,7 @@ defmodule ExAws.Operation.S3 do
 
       url =
         operation
+        |> normalize_path()
         |> add_resource_to_params()
         |> ExAws.Request.Url.build(config)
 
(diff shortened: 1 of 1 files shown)


lib/ex_aws/operation/s3.ex
```diff
@@ -63,6 +64,19 @@ defmodule ExAws.Operation.S3 do
...
+      normalized_path =
+        if String.first(operation.path) === "/" do
+          operation.path
+        else
+          Path.join(["/", operation.path])
+        end
+
+      operation |> Map.put(:path, normalized_path)
+    end
+
```
bernardd, line 79: Could this whole thing perhaps be simplified to something like:

```elixir
defp ensure_absolute_path(operation), do: put_in(operation.path, Path.join("/", operation.path])
```

since `Path.join/2` won't add supurflous slashes anyway?

bernardd:
Thanks for the contribution! I think it's probably a good idea, however I've added a comment about the code itself.  Cheers

lib/ex_aws/operation/s3.ex
VasileCimpean: Thanks for the idea, I changed it

VasileCimpean:
@bernardd I see that there is already code to handle absolute paths, so I think this can be closed

PR #765, opened 2021-03-21
Allow separate configuration of http_opts used for AWS metadata

TheoLeanse:
It is possible to configure the HTTP client on per-service basis. However this is not possible when fetching AWS metadata (specifically, fetching IAM credentials from the EC2 instance role). This risks metadata requests competing for connections with other requests and can lead to starvation in extreme cases.

This PR is an attempt to enable providing configuration separately for metadata requests, like so:

```
config :ex_aws, :metadata,
  http_opts: [
    pool: :ex_aws_metadata
]
```

More context is here: https://github.com/ex-aws/ex_aws/issues/598

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -35,3 +35,5 @@ config :ex_aws, :s3,
   scheme: "https://",
   host: "s3.amazonaws.com",
   region: "us-east-1"
+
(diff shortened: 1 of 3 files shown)


bernardd:
Very nice, thanks a lot @TheoLeanse.

PR #768, opened 2021-04-02
Relax requirement on Jason optional library to any 1.x version

jeffgrunewald:
Hi there!

Despite the mix.lock file currently using Jason 1.2.2 the way mix.exs currently tags the minimum required version for Jason is causing our applications to run into dependency conflicts with other libraries that have Jason as a dependency even though we are using Jason independently of any libraries and have upgraded to 1.2.2.

This appears to relax the restriction while still allowing you to use the latest version.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -36,7 +36,7 @@ defmodule ExAws.Mixfile do
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
       {:ex_doc, "~> 0.16", only: [:dev, :test]},
       {:hackney, "~> 1.9", optional: true},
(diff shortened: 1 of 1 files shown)


bernardd:
Hi @jeffgrunewald - Could you explain exactly what error you were getting? I'm not quite clear how this would be causing a problem if you're actually using 1.2.2. I don't object to the change as such, I'm just concerned that there may be issues with running ex_aws against Jason 1.0.x that I'm not aware of, and making this change could allow people to hit them.

jeffgrunewald:
I guess it's a moot point b/c it looks like we've separately removed the dependency that was causing the conflict so I'll go ahead and close this. Thanks!

PR #769, opened 2021-04-03
Add us-west-1 to list of supported ses services.

7hoenix:
What does this do?
-------------------

I was looking to send emails from my app and then I was running into the `<service> not supported in region <region> for partition aws` error as per this issue: https://github.com/ex-aws/ex_aws/issues/682

So this adds `us-west-1` as an option for sending emails.

Technical Considerations
-------------------------

- I *think* that `email` is the correct key in the configuration mapping (I was originally looking for `ses`).

- This is where I looked to see if `us-west-1` *should* be supported: https://docs.aws.amazon.com/general/latest/gr/ses.html

Happy to look in other places as well if that is not the best spot!

### DIFF
--- a/priv/endpoints.exs
+++ b/priv/endpoints.exs
@@ -1551,6 +1551,7 @@
             "eu-west-2" => %{},
             "us-east-1" => %{},
             "us-east-2" => %{},
+            "us-west-1" => %{},
             "us-west-2" => %{},
             "ca-central-1" => %{},
             "sa-east-1" => %{}


7hoenix:
Unfortunately, I've also checked the configuration in the `botocore/endpoints.json` file https://github.com/boto/botocore/blob/a0e22400396eabd5a58e4bae00f7e2c5dff2fdaf/botocore/data/endpoints.json#L2259-L2267 and it seems to match what we have here.

So I'm not sure if the solution of locking our config to botocore's as per this issue: https://github.com/ex-aws/ex_aws/issues/740#issuecomment-792186036 will be sufficient long term. But maybe I'm totally wrong?

bernardd:
Well that's a bit unfortunate. Thanks for checking it out though. I guess it's back to the drawing board to find an authoritative source for that data. Thanks for the patch :)

PR #773, opened 2021-04-12
Fix lead time in auth next refresh calculation

howleysv:
The current auth refresh time calculation

```elixir
max(expires_in_ms, expires_in_ms - @refresh_lead_time)
```

will always result in the auth being refreshed at the moment it expires rather than `lead_time` before expiration since `@refresh_lead_time` is positive. I believe the intention of this function is to return the ms to `lead_time`, unless it's negative and otherwise return 0 - meaning refresh now.


bernardd:
Good catch - thanks @howleysv

PR #775, opened 2021-04-12
make :http_opts for AWS metadata requests optional

velimir:
The change in 01fad68 introduced a new required parameter


bernardd:
Ah dangit - thanks very much @velimir. I'll get a new release with this fix shortly.

PR #783, opened 2021-04-28
Ensure absolute path for virtual hosted style S3 urls

VasileCimpean:
Closed an older PR (748) when I saw there was already a function called `ensure_absolute`. Then I tried out this latest version and noticed that it still does not work correctly with virtual hosted style S3 urls: the buckets are added to the url as subdomains, but the resource path is not ensured to be absolute. This PR fixes that issue.


bernardd:
Ah, nice one, thanks. It got added to the non-virtual-host style code, but not the virtual host one.

bernardd:
Whoops. Except this change breaks a unit test:
```
  1) test ensure paths ending with / preserve the trailing / (virtual-host) (ExAws.Operation.S3Test)
Error:      test/ex_aws/operation/s3_test.exs:67
     Assertion with == failed
     code:  assert processed_operation.path() == "folder/"
     left:  "/folder/"
     right: "folder/"
```

bernardd:
Looks like perhaps we may need a variant of `ensure_absolute` that preserves trailing slashes. Would that make sense?

VasileCimpean:
@bernardd Whoops, I'll take a look

PR #784, opened 2021-05-06
add "af-south-1" to S3 region in EX_AWS

KingMonty:
Was getting an error when trying to do operations of files in S3 inside the South Africa (af-south-1) region:

`** (RuntimeError) s3 not supported in region af-south-1 for partition aws`

Noticed the region wasn't added to the endpoints file for S3.


bernardd:
Thanks @KingMonty !

KingMonty:
Glad to help.

PR #785, opened 2021-05-07
Add telemetry events to all requests

claudioStahl:
As mentioned [here](https://github.com/ex-aws/ex_aws/issues/661).

> It would allow users of this library to capture important metrics and send them to their monitoring solution. This way, we save everyone the time to instrument their own usage of this library.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -39,13 +39,7 @@ defmodule ExAws.Request do
         )
       end
 
-      case config[:http_client].request(
-             method,
-             safe_url,
-             req_body,
-             full_headers,
-             Map.get(config, :http_opts, [])
-           ) do
+      case do_request(config, method, safe_url, req_body, full_headers, attempt) do
         {:ok, %{status_code: status} = resp} when status in 200..299 or status == 304 ->
           {:ok, resp}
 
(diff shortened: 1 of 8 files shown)


lib/ex_aws/request.ex
```diff
@@ -104,6 +98,31 @@ defmodule ExAws.Request do
...
+        {:ok, %{status_code: status}} = result when status in 200..299 or status == 304 ->
+          telemetry_metadata = Map.put(telemetry_metadata, :result, :ok)
+          {result, telemetry_metadata}
+
+        result ->
+          telemetry_metadata = Map.put(telemetry_metadata, :result, :error)
+          {result, telemetry_metadata}
+      end
+    end)
+  end
```
bernardd, line 124: A minor stylistic thing - rather than duplicating the `Map.put` and result tuple construction in each case block, it might be a little cleaner to do something like:

```elixir
result = config[:http_client].request(...)
telemetry_result = case result do
  {:ok, %{status_code: status}} = result when status in 200..299 or status == 304 -> :ok
  _ -> :error
end

telemetry_metadata = Map.put(telemetry_metadata, :result, telemetry_result)
{result, telemetry_metadata}
```

bernardd:
That one minor comment notwithstanding, this looks fantastic - thanks heaps (and especially for including documentation and tests).

lib/ex_aws/request.ex
claudioStahl: thanks for the tip :)

PR #788, opened 2021-05-12
Replaces pattern-matching by regex to remove leading duplicate spaces on auth request headers

luizmiranda7:
When executing some profiles on services that uses ex_aws, for low CPU resource services, on of the main performance issues is the way this code is replacing the leading duplicate spaces.

To avoid this type of performance issue, this PR aims to replace the pattern matching code style by a regex match/replace mechanism.

### DIFF
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -192,8 +192,11 @@ defmodule ExAws.Auth do
     |> IO.iodata_to_binary()
   end
 
-  def remove_dup_spaces(""), do: ""
-  def remove_dup_spaces(<<value::binary>>), do: Regex.replace(~r/^\s+/, value, " ")
+  defp remove_dup_spaces(str), do: remove_dup_spaces(str, "")
+  defp remove_dup_spaces(str, str), do: str
+
+  defp remove_dup_spaces(str, _last),
+    do: str |> String.replace("  ", " ") |> remove_dup_spaces(str)
 
   defp string_to_sign(request, service, datetime, config) do
     request = hash_sha256(request)
--- a/lib/ex_aws/auth.ex
+++ b/lib/ex_aws/auth.ex
@@ -192,12 +192,8 @@ defmodule ExAws.Auth do
     |> IO.iodata_to_binary()
   end
 
-  defp remove_dup_spaces(""), do: ""
-  defp remove_dup_spaces("  " <> rest), do: remove_dup_spaces(" " <> rest)
-
-  defp remove_dup_spaces(<<char::binary-1, rest::binary>>) do
-    char <> remove_dup_spaces(rest)
-  end
+  def remove_dup_spaces(""), do: ""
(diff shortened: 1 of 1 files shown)


luizmiranda7:
@bernardd, what you think about it?

bernardd:
So a couple of things - first, for questions of performance it's nice to have some numbers or at least example data to deal with. That way we can be sure that a) there's a problem, and b) the change will fix it.

Second, the regex in the PR doesn't actually produce the same output as the original function. It only matches spaces at the start of the string, rather than those anywhere in the string. That's easily fixed by removing the `^` though.

Third, I'm predisposed to avoiding regexs by default if I can - they aren't especially easy to read, and can themselves have performance characteristics that are unpredictable (I say this having listened to many a rant from the RE2 maintainer who is firmly of the opinion that regexes should be avoided wherever possible :)). I'm not saying I think this particular regex is going to be problematic in terms of performance - it's quite simple - just that if there's an equally easy way to solve the problem without them, that's generally what I prefer.

Having said all that, I put together a few tests:

```elixir
defmodule Test do
  # Original
  def remove_dup_spaces1(""), do: ""
  def remove_dup_spaces1("  " <> rest), do: remove_dup_spaces1(" " <> rest)

  def remove_dup_spaces1(<<char::binary-1, rest::binary>>) do
    char <> remove_dup_spaces1(rest)
  end


  # Regex
  def remove_dup_spaces2(""), do: ""
  def remove_dup_spaces2(<<value::binary>>), do: Regex.replace(~r/\s+/, value, " ")


  # Dumb string ops
  def remove_dup_spaces3(str), do: remove_dup_spaces4(str, "")
  defp remove_dup_spaces3(str, str), do: str
  defp remove_dup_spaces3(str, _last), do: str |> String.replace("  ", " ") |> remove_dup_spaces3(str)
end
```

(I removed a couple of other attempts I made at being smart because they were both more complicated and less performant than the regex).

Then I just fed them a "longish" string with a few runs of around 10 spaces. The runtimes in microseconds were around:
Original: 140-150
Regex: 30-80
String ops: 20-25

So already the string op is quicker, but what's really interesting is when I fed it the same string but with the spaces already de-duplicated (which honestly seems likely to be the majority of cases):

Original: 100-150
Regex: 30-40
String ops: 9-14

So while I totally buy the claim that the function is taking up a non-trivial time on low performance machines, I think we can do better than a regex to fix it. If you can try out the string op version and let me know if that works for you, that'd be great.

Cheers

luizmiranda7:
Hey @bernardd, nice caught!

First, thank for the tips of my PR description, it has a lot to improve. And finally, your approach keeps the "core" of the solution that is to reduce the recursion, using the direct replace of double spaces by single, in order to improve the CPU usage on low resource machines.

I believe it will works better them mine and also as soon as it is applied I can increase the version on my projects and test in in "real life" production services.

I have applied your changes.

PR #789, opened 2021-05-16
LICENSE file into hexpm

philippneugebauer:
Also I there's a license in `mix.exs` stated, I think it's better to also copy the LICENSE file into hexpm for better automatic license detection

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -47,7 +47,7 @@ defmodule ExAws.Mixfile do
   defp package do
     [
       description: description(),
-      files: ["priv", "lib", "config", "mix.exs", "README*"],
+      files: ["priv", "lib", "config", "mix.exs", "README*, LICENSE"],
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -47,7 +47,7 @@ defmodule ExAws.Mixfile do
   defp package do
     [
       description: description(),
-      files: ["priv", "lib", "config", "mix.exs", "README*"],
+      files: ["priv", "lib", "config", "mix.exs", "README*, LICENSE"],
```
bernardd, line 50: Should this be `["priv", "lib", "config", "mix.exs", "README*", "LICENSE"],` instead?

philippneugebauer: ups, yeah

philippneugebauer: should be fixed now

bernardd:
Thanks @philippneugebauer !

PR #790, opened 2021-05-17
Add expected sequence token to error response

VasileCimpean:
After https://github.com/ex-aws/ex_aws/commit/ed5d8ce628cf06951fb89e2a49a0c26c57af861d# a breaking change was introduced. 
Before this, exceptions like DataAlreadyAcceptedException and InvalidSequenceTokenException produced errors that contained the expected sequence token and it could be directly used from the error map itself. Currently, errors are of the format `{:error, {type, message}}`, with the message containing the expected sequence token. To avoid parsing that error message, we can provide that token directly, like `{:error, {type, message, token}}` for the types of errors that contain this token.
I've also looked at the possibility of just having `{:error, reason}` where reason is the whole map returned by aws, containing the error type, message, and any other data like the token, but this would have broken too many clients.
This is still a breaking change, but it seems like the least invasive one.

### DIFF
--- a/lib/ex_aws/operation/s3.ex
+++ b/lib/ex_aws/operation/s3.ex
@@ -55,7 +55,8 @@ defmodule ExAws.Operation.S3 do
 
     def add_bucket_to_path(operation, %{virtual_host: true, host: base_host} = config) do
       vhost_domain = "#{operation.bucket}.#{base_host}"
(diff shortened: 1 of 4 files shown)


VasileCimpean:
I'll do some proper branching for this, closing

PR #791, opened 2021-05-17
Return expected sequence token as part of error result

VasileCimpean:
After https://github.com/ex-aws/ex_aws/commit/ed5d8ce628cf06951fb89e2a49a0c26c57af861d# a breaking change was introduced. 
Before this, exceptions like DataAlreadyAcceptedException and InvalidSequenceTokenException produced errors that contained the expected sequence token and it could be directly used from the error map itself. Currently, errors are of the format `{:error, {type, message}}`, with the message containing the expected sequence token. To avoid parsing that error message, we can provide that token directly, like `{:error, {type, message, token}}` for the types of errors that contain this token.
I've also looked at the possibility of just having `{:error, reason}` where reason is the whole map returned by aws, containing the error type, message, and any other data like the token, but this would have broken too many clients.
This is still a breaking change, but it seems like the least invasive one.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -123,15 +123,19 @@ defmodule ExAws.Request do
     {:error, {:http_error, status, error}}
   end
 
-  def handle_aws_error("ProvisionedThroughputExceededException" = type, message) do
+  def handle_aws_error("ProvisionedThroughputExceededException" = type, message, _) do
     {:retry, {type, message}}
   end
 
-  def handle_aws_error("ThrottlingException" = type, message) do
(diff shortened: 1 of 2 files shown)


bernardd:
Yep, that seems reasonable (especially since that's more or less the form the error was in prior to that change anyway), and given that it's only really "breaking" for stuff that the previous change already broke :) Thanks for the fix.

ruslandoga:
@VasileCimpean @bernardd just a heads up: it seems to be [breaking](https://github.com/heyoutline/cloud_watch/issues/24) `:cloudwatch`.

bernardd:
Dangit, thanks @ruslandoga. I probably should have done at least a minor version bump for this one.

PR #796, opened 2021-05-24
fix: prevent unrelated `awscli` config from leaking

### LINKED ISSUE #521: ExAws Config Overriding Not Applying When Config File Present
Followup from the elixirforum thread [here](https://elixirforum.com/t/exaws-config-overriding-not-applying/10577)

### Environment

* Elixir version: 1.5.2
* ExAws version: 2.0.2
* Hackney version: 1.10.1

### Current behavior

Trying to override the region of requests in the following manner results in an error. 

```elixir
iex(2)> conf = ExAws.Config.new(:sqs, [region: "us-west-1"])
%{access_key_id: "MY_ACCESS_KEY_ID", host: "sqs.us-west-1.amazonaws.com",
  http_client: ExAws.Request.Hackney, json_codec: Poison, port: 443,
  region: "us-east-1",
  retries: [max_attempts: 10, base_backoff_in_ms: 10, max_backoff_in_ms: 10000],
  scheme: "https://",
  secret_access_key: "MY_SECRET_KEY"}

iex(4)> ExAws.SQS.list_queues() |> ExAws.request(conf)
{:error,
 {:http_error, 403,
  %{code: "SignatureDoesNotMatch", detail: "",
    message: "Credential should be scoped to a valid region, not 'us-east-1'. ",
    request_id: "my_request_id", type: "Sender"}}}
```

It is important to notice that my `~./aws/config` file has its region set to `us-east-1`; however, it should still be possible to override this. 

### Expected behavior

The above snippet should return a configuration map equivalent to the following:
```elixir
iex(4)> conf2 = conf |> Map.merge(%{:region => "us-west-1"})
iex(5)> ExAws.SQS.list_queues() |> ExAws.request(conf2) # returns the correct result
```

superhawk610:
Here's a simple config setup:

```elixir
# config.exs

config :ex_aws,
  region: "us-east-1",
  access_key_id: {:awscli, "default", 30},
  secret_access_key: {:awscli, "default", 30}
```

```toml
# ~/.aws/config

[default]
region = "us-west-2"
```

I would expect `ExAws.Config.new(:s3)` to return a config where `region` is `"us-east-1"`, but it incorrectly returns `region: "us-west-2"` (all values configured in `~/.aws/{config, credentials}` take precedence over any other configured values, even if their keys aren't specified).

This PR modifies `ExAws.Config.retrieve_runtime_config/1` to only set the requested key(s), instead of merging in all available values.

Resolves #521.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -77,7 +77,7 @@ defmodule ExAws.Config do
 
       {k, v}, config ->
         case retrieve_runtime_value(v, config) do
-          %{} = result -> Map.merge(config, result)
+          %{} = result -> Map.put(config, k, result[k])
           value -> Map.put(config, k, value)
         end
     end)


bernardd:
Ah, nice find @superhawk610. Thanks heaps.

bernardd:
Hey @superhawk610 - just a note to let you know that I'm reverting this in v2.2.5 because it caused issues with some other config stuff (see #814). The issue is that some types of config (notably `:instance_role`) rely on the behaviour of *all* the config values being merged. And the same goes for some values in `:awscli`, for that matter (though I have no idea if anyone is relying on that particular behaviour). So what I plan to do is a bit of a rework of that whole thing so that those merges still happen but only if the value is not explicitly set (as in your case). Unfortunately it's a bit more than a 1-liner, so it's going to take me a little while.

superhawk610:
@bernardd Thanks for the explanation, hope this doesn't cause too much headache! I figured there was a reason that `.merge()` was used originally, but I couldn't figure out why. This should provide good context for anyone that stumbles across this in the future.

Thanks again for your work on this!

PR #801, opened 2021-07-13
Improving AWS partition detection

Nitrino:
I am trying to use `ex_aws` with Oracle Objects Storage and I receive an error `s3 not supported in region us-sanjose-1 for partition aws`. This happens due to the same region formats for AWS and Oracle.

Example: 
`us-sanjose-1` matches the given regular expression
https://github.com/ex-aws/ex_aws/blob/56392c150609cb4ecb1081f88998c7468c5079f6/lib/ex_aws/config/defaults.ex#L80

This PR improves AWS region detection logic by clarifying the second part of the region name.

I also added support for the `me` zone for `me-south-1` Middle East (Bahrain) region


bernardd:
Hi Nitrino - thanks for the PR! Just to help me understand, what's the issue with using OOS? Is there some module to support it that plugs into ex_aws or something? Might this still be a problem if OOS ended up using a region name which matched one of AWS's?

Nitrino:
Hi @bernardd 
I use `ex_aws` without third-party modules. It works great, as OOS API has compatibility with AWS S3.

Yes, in the future there may be a problem if Oracle adds the same name of the region as the AWS. In this case, it will be better to keep the accurate list of AWS regions or add setting the control of this logic. 
But while I checked the full list of Oracle regions and there is no overlap with AWS, so I decided to just clarify the regular expression that would not make big changes to the library

bernardd:
Hey - sorry for not getting back to you sooner. Thinking about this, I wonder if a better approach might not be some config parameter that allows us to skip the check entirely. Something like `default_endpoint_config: false` (defaulting to true). That would allow any compatible system, even ones with identical region names, to work without issue. Does that sound like a reasonable plan?

Nitrino:
Hey @bernardd At the moment, I do not know of compatible systems with same regions, but I think the config will still be a more correct solution.
Let me change the PR for the config?

PR #804, opened 2021-07-28
Update regions for SageMaker runtime endpoint

ashatgit:
Update regions for SageMaker runtime endpoint as per https://docs.aws.amazon.com/general/latest/gr/sagemaker.html


bernardd:
Thanks @ashatgit !

PR #806, opened 2021-08-04
Remove unimplemented task

josevalim:
Perhaps, if it is to be implemented, it should be done in ExAWS.Kinesis.


bernardd:
Nice, thanks @josevalim. Looks like that's been sitting there quite some time, before the different service modules were even split off.

PR #807, opened 2021-08-05
Update regions for SES (mail)

adarsh-sonio:
Updating the regions under SES (mail) as per the latest docs - https://docs.aws.amazon.com/general/latest/gr/ses.html


bernardd:
Thanks @adarsh-sonio !

Issue #802, opened 2021-07-19
Support SESSION_TOKEN to authenticate


benkeil:
It would be nice to have more login options in order to work with assume role or awsume.

trbngr:
this already works for me in a lambda.

```
config :ex_aws,
  access_key_id: {:system, "AWS_ACCESS_KEY_ID"},
  security_token: {:system, "AWS_SESSION_TOKEN"},
  secret_access_key: {:system, "AWS_SECRET_ACCESS_KEY"}
```

benkeil:
Ok, it works. Then we should just add it to the documentation. And maybe we could rename security_token to session_token to be more conform with the aws api.

bernardd:
If you're able to do a PR to add it to the docs that would be awesome.

benkeil:
It would be a breaking change, how to deal with it? add a new parameter and deprecate the old one?

bernardd:
No, I just meant a PR to add it to the docs. We could make both names work, sure, but the feature already exists so I don't see huge value in doing that.

Issue #709, opened 2020-08-18
Could not find how to add security token (AWS Session Token) in the documentation


sulabh-kothari:
* Elixir  1.10.4 
* ExAws version 2.1.4

The README file does not tell how to set the AWS session token. Our call was continuously failing to connect with AWS with the 2 parameters set (`access_key_id` and `secret_access_key`). These are the only 2 parameters described in the readme documentation. We had to figure out from the source code that aws session token can be set as `security_token`. It will be better if this is also added to the readme doc file. 
We also asked this question  in the Elixir forum https://elixirforum.com/t/unable-to-configure-ex-aws-for-sqs/33795 and updated with the answer.
You can reach out to me for more details around it at sulabh.kothari@deserve.com.

Nitrodist:
@sulabhk-deserve same

benwilson512:
https://github.com/ex-aws/ex_aws#deprecation-notice

bernardd:
I'm more than happy for this to be added to the docs if someone would be kind enough to sumit a PR.

PR #812, opened 2021-08-30
Update README.md to document Session Token support

kw7oe:
Address #802, #709.


bernardd:
Thanks very much @kw7oe !

PR #816, opened 2021-09-04
Update deps, allow telemetry 1.0, restrict mime to 1.x

ZombieHarvester:
- There are no changes in the telemetry 1.0.0 compared to 0.4.3 - it marks the stability of the API.
- Restrict mime version to 1.x as mime 2.x requires Elixir ~> 1.10


bernardd:
Nice, thanks @ZombieHarvester.

PR #820, opened 2021-09-14
Pass Headers to Request

jacobsmith:
This change allows HTTP headers to be passed in. This was originally questioned when it was merged in here: https://github.com/ex-aws/ex_aws/pull/286#discussion_r90781530 but wasn't necessary for the use case at the time. However, attempting to use this with Cloudfront (and presumably any other service that utilizes ETag) makes this necessary.

This should add Headers if present, otherwise, leave them as an empty array. So, there should be no functional impact for existing users of this method.

### DIFF
--- a/lib/ex_aws/operation/rest_query.ex
+++ b/lib/ex_aws/operation/rest_query.ex
@@ -12,7 +12,7 @@ end
 
 defimpl ExAws.Operation, for: ExAws.Operation.RestQuery do
   def perform(operation, config) do
-    headers = []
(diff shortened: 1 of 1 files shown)


bernardd:
Looks good to me - thanks @jacobsmith

jacobsmith:
Thanks for your work on this, @bernardd !

bernardd:
Short answer is "when I get around to it", which is usually when I have a few minutes of spare time on the weekend. I'll try to get a new release this weekend for you.

PR #827, opened 2021-10-11
Relax JSX requirement

dbern-stripe:
Adjusts [dependabot's automated PR](https://github.com/ex-aws/ex_aws/pull/759)
I don't see that jsx 3.x is actually required, so this makes it flexible again.


bernardd:
Thanks - I have no idea why dependabot actually edited the mix.exs file - I've never seen it do that before.

PR #832, opened 2021-10-18
Support passing headers

jacobsmith:
Passing Headers, such as `If-Match`, was previously causing issues. This is because the following code was executing when passing an array of header values:

```elixir
  def retrieve_runtime_value(values, config) when is_list(values) do
    values
    |> Stream.map(&retrieve_runtime_value(&1, config))
    |> Enum.find(& &1)
  end
```

which destructured the data from `[{ "header-name", "header-value"}]` to `{ "header-name", "header-value"}` (note that it is no longer a list containing a tuple, but rather simply a tuple.

This caused issues later when trying to build up the headers to send in the request. Abbreviated for simplicity, we had:

```elixir
headers = { "some-key", "some-value" }
headers = [ defaultHeaders | headers ]
```

Because `headers` was no longer a `List`, the `headers` variable now contains an improper List.

This change passes `config.headers` through to the underlying functions without destructuring it first.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -74,7 +74,7 @@ defmodule ExAws.Config do
 
       {:telemetry_options, telemetry_options}, config ->
         Map.put(config, :telemetry_options, telemetry_options)
-        
+
       {:headers, headers}, config ->
         Map.put(config, :headers, headers)
 
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -74,6 +74,9 @@ defmodule ExAws.Config do
 
       {:telemetry_options, telemetry_options}, config ->
         Map.put(config, :telemetry_options, telemetry_options)
+        
+      {:headers, headers}, config ->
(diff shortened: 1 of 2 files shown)


jacobsmith:
Sure thing @bernardd

anthonator:
@bernardd is there anything else we can do to help move this along?

bernardd:
Hey - sorry, been super busy with work. I'm hoping to catch up on this and other items this weekend.

jacobsmith:
@bernardd No worries - thanks for all your work on this/these libraries! We're currently running a fork with this one patch applied, so definitely not a blocker for us, but for obvious reasons we don't want to maintain a fork forever (:

bernardd:
Thanks again @anthonator and apologies for the delay getting this in.

anthonator:
@jacobsmith is the guy to thank but thanks for getting this merged in @bernardd!!!

jacobsmith:
Thanks @bernardd !

bernardd:
Whoops - thanks @jacobsmith :)

PR #838, opened 2021-11-03
Use '?\s' instead of '?\ '

wingyplus:
Found on Elixir 1.13, the compiler warning about usage of `?\ `. Fix
by using `?\s` instead.


bernardd:
Nice, thanks @wingyplus

PR #839, opened 2021-11-16
Show docs for `ExAws.Config`

### LINKED ISSUE #673: Documentation hidden for `ExAws.Config`
Hi, I've been trying to use the library with s3 (specifically `ExAws.S3.presigned_url`) and was completely baffled by what the first argument, `config` was. While the documentation there could definitely be improved, it didn't help that the required module (`ExAws.Config`) was completely hidden!

Is there any reason for this since a cursory search through the code and tests show that this is the correct way to create the required config?

It is explicitly disabled here:
https://github.com/ex-aws/ex_aws/blob/7843167836f6ee291241073be21118314e554533/lib/ex_aws/config.ex#L2

kenny-evitt:
Closes #673

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -1,10 +1,13 @@
 defmodule ExAws.Config do
-  @moduledoc false
+  @moduledoc """
+  Generates the configuration for a service
 
-  # Generates the configuration for a service.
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @kenny-evitt

PR #845, opened 2021-12-10
Support for me-south-1 - region regex, CW Logs and Metrics

### LINKED ISSUE #844: Support for me-south-1 for cloud watch logs and cloud watch metrics
I am using the library https://github.com/noxdafox/rabbitmq-cloudwatch-exporter which internally uses this library for the AWS functions.
We have deployed in the me-south-1(Bahrain region).
The me-south-1 region supports cloudwatch metrics and also cloud watch logs but the library needs to be updated for the same.

I am thinking that the changes are minor and would require changes in the below sections:
https://github.com/ex-aws/ex_aws/blob/main/priv/endpoints.exs#L12

https://github.com/ex-aws/ex_aws/blob/main/priv/endpoints.exs#L614
and
https://github.com/ex-aws/ex_aws/blob/main/priv/endpoints.exs#L291

Is this fine??

avivek:
Added Support for me-south-1 in region regex and added endpoint support for CW Logs and CW Metrics.
closes #844


bernardd:
Perfect, thanks @avivek!

avivek:
@bernardd would it be possible to create a release tag so that the downstream library can update their dependencies??

Issue #770, opened 2021-04-06
Improvement: raise error when no bucket is available


HoneyryderChuck:
This is a kind improvement request, to save some times to the next person that falls into this trap.

I'd like that calls to S3 fail reliably when no bucket is passed, instead of passing nils downstream and let the lower layers fail in unreliable and cryptic ways. This refers to a given example like:

```elixir
ExAws.S3.get_object(
  bucket_name(), # read from env
  "file-name-#{uuid}"
)
|> ExAws.request()
#  ...
```

and when `bucket_name()` call returns nothing, because the environment variable isn't set.

### Environment

* Elixir & Erlang versions (elixir --version): irrelevant, but `Elixir 1.7.4`
* ExAws version: 2.1.0
* HTTP client version: hackney 1.15.2

### Current behavior

You get a stacktrace similar to:

```
(FunctionClauseError) no function clause matching in :inet_tcp.do_connect/4
             (kernel) inet_tcp.erl:113: :inet_tcp.do_connect({127, 0, 0, 1}, 363577, [{:packet, :raw}, {:active, false}, :binary], 8000)
             (kernel) gen_tcp.erl:185: :gen_tcp.try_connect/6
             (kernel) gen_tcp.erl:163: :gen_tcp.connect/4
             (hackney) /app/deps/hackney/src/hackney_connect.erl:277: :hackney_connect.do_connect/5
             (hackney) /app/deps/hackney/src/hackney_connect.erl:47: :hackney_connect.connect/5
             (hackney) /app/deps/hackney/src/hackney.erl:333: :hackney.request/5
             (ex_aws) lib/ex_aws/request/hackney.ex:21: ExAws.Request.Hackney.request/5
             (ex_aws) lib/ex_aws/request.ex:42: ExAws.Request.request_and_retry/7
             (ex_aws) lib/ex_aws/operation/s3.ex:39: ExAws.Operation.ExAws.Operation.S3.perform/2
```

I got such errors from service level tests, where a random available port is selected for each test. Notice that `363577` is an invalid port number (0..65535 range it should). After some more debugging, I found out that a 7 was being appended to every port number the test server was connecting to. So for the case above, the server accepts binds to `36357`, but ex_aws connects to 363577, and so on.

I didn't get further down the spiral to find out where that 7 is coming though.

### Expected behavior

I'd expect an error such as "bucket name is missing", given that this is required.

bernardd:
[...]

PR #853, opened 2022-01-16
Raise when no bucket is available

lex-mala:
Attempting to address the request here: #770 

This could also go somewhere in `ExAws.S3`. Happy to amend the error itself and its message to something more idiomatic to the project if necessary.


bernardd:
Looks good, thanks @lex-mala!

bernardd:
Resolves #770

PR #855, opened 2022-01-19
Add all supported regions for transcribe

aymanosman:
Taken from https://docs.aws.amazon.com/general/latest/gr/transcribe.html


bernardd:
Thanks @aymanosman!

PR #856, opened 2022-01-23
Add a few specs and polish docs here and there

whatyouhide:


bernardd:
Nice one, thanks @whatyouhide!

PR #858, opened 2022-02-04
Add Location service to "aws" partition

thecristen:
Enables the usage of [AWS Location Service](https://docs.aws.amazon.com/location/latest/developerguide/what-is.html) for the relevant regions.

This includes the maps, places, geofencing, tracking, and routes services, for [nine supported regions](https://docs.aws.amazon.com/location/latest/developerguide/location-regions.html).


bernardd:
Nice, thanks @thecristen!

PR #860, opened 2022-02-20
Instance Metadata Service (IMDS) v2

justincjohnson:
I finally took the time to revive https://github.com/ex-aws/ex_aws/pull/739 with passing tests.  Sorry to anyone who was waiting for this.  From the original PR...

> I'm using AWS's Security Hub service, and it recently created a finding for my environment, specifically [EC2.8] EC2 instances should use IMDSv2. The remediation details can be found at https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-8-remediation, but in short, the new version of the service requires an extra API call to generate a token, and then subsequent API calls need to pass that token in a header.

The IMDSv2 code will work even on a system that isn't enforcing v2, so from my experience it is safe to use this code regardless of how your EC2 instances are configured.

I also officially added HTTPotion as a test dependency since it was used in some tests, but since it is currently deprecated, a follow up change should probably switch that to Tesla or something.

Thank you for your time @bernardd.  🙏🏻


justincjohnson:
Gentle bump @bernardd, at your convenience.  Thank you.

bernardd:
Looks great, thanks @justincjohnson  - apologies for the delay - the first request slipped through the cracks.

PR #865, opened 2022-04-01
Add support for AWS SSO temporary credentials

james-gibson-fr:
Work to resolve https://github.com/ex-aws/ex_aws/issues/843 based on the fork from @joeybaer. Huge thank you to both Joey and @patrickcsullivan for help and feedback on the PR!

Adds support in `ExAws.CredentialsIni.File` to pull the credentials expected throughout `ex_aws` from the temporary credentials provided by `aws sso login` stored in `.aws/sso/cache/`. This way users (like me) don't have to run a script prior to running elixir to generate the credentials and can instead just rely on the sso config like the aws cli does.


james-gibson-fr:
Whoops, going to try to clean up the history here

PR #871, opened 2022-04-13
Enhance HTTP methods support docs

### LINKED ISSUE #870: HEAD and OPTIONS verbs missing from the HTTP method type spec
### Environment

* Elixir & Erlang versions (elixir --version): 1.12.3
* ExAws version 2.3.0
* ExAwsS3 version 2.3.2
* HTTP client version. 1.18.1

### Current behavior

The HTTP method [type spec](https://github.com/ex-aws/ex_aws/blob/main/lib/ex_aws/request/http_client.ex#L48) is a subset of the supported verbs. [HEAD](https://github.com/ex-aws/ex_aws_s3/blob/main/lib/ex_aws/s3.ex#L824) and [OPTIONS](https://github.com/ex-aws/ex_aws_s3/blob/main/lib/ex_aws/s3.ex#L848) are also in use.

### Expected behavior
The extra verbs should be added to the typespec

gesta:
Closes #870

### DIFF
--- a/lib/ex_aws/request/http_client.ex
+++ b/lib/ex_aws/request/http_client.ex
@@ -45,7 +45,7 @@ defmodule ExAws.Request.HttpClient do
 
   """
 
-  @type http_method :: :get | :post | :put | :delete
+  @type http_method :: :get | :post | :put | :delete | :options | :head
   @callback request(
               method :: http_method,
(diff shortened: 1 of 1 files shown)


bernardd:
Nice one, thanks @gesta

PR #877, opened 2022-04-27
Convert IMDSv2 header value to string

BryanLeong:
`InstanceMetaTokenProvider.request_token/1` currently uses an integer header value (`@metadata_token_ttl_seconds`), which breaks the contract for the `request` callback on `ExAws.Request.HttpClient`.

Similar to ex-aws/ex_aws#710


bernardd:
Thanks @BryanLeong!

PR #881, opened 2022-05-17
Handling empty paths sent to AWS

pcvyas:
Sometimes a URL may not specify a path to AWS, which in turn can cause an issue while a signature is being created for that URL.

**Current Behaviour:**

The following flow is executed when populating the headers

headers -> auth_headers -> signature -> get_path 

In get_path's current implementation, if the path is empty, it will return a nil value (which causes a matching error further on)

For example:

image

**Expected Behaviour:**

This PR adds a small change that allows for empty paths to be sent and the signature still being created for the specified URL. Instead of returning a nil value, get_path would return "/" upon discovering an empty path

**Note**:

Following the contribution guide here https://github.com/ex-aws/ex_aws/blob/bdc758048042e5db86179c843fdbbd9cbd2bbd4e/CONTRIBUTING.md

I tested my changes locally, added a unit test and also ran the automated tests (with a local DynamoDB instance running). I also ran mix format and mix dialyzer before pushing my changes.

### DIFF
--- a/lib/ex_aws/request/url.ex
+++ b/lib/ex_aws/request/url.ex
@@ -103,7 +103,7 @@ defmodule ExAws.Request.Url do
     path
   end
 
-  def get_path(url, _), do: URI.parse(url).path
+  def get_path(url, _), do: URI.parse(url).path || "/"
 
   def uri_encode(url), do: URI.encode(url, &valid_path_char?/1)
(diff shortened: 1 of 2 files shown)


bernardd:
Nice, thanks @pcvyas !

Issue #882, opened 2022-05-20
Unable to get error details


SteffenDE:
So I'm currently experimenting with AWS Timestream. The ex_aws_timestream repo works fine, but when there is a `RejectedRecordsException`, there seems to be no way of getting the index of the failed record.

This seems to be caused by this line https://github.com/ex-aws/ex_aws/blob/903eed1550949a0df43f025b126f19868218b22b/lib/ex_aws/request.ex#L152 that does not return any error details.

The plain response looks like this:
```elixir
{:ok, 419,
 [
   {"x-amzn-RequestId", "..."},
   {"Content-Type", "application/x-amz-json-1.0"},
   {"Content-Length", "370"},
   {"Date", "Fri, 20 May 2022 15:01:56 GMT"},
   {"Connection", "close"}
 ],
 "{\"__type\":\"com.amazonaws.timestream.v20181101#RejectedRecordsException\",\"Message\":\"One or more records have been rejected. See RejectedRecords for details.\",\"RejectedRecords\":[{\"ExistingVersion\":1,\"Reason\":\"A record already exists with the same time, dimensions, measure name, and version 1. A higher version is required to update the measure value.\",\"RecordIndex\":17}]}"}
```

ex_aws only returns `{:error, {"RejectedRecordsException", "One or more records have been rejected. See RejectedRecords for details."}}`

### Environment

* Elixir & Erlang versions (elixir --version): Elixir 1.13.4 (compiled with Erlang/OTP 24)
* ExAws version `mix deps |grep ex_aws`: ex_aws 2.3.2
* HTTP client version. IE for hackney do `mix deps | grep hackney`: hackney 1.18.1

### Current behavior

There is no way to get any error details.

### Expected behavior

ex_aws should provide a way to access error details.

bernardd:
The problem here is that we have a common error handling system for all services, even though they can all return different service-specific error info. We currently have a couple of special-case handlers, but really the handling should be delegated to the service that made the request in the first place.

What I'd like to do to resolve this is allow the service module to define a particular callback, and have the generic error handler look for that and call it if it exists (falling back to the existing defaults if it doesn't). I envisage this being done by adding an optional `service_callback_module` field to the `Operation` structs which is then checked by `handle_aws_error` for the required callback. That would also make it nicely extensible for any other service-specific post-request functionality we wanted to add down the road.

bernardd:
This feature is now available as of 2.4.0. See #902

PR #884, opened 2022-06-06
Log QueueUrl for failed SQS requests

johnnyt:
Currently when SQS requests fail, a warning is logged out which includes a URL.

`[warn] ExAws: HTTP ERROR: :request_timeout for URL: "https://sqs.us-west-2.amazonaws.com/" ATTEMPT: 1`

 This URL does not include the QueueUrl which, when working with multiple SQS queues, makes it impossible to know which SQS queue is having a problem.

This MR updates this log message when working with the `:sqs` service to include the QueueUrl instead:

`[warn] ExAws: HTTP ERROR: :request_timeout for QUEUE: "https://sqs.us-west-2.amazonaws.com/ACCOUNT_NUMBER/QUEUE_NAME" ATTEMPT: 1`


bernardd:
Hi @johnnyt - thanks for the PR!
As it happens, I was just thinking about this exact type of problem over in https://github.com/ex-aws/ex_aws/issues/882#issuecomment-1149519665

The issue is that I don't really want to be jamming all the different error handling for every single service into the main ExAws library - kind of defeats the point of having different service libraries in the first place.

I realise it's a bit more of a task than just sticking an `if service == :my_service` block in here, but really that's not a sustainable way to deal with the problem. If you're keen, though, I'd definitely welcome (and be happy to help with) a PR implementing the solution mentioned in the comment above.

johnnyt:
@bernardd - thanks for the feedback. I'll look into that other comment in #882 and the proposed solution there.

bernardd:
The feature discussed in #882 is now available in version 2.4.0 - please check it out.

PR #886, opened 2022-06-13
Include service in telemetry events

hoyon:
Exposes service in telemetry events so that it's easier to see which services are being used in our monitoring.

I'm also wondering if there's a way to expose the name of the API being used. eg `GetObject` or `PutObject` for S3.


bernardd:
> I'm also wondering if there's a way to expose the name of the API being used. eg GetObject or PutObject for S3.

I think you'd need an extra field in the request that was set by every service. It's certainly doable, but would be a non-trivial task (unless you can think of a more clever way).

bernardd:
Looks good to me, thanks @hoyon!

PR #888, opened 2022-06-19
Improve resiliency on auth adapter when transient errors happen

### LINKED ISSUE #798: ExAws.Config.retrieve_runtime_value getting throttled
### Environment

Elixir 1.11.4
Erlang/OTP 23 [erts-11.2.2.3] [source] [64-bit] [smp:2:1] [ds:2:1:10] [async-threads:1]

```
      {:ex_aws, "~> 2.2.1"},
      {:ex_aws_ssm, "~> 2.1.0"},
      {:ex_aws_s3, "~> 2.1.0"},
      {:ex_aws_sts, "~> 2.1.0"},
      {:ex_aws_sqs, "~> 3.3.1"},
```

```
      {:hackney, "~> 1.17.4"},
```

### Current behavior

I am using Broadway's SQS module to listen to a queue. Every once and a while, since June 8th mid day EST, several of our environments have seen the VM crash after emitting many of the following messages very quickly:

```
GenServer S3Events.Consumer.Broadway.Producer_0 terminating
** (BadMapError) expected a map, got: {:error, {:http_error, 400, %{code: "Throttling", detail: "", message: "Rate exceeded", request_id: "a3c1cfc8-20ef-4579-94a3-120efef98375", type: "Sender"}}}
    (elixir 1.11.4) lib/map.ex:422: Map.take/2
    (ex_aws 2.2.1) lib/ex_aws/config.ex:91: ExAws.Config.retrieve_runtime_value/2
    (elixir 1.11.4) lib/stream.ex:572: anonymous fn/4 in Stream.map/2
    (elixir 1.11.4) lib/enum.ex:3776: Enumerable.List.reduce/3
    (elixir 1.11.4) lib/stream.ex:1609: Enumerable.Stream.do_each/4
    (elixir 1.11.4) lib/enum.ex:974: Enum.find/3
    (ex_aws 2.2.1) lib/ex_aws/config.ex:71: anonymous fn/2 in ExAws.Config.retrieve_runtime_config/1
    (stdlib 3.14.2) maps.erl:233: :maps.fold_1/3
Last message: {:"$gen_producer", {#PID<0.18472.0>, #Reference<0.4239996247.2497708033.129821>}, {:ask, 10}}
```

the relevant ex_aws config is 
```
config :ex_aws,
  secret_access_key: [{:awscli, "default", 30}],
  access_key_id: [{:awscli, "default", 30}],
  region: {:system, "AWS_REGION"},
  awscli_auth_adapter: ExAws.STS.AuthCache.AssumeRoleWebIdentityAdapter,
  role_session_name: "ex_aws_sts"
```

also, the VM is running in a kubernetes cluster, and the container is set up to use "service account tokens" in the way AWS advises. something like this, in the k8s config

```
    - name: AWS_WEB_IDENTITY_TOKEN_FILE
      value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
...
    volumeMounts:
    - mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
      name: aws-iam-token
      readOnly: true
...
  volumes:
  - name: aws-iam-token
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: sts.amazonaws.com
          expirationSeconds: 86400
          path: token
  - name: token-vol
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3600
          path: token
```

juanazam:
Closes #798 

As described on the issue, when an error happens when retrieving credentials, the error itself is stored on the auth cache, which results in weird errors happening on subsequent calls.

In order to not make extensive changes but still make the library more resilient, a simple retry algorithm is introduced:

* 30 second window till next credentials refresh is divided into 6 blocks of 5 seconds.
* On each block, a node will wait from 0 to 5 seconds till it retries the auth request
* If after the threshold nodes haven't been able to refresh the credentials successfully, an error is raised

This implementation doesn't include any breaking changes to the current API.

We were experiencing the issue happening on #798 every other day, and after running this patch, we haven't seen it happen once.

@bernardd I'm tagging you because you participated on the discussion on the issue and you have (at least some) context

Thanks!

### DIFF
--- a/lib/ex_aws/config/auth_cache.ex
+++ b/lib/ex_aws/config/auth_cache.ex
@@ -81,21 +81,18 @@ defmodule ExAws.Config.AuthCache do
 
   defp attempt_credentials_refresh(adapter, auth, profile, expiration, retries \\ 6) do
     case adapter.adapt_auth_config(auth, profile, expiration) do
-      {:error, error} ->
-        if retries == 1 do
-          Process.send_after(self(), {:refresh_awscli_config, profile, expiration}, expiration)
+      {:error, error} when retries == 1 ->
+        Process.send_after(self(), {:refresh_awscli_config, profile, expiration}, expiration)
 
-          raise "Could't get credentials from auth adapter after 6 retries, last error was #{inspect error}"
-        else
-          Process.sleep(:rand.uniform(5_000))
-          attempt_credentials_refresh(adapter, auth, profile, expiration, retries - 1)
-        end
+        raise "Could't get credentials from auth adapter after 6 retries, last error was #{inspect(error)}"
+
+      {:error, _error} ->
+        Process.sleep(:rand.uniform(5_000))
+        attempt_credentials_refresh(adapter, auth, profile, expiration, retries - 1)
 
       # Always store a map on AuthCache
       auth when is_map(auth) ->
         auth
-
-      _unexpected -> %{}
     end
   end
 


lib/ex_aws/config/auth_cache.ex
```diff
@@ -72,14 +70,35 @@ defmodule ExAws.Config.AuthCache do
...
+    Process.send_after(self(), {:refresh_awscli_config, profile, expiration}, expiration)
     :ets.insert(ets, {{:awscli, profile}, auth})
 
     auth
   end
 
+  defp attempt_credentials_refresh(adapter, auth, profile, expiration, retries \\ 6) do
+    case adapter.adapt_auth_config(auth, profile, expiration) do
+      {:error, error} ->
+        if retries == 1 do
```
bernardd, line 85: Rather than a separate `if` block, you could check this condition in the `case` by doing
```
{:error, error} when retries == 1 ->
    ...do stuff
{:error, _error} ->
    do sleep and retry

File: lib/ex_aws/config/auth_cache.ex [hunk 2]
```diff
@@ -72,14 +70,35 @@ defmodule ExAws.Config.AuthCache do
...
+        else
+          Process.sleep(:rand.uniform(5_000))
+          attempt_credentials_refresh(adapter, auth, profile, expiration, retries - 1)
+        end
+
+      # Always store a map on AuthCache
+      auth when is_map(auth) ->
+        auth
+
+      _unexpected -> %{}
```
bernardd, line 98: Should we ever hit this? If so, in what case?
If it's not a thing we should ever hit, and we don't know what to do if we do, then I think it's sensible to leave out this block and let it crash, since otherwise we're leaving the system in an undefined state.

bernardd:
Hi @juanazam - thanks for the PR!

It looks good, I'm just a little concerned that it introduces a non-deterministic bit of behaviour in terms of its retry interval. What's the reasoning behind randomising the delay between each try?

I think I'd also like a config option for the number of retries - I'm sure 6 is a sensible default, but it's the kind of thing you can imagine people might like to have fail faster or retry more on.

Also, if you could give it a quick `mix format` so that it passes CI, that would be great, thanks :)

juanazam:
@bernardd Thanks for your review

The reason for randomizing the delays is because the main reason we will see an error here is because of Throttling. When there are several nodes trying to get fresh credentials, Amazon has a 4 per second limit. Lets say 10 node crash, and both 10 nodes try to get fresh credentials at the same time, most of them will hit the 4 per second threshold. By introducing randomization as jitter, we are reducing the chances of requests colliding on the same second, resulting in less changes of nodes getting throttling issues.

We chose 6 retries, because we have 30 seconds for the next credentials refresh, and that's why we randomize the wait to 5 seconds top (6 * 5 = 30).

Making that a configuration would make this logic harder, and given right now we are just storing an error tuple on the cache, I think this may be already a win (more resiliency and no updates on the current API).

I addressed your comments are run `mix format`, thanks again 🙏

bernardd:
Okay, that makes sense. It's not necessarily an ideal solution, but you're right that it's still an improvement from the current behaviour. Thanks very much for your work on it :)

juanazam:
@bernardd Thanks for taking the time to take a look and maintaining this project, it's used by many :)

PR #893, opened 2022-07-06
Remove warnings in test environment

prodis:
It removes the following warnings when running the tests:

```
warning: use Mix.Config is deprecated. Use the Config module instead
  config/config.exs:1

warning: use Mix.Config is deprecated. Use the Config module instead
  config/test.exs:1

warning: variable "opts" is unused (if the variable is not meant to be used, prefix it with an underscore)
  test/ex_aws/instance_meta_test.exs:80: ExAws.InstanceMetaTest."test IMDSv2 when initial metadata request fails with a 401, fallback to IMDSv2 flow"/1

warning: variable "opts" is unused (if the variable is not meant to be used, prefix it with an underscore)
  test/ex_aws/instance_meta_test.exs:104: ExAws.InstanceMetaTest."test IMDSv2 configuration to use IMDSv2 by default"/1
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,4 +1,4 @@
-use Mix.Config
+import Config
 
 config :logger, :console,
   level: :debug,
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,4 +1,4 @@
-use Mix.Config
+import Config
 
 config :ex_aws,
   debug_requests: true,
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -1,1 +1,1 @@
-use Mix.Config
+import Config
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,4 +1,4 @@
-use Mix.Config
+import Config
 
 config :logger, level: :warn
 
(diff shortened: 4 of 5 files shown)


bernardd:
Nice one, thanks @prodis! Those have been bugging me for a while, I just never had a moment to fix them.

bernardd:
Hrm. Those CI errors are probably not your fault. I'm gonna have to look at them later when I have a moment.

bernardd:
Oh right. the issue is that the `Config` module didn't actually exist prior to 1.10, and ExAws still officially supports back to 1.7. There are probably magic macros you can use to work around that, and at some point in the not too distant future I'll probably drop support for them, but for now that change does break support for older versions.

prodis:
I see. An alternative is to check the Elixir version in the config files and use the compatible approach.

@bernardd Let me know if you agree and I can do it.

bernardd:
Hey, sorry for the delay - minor case of Covid over here. Yes, that would be fine too if you're happy to go with that solution. Cheers.

bernardd:
I think this is resolved now that I've bumped the minimum version to 1.10. Thanks for the contribution @prodis.

PR #900, opened 2022-08-30
Add EKS support

afatsini:
Adds support to EKS, allowing to create operations like:

```
data = %{scalingConfig: %{desiredSize: replicas}}

 operation = %ExAws.Operation.JSON{
      http_method: :post,
      path: "/clusters/#{cluster()}/node-groups/#{node_group()}/update-config",
      data: data,
      service: :eks
    }
```

### DIFF
--- a/priv/endpoints.exs
+++ b/priv/endpoints.exs
@@ -1182,6 +1182,27 @@
             "us-west-2" => %{}
           }
         },
+        "eks" => %{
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @afatsini ! I don't suppose you feel like writing and maintaining an ExAwsEks module while you're at it? ;)

PR #904, opened 2022-09-11
Force refresh of config auth

### LINKED ISSUE #824: Got `ExpiredToken` error when `ExAws.S3.list_objects("my-bucket") |> ExAws.stream!` runs for a long time
### Environment

  ```
  Erlang/OTP 24 [erts-12.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]

  Elixir 1.12.3 (compiled with Erlang/OTP 24)
  ```
  ```
  * ex_aws (Hex package) (mix)
    locked at 2.2.3 (ex_aws) 8ddf1da9
  * ex_aws_s3 (Hex package) (mix)
    locked at 2.3.0 (ex_aws_s3) 0b13b114
  * ex_aws_sts (Hex package) (mix)
    locked at 2.2.0 (ex_aws_sts) 68a75265
  ```
  ```
  * hackney (Hex package) (rebar3)
    locked at 1.18.0 (hackney) 9afcda62
  ```

### Current behavior

Currently, `ExAws.stream!` would pass a new `config` struct to the lazy function,
https://github.com/ex-aws/ex_aws/blob/87ec5641e53983fc627918a336d1a2c310489b70/lib/ex_aws.ex#L105
when combined with `list_objects`, this `config` (and the `token` inside this `config`) would be reused to fetch all the pages,
but **lazily**

If we pass the `list_objects` stream to a long running `Task.async_stream`, the token might expire when fetching a new page:
```elixir
ExAws.list_objects(...)
|> ExAws.stream!
|> Task.async_stream(fn object -> Process.sleep(24h) end)
```

And the error would be:
```
** (ExAws.Error) ExAws Request Error!

{:error, {:http_error, 400, %{body: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>ExpiredToken</Code><Message>The provided token has expired.</Message><Token-0>...</Token-0><RequestId>...</RequestId><HostId>...</HostId></Error>", headers: [...], status_code: 400}}}

    (ex_aws 2.2.3) lib/ex_aws.ex:87: ExAws.request!/2
    (ex_aws_s3 2.3.0) lib/ex_aws/s3/lazy.ex:7: anonymous fn/4 in ExAws.S3.Lazy.stream_objects!/3
    (ex_aws_s3 2.3.0) lib/ex_aws/s3/lazy.ex:18: anonymous fn/2 in ExAws.S3.Lazy.stream_objects!/3
    (elixir 1.12.3) lib/stream.ex:1531: Stream.do_resource/5
    (elixir 1.12.3) lib/stream.ex:1719: Enumerable.Stream.do_each/4
    (elixir 1.12.3) lib/task/supervised.ex:336: Task.Supervised.stream_reduce/7
    (elixir 1.12.3) lib/stream.ex:649: Stream.run/1
```

### Expected behavior

No `ExpiredToken` error would be raised, no matter how much time the stream needs to be processed.

I'd propose we fetch a new token every time we request a new page from S3.

cjbottaro:
## Problem

When ExAws streams an operation, it calculates a config map once, then passes that config as "overrides" for each request during the stream. This means that auth creds are resolved _once_ and used for the entire duration of the stream. If you have a stream that takes longer than 6 hours on an EC2 instance, you'll eventually get "token expired" errors.

## Fix

This PR detects refreshable creds in a config map, and uses `ExAws.Config.AuthCache` to resolve the creds on every request, even if an exist config map is passed as overrides.

This fix is completely transparent to downstream stream builders.

Fixes #824

## Testing

I tested locally by printing to stdout, but I have no idea how to unit test it. I'm currently testing in production to see if it's ultimately effective.

## How it works

Currently two types of configuration values (`:instance_role` and `:awscli`) use `ExAws.Config.AuthCache`, and thus have refreshable content.

The config map is augmented with a new field `:refreshable` which can be `false` or a list of refreshable values. For example:
```elixir
iex> config = ExAws.Config.new(:s3)
%{
  ...
  refreshable: [:awscli],
  ...
}
```

Now on subsequent calls, passing that config in as overrides, `new/2` knows to remove certain fields from the overrides which needs to be refreshed.

```elixir
iex> ExAws.Config.new(:s3, config)
%{
  ...
  refreshable: [:awscli],
  access_key_id: ..., # Potentially different
  secret_access_key: ..., # Potentially different
  security_token: ..., # Potentially different
  ...
}
```

Because `config` has `refreshable: [:awscli]`, we know we can delete things like `:access_key_id`, `:secret_access_key`, and `:security_token` from the overrides and let them be recalculated.

You can opt out of this behavior with...
```elixir
iex> config = ExAws.Config.new(:s3, refreshable: false)
%{
  ...
  refreshable: false,
  ...
}
```

But I don't know why you would want to.

### DIFF
--- a/lib/ex_aws/config.ex
+++ b/lib/ex_aws/config.ex
@@ -24,6 +24,23 @@ defmodule ExAws.Config do
     :telemetry_options
   ]
 
+  @instance_role_config [
+    :access_key_id,
+    :secret_access_key,
+    :security_token
+  ]
+
+  @awscli_config [
+    :source_profile,
+    :role_arn,
+    :access_key_id,
+    :secret_access_key,
+    :region,
+    :security_token,
+    :role_session_name,
+    :external_id
+  ]
+
   @type t :: %{} | Keyword.t()
 
   @doc """
@@ -69,10 +86,41 @@ defmodule ExAws.Config do
 
     defaults = ExAws.Config.Defaults.get(service, region)
 
-    defaults
+    config = defaults
     |> Map.merge(common_config)
     |> Map.merge(service_config)
-    |> Map.merge(overrides)
+    |> add_refreshable_metadata()
+
+    # (Maybe) do not allow overrides for refreshable config.
+    overrides = if refreshable = overrides[:refreshable] do
+      Enum.reduce(refreshable, overrides, fn
+        :awscli, overrides -> Map.drop(overrides, @awscli_config)
+        :instance_role, overrides -> Map.drop(overrides, @instance_role_config)
+      end)
+    else
+      overrides
+    end
+
+    Map.merge(config, overrides)
+  end
+
+  # :awscli and :instance_role both read creds from ExAws.Config.AuthCache which
+  # which is "refreshable". This is useful for long running streams where the
+  # creds can change while the stream is still running.
+  defp add_refreshable_metadata(config) do
+    refreshable = Enum.flat_map(config, fn {_k, v} -> List.wrap(v) end)
+    |> Enum.reduce([], fn
+      {:awscli, _, _}, acc -> [:awscli | acc]
+      :role_instance, acc -> [:role_instance | acc]
+      _, acc -> acc
+    end)
+    |> Enum.uniq()
+
+    if refreshable != [] do
+      Map.put(config, :refreshable, refreshable)
+    else
+      config
+    end
   end
 
   def retrieve_runtime_config(config) do
@@ -95,6 +143,9 @@ defmodule ExAws.Config do
       {:headers, headers}, config ->
         Map.put(config, :headers, headers)
 
+      {:refreshable, refreshable}, config ->
+        Map.put(config, :refreshable, refreshable)
+
       {k, v}, config ->
         case retrieve_runtime_value(v, config) do
           %{} = result -> Map.merge(config, result)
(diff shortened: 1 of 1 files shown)


bernardd:
Aw dang, I just notice that this had slipped through the cracks - really sorry I haven't gotten to it sooner. Rest assured it is (now) on my list.

bernardd:
This looks good @cjbottaro. Would you be able to add a note to the docs about setting `refreshable: false`, just so that people know they can override it if they need to? Then we'll get it merged. Cheers.

cjbottaro:
Oh shoot. Ha, I lost track of this also. Yeah, I'll try to write up some docs.

cjbottaro:
@bernardd I added the docs in 7a23314. Let me know if that looks good, it's been a while since I thought about this; I kinda forgot what's going on and how it works a little... 😂 On the up side, we've been using this branch in prod since Sept of last year!

bernardd:
Thanks @cjbottaro! Would you be able to give it a quick `mix format` pass, please? Then I'll get it merged.

aglundahl:
@bernardd Do you want me to close #924? At least I won't need it, if we're moving forward with this PR.

Getting rid of the premature config resolution might be nice for other reasons, though, if you think it's worth a potentially breaking change.

bernardd:
Thanks - yeah, let's close #924 for now. It's not ideal, but I'd rather avoid a breaking change if we can.

PR #906, opened 2022-09-20
Add ca-central-1 region to cognito-idp service

okanatas:
This commit authenticates Amazon Cognito users requesting access from Canada(Central) region.


bernardd:
Thanks @okanatas!

PR #907, opened 2022-10-01
Fix unused opts variables in tests

kianmeng:
Resolve below warning:

    warning: variable "opts" is unused (if the variable is not meant to
    be used, prefix it with an underscore)


kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #910, opened 2022-10-18
Adds missing region for Rekognition in priv/endpoints.exs

silvagustin:
Problem
---------
I came across with the following error when I had tried to run `AWS Rekognition` for an image stored in S3 in London region (eu-west-2):
Screen Shot 2022-10-18 at 12 45 32

Fix
---
This PR adds missing `eu-west-2` region for Rekognition in 'priv/endpoints.exs file.

All service endpoints availables can be read in [Amazon Rekognition endpoints and quotas](https://docs.aws.amazon.com/general/latest/gr/rekognition.html) page.

### DIFF
--- a/priv/endpoints.exs
+++ b/priv/endpoints.exs
@@ -1313,6 +1313,7 @@
             "ca-central-1" => %{},
             "eu-central-1" => %{},
             "eu-west-1" => %{},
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @silvagustin! Apologies for the delay getting to this.

PR #915, opened 2022-11-16
Add Jakarta region in partitions

aakashjhawar:
Add support for Jakarta region.


bernardd:
Thanks @aakashjhawar!

PR #921, opened 2022-12-18
Increase AWS -related auth. compatibility

paulo-ferraz-oliveira:
I'm using [`localstack`](https://github.com/localstack/localstack) to perform local tests on top of `ex_aws`.

I'm getting the following error 

```
"Authorization header requires \\Signature\\ parameter. Authorization header requires \\SignedHeaders\\ parameter.
```

as part of HTTP calls. I've run the request with `debug_requests: true` and found that the `Authorization` header does have those parameters, but... they're written like this

```
AWS4-HMAC-SHA256 Credential=<cred>,SignedHeaders=<sign_head>,Signature=<sign>
```

instead of

```
AWS4-HMAC-SHA256 Credential=<cred>, SignedHeaders=<sign_head>, Signature=<sign>
                                   ^ notice the extra space
```

This pull request adds the extra space, which is what's issued by the official AWS client, in any case.

I also add LocalStack to CI for future regression testing purposes (if CI isn't running - because this is my first contribution to the repo., you can check results here: https://github.com/paulo-ferraz-oliveira/ex_aws/actions/runs/3726712502/jobs/6320334886).

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -22,6 +22,10 @@ jobs:
         with:
           port: 8000
           cors: "*"
+      - name: LocalStack - Kinesis
+        run: |
+          docker run --detach --publish 4567:4566 localstack/localstack
+          until curl -s http://localhost:4567; do sleep 0.5; done
       - run: mix deps.get
       - run: mix compile
       - run: mix format --check-formatted
--- a/config/test.exs
+++ b/config/test.exs
@@ -35,3 +35,9 @@ config :ex_aws, :s3,
   scheme: "https://",
   host: "s3.amazonaws.com",
   region: "us-east-1"
+
+config :ex_aws, :kinesis,
+  scheme: "http://",
+  host: "localhost",
+  port: 4567,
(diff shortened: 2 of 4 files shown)


paulo-ferraz-oliveira:
I've pushed a commit to `paulo-ferraz-oliveira/ex_aws` where we can see the old code fail: https://github.com/paulo-ferraz-oliveira/ex_aws/actions/runs/3726731391/jobs/6320365816#step:7:146

paulo-ferraz-oliveira:
I'm now reconsidering if this ("fixing" `ex_aws`) is the best approach to the problem explained above (since we'd be adapting the client to the server behaviour, but that one might not be fully compliant with expectations) so would be Ok to have the pull request closed without merge.

References:
* localstack/localstack#7358
* etspaceman/kinesis-mock#387

paulo-ferraz-oliveira:
The issue, as presented before, was found to be in LocalStack's default imported `kinesis-mock`. The latter is now fixed, while the former allows importing a specific version of the latter. All good!

PR #922, opened 2022-12-20
Add eu-west-2 to sagemaker endpoints

karolinelende:
Added region `eu-west-2` to SageMaker endpoints so we can make use of [ex_aws_sagemaker_runtime](https://github.com/dideler/ex_aws_sagemaker_runtime) in this region.


bernardd:
Awesome, thanks heaps @karolinelende

PR #923, opened 2022-12-21
Fix happy path-only http client doc

adworse:
Thank you for all your hard work! Below is a slight docs improvement.

A current example of custom `ExAws.Request.HttpClient` behaviour implementation always packs the response in `{:ok, result}` tuple, which only works for happy path and breaks if the underlying client returns an error.

The proposed example correctly packs errors in the error tuple that ExAws.Request.request_and_retry/7 function is expecting.


bernardd:
Great, thanks very much @adworse

PR #931, opened 2023-02-06
Add eu-west-2 to firehose endpoints

ashutoshrishi:
Firehose's eu-west-2 endpoint is not listed as active in the endpoints config. Therefore you get the error:

```
** (RuntimeError) firehose not supported in region eu-west-2 for partition aws
```

Checking the endpoints available for firehose, `eu-west-2` is listed as an active endpoint:
https://docs.aws.amazon.com/general/latest/gr/fh.html


bernardd:
Thanks @ashutoshrishi!

PR #947, opened 2023-04-18
Add support for IVS endpoints

yashh:
Just supporting [IVS as authorized](https://docs.aws.amazon.com/ivs/latest/APIReference/API_CreateChannel.html) AWS service.


bernardd:
Thanks @yashh!

PR #955, opened 2023-06-16
Add `me-south-1` to SNS endpoints

infinityfye:
Add `me-south-1` to the list of SNS endpoints so ex_aws won't error when connecting to valid sns endpoints.

[x] Run `mix test` to ensure no tests have broken (also please make sure you've added tests for your particular change, where appropriate).

I couldn't find tests that check this particular use case. I would be more than happy to add any if required.


bernardd:
Thanks @infinityfye!

PR #959, opened 2023-06-26
Bump elixir version

wkirschbaum:
Since Elixir 1.15 is out, 1.10 is not supported anymore.  This also
allows us to get rid of deprecation warnings.

https://hexdocs.pm/elixir/1.15.0/compatibility-and-deprecations.html


wkirschbaum:
It might also not be productive to bump the minimum version unless there is an actual reason. The only requirement from my knowledge is using the Req library in another PR, but Finch is fine with 1.7 afaik.

bernardd:
Ah dang, I totally forgot you'd already done this and I made a bunch of similar changes. Not nearly as thorough as yours, though. Let me get this merged. Thanks!

bernardd:
Thanks again @wkirschbaum - I've rebased this and merged it in #975

PR #960, opened 2023-06-26
Add Req compatibility and replace Httpotion as an example

wkirschbaum:
Httpotion has been retired and this introduces a new alternative.  To make compatibility easier there is also a new map_response function.

This fixes: https://github.com/ex-aws/ex_aws/issues/958


wkirschbaum:
The breaking build should be fixed by: https://github.com/ex-aws/ex_aws/pull/959

bernardd:
Thanks @wkirschbaum! I've rebased this and merged it in #976

PR #961, opened 2023-06-30
Add eu-central-2 to SQS endpoints

shevchenko-yegor:
Adds SQS endpoint for eu-central-2 Europe (Zurich)


bernardd:
Thanks @shevchenko-yegor!

PR #972, opened 2023-08-12
Add eu-south-1 to SQS endpoints

biserz:
Adding the eu-south-1 (Milan) region to the list of endpoints.


bernardd:
Thanks @biserz!

PR #975, opened 2023-08-15
Bump min elixir version

bernardd:
Since Elixir 1.15 is out, 1.10 is not supported anymore.  This also allows us to get rid of deprecation warnings.

https://hexdocs.pm/elixir/1.15.0/compatibility-and-deprecations.html


bernardd:
Replaces #959

PR #976, opened 2023-08-15
Add Req compatibility and replace Httpotion as an example

bernardd:
Httpotion has been retired and this introduces a new alternative.  To make compatibility easier there is also a new map_response function.

* lib/ex_aws/request.ex (map_response): New method.
(do_request): Add map_response call after http_client.request.
* lib/ex_aws/request.ex: Update documentation.
* mix.exs: Add Req and remove Httpotion.
* test/alternate_helper.exs: Replace Httpotion.

PR #978, opened 2023-08-16
Ensure that response is transformed with all requests

metabren:
### Issue

Thanks for the previous work to support Req in #976. Unfortunately it didn't _quite_ go far enough. There are a few more places that `.request` is called on the `http_client`, and they all need to have their responses transformed in case it's a Req response. 

### Necessary bits

- [x] Run `mix format` using a recent version of Elixir
- [x] Run `mix dialyzer` to make sure the typing is correct
- [x] Run `mix test` to ensure no tests have broken (also please make sure you've added tests for your particular change, where appropriate).


bernardd:
Brilliant, thanks very much @metabren (and apologies for taking so long to get to this).

PR #995, opened 2023-09-24
Improve docs regarding Req

a8t:
There was a recent change to replace the `HTTPotion` example with `Req`, but one reference to `HTTPotion` was accidentally left in.

I changed that reference, but I also clarified the docs a tiny bit.

Thanks for the library!


a8t:
actually, this still isn't good. Req uses `status`, not `status_code`. I will update again

bernardd:
Nice one, thanks very much @a8t!

PR #1006, opened 2023-10-24
Update request logic to handle error reasons in keyword lists

maciejgryka:
Fixes https://github.com/ex-aws/ex_aws/issues/1005 by updating `request.ex` to also handle errors in the form of `[reason: reason]` in addition to `%{reason: reason}`.

[x] Run `mix test` to ensure no tests have broken (also please make sure you've added tests for your particular change, where appropriate).


bernardd:
Great, thanks very much @maciejgryka. Apologies for the delay in getting to this.

PR #1009, opened 2023-10-30
Add support for Personalize

nkezhaya:
This allows `ExAws.Config.new(:personalize)` to run without throwing an error.


bernardd:
Thanks @nkezhaya! Apologies for the delay in getting to this.

PR #1015, opened 2023-11-21
update Req HttpClient example

woylie:
This updates the Req example in the `ExAws.Request.HttpClient` documentation.

- Set `decode_body: false` to prevent Req from decoding JSON responses on its own (behaviour expects binary body).
- Rename `status` to `status_code` in response to conform to behaviour. I know ex_aws was updated to map the response on its own in the main branch, but the behaviour still prescribes a `status_code` field.
- Wrapped error into reason map as prescribed by behaviour.
- Added missing `@impl`.
- Added `moduledoc` to example for easier copy-and-paste into projects that enforce module docs with linter rules.
- Updated headers to avoid duplicate links in the sidebar.

### DIFF
--- a/lib/ex_aws/request/http_client.ex
+++ b/lib/ex_aws/request/http_client.ex
@@ -7,7 +7,7 @@ defmodule ExAws.Request.HttpClient do
 
   The default is `:hackney`.
 
-  ## Example
+  ## Example: Req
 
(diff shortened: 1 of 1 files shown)


AmineFrioua (approved):

bernardd:
Thanks @woylie! Apologies for the delay in getting to this.

PR #1017, opened 2023-12-01
Fix s3 gov-us-east-1 and gov-us-west-1 URLs.

mmzx:
Govcloud uses FIPS endpoint. The URL references are here according to AWS docs:

https://aws.amazon.com/compliance/fips/

Screenshot 2023-12-01 at 15 45 09


bernardd:
Thanks @mmzx!

PR #1022, opened 2023-12-21
Add `me-south-1` to s3 endpoints

infinityfye:
`me-south-1` is supported in the `aws` partition as can be seen here: https://docs.aws.amazon.com/general/latest/gr/s3.html

* Run `mix format` using a recent version of Elixir (check)
* Run `mix dialyzer` to make sure the typing is correct (check)
* Run `mix test` to ensure no tests have broken (also please make sure you've added tests for your particular change, where appropriate). (check)


bernardd:
Thanks @infinityfye!

PR #1026, opened 2024-01-03
Fix Req example HttpClient implementation for Req 0.4+

reisub:
This builds on work in #1015 adding a bit of code to adjust the header format to the expected list of 2-element tuples.

## Context

Req 0.4+ changed the headers format to be a map where the key is a string and the value is a list of strings, e.g.:

```elixir
%{
  "content-length" => ["12345"]
}
```

This req change makes the current example fail whenever you try to perform a streaming operation because the code in `ExAws.S3.Download` assumes a single value:

```elixir
defp get_file_size(bucket, path, config) do
  %{headers: headers} = ExAws.S3.head_object(bucket, path) |> ExAws.request!(config)

  headers
  |> Enum.find(fn {k, _} -> String.downcase(k) == "content-length" end)
  |> elem(1)
  |> String.to_integer() # <- fails here because it gets a list
end
```

With the proposed change, the example implementation should handle both new and old Req versions; though I've only tested with a recent one.

### DIFF
--- a/lib/ex_aws/request/http_client.ex
+++ b/lib/ex_aws/request/http_client.ex
@@ -22,22 +22,25 @@ defmodule ExAws.Request.HttpClient do
 
     @impl ExAws.Request.HttpClient
     def request(method, url, body, headers, _http_opts) do
-      case Req.request(
-           method: method,
-           url: url,
-           body: body,
-           headers: headers,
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @reisub!

PR #1043, opened 2024-03-23
add changelog to package files

woylie:
This adds the changelog to the package files, so that it is visible in hex diff.


bernardd:
Nice one, thanks @woylie

PR #1053, opened 2024-05-09
allow iodata body in s3 requests

ruslandoga:
👋 

This PR makes iodata ops possible in ExAws.S3.


bernardd:
Nice, thanks @ruslandoga, and apologies for the delay getting to this.

ruslandoga:
Thank you!

ruslandoga:
I missed https://github.com/ex-aws/ex_aws/blob/8020fa1c0a7221c3931a91ec8089e52c868667c0/lib/ex_aws/request.ex#L16

Is there a way to allow iodata in that case statement?

PR #1055, opened 2024-05-09
upgrade Elixir and Erlang to latest versions, add Elixir 1.16 to CI

ruslandoga:


.tool-versions
```diff
@@ -1,2 +1,2 @@
-elixir 1.15
-erlang 26.1.2
+elixir 1.16.2-otp-26
+erlang 26.2.5
```
ruslandoga, line 2: I wonder if there is a way to make it less restrictive?

bernardd: The point of .tools-versions is to get everyone who's working on this directly using the same version - it's deliberately highly prescriptive so that, for example, we're not all using different versions to run `mix format` :)

bernardd:
Thanks @ruslandoga - this is great. Could you move the `lint` tag in the workflow onto the new pair so that the CI formatting version check matches the default `.tool-versions` one? Then we'll get this merged. Sorry for taking a while to get to it.

bernardd:
Oh, nevermind, turns out I can edit PRs myself :) Thanks :)

PR #1058, opened 2024-06-03
Add regions and endpoints for the Pinpoint `mobiletargeting` service

marramgrass:
Endpoints and regions as listed at https://docs.aws.amazon.com/general/latest/gr/pinpoint.html.

---


bernardd:
Thanks @marramgrass!

PR #1061, opened 2024-06-04
move json encoding to json operation

ruslandoga:
Continues #1053

### DIFF
--- a/lib/ex_aws/operation/json.ex
+++ b/lib/ex_aws/operation/json.ex
@@ -50,7 +50,7 @@ defimpl ExAws.Operation, for: ExAws.Operation.JSON do
     ExAws.Request.request(
       operation.http_method,
       url,
-      operation.data,
+      config[:json_codec].encode!(operation.data),
       headers,
       config,
       operation.service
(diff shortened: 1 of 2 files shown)


lib/ex_aws/request.ex
```diff
@@ -9,14 +9,7 @@ defmodule ExAws.Request do
   @type error_t :: {:error, {:http_error, http_status, binary}}
   @type response_t :: success_t | error_t
 
-  def request(http_method, url, data, headers, config, service) do
-    body =
-      case data do
-        [] -> "{}"
```
ruslandoga, line 15: When is `[]` possible? Does it represent `nil`? Otherwise an empty list and an empty object are not equivalent in JSON.

lib/ex_aws/operation/json.ex
```diff
@@ -50,7 +50,7 @@ defimpl ExAws.Operation, for: ExAws.Operation.JSON do
     ExAws.Request.request(
       operation.http_method,
       url,
-      operation.data,
+      config[:json_codec].encode!(operation.data),
```
ruslandoga, line 53: Here `[]` would be encoded as `[]` and not `{}`

PR #1069, opened 2024-06-29
Adds ap-south-2 in endpoints.exs

alpeshjamgade:
Adds support for ap-south-2 (Hyderabad) region


bernardd:
Thanks @alpeshjamgade!

PR #1094, opened 2024-10-25
Configurable max attempts for client errors

studzien:
We want to be able to apply different retry strategies depending on the error type we get back from AWS.
More specifically:
  1. We don't want to retry throughput exceeded errors to fail fast in case we temporarily overloaded AWS and bubble up the backpressure to the caller.
  2. If AWS has some internal problems and returns 500 errors for a few seconds, we want to retry them.

This PR proposes a naive approach to enabling different `max_attempts` to be applied to the client errors (`400..499`).

A more sophisticated approach is to override the whole retry strategy (including backoff values) for specific errors or error ranges.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -53,7 +53,7 @@ defmodule ExAws.Request do
                 config,
                 headers,
                 req_body,
-                attempt_again?(attempt, reason, config)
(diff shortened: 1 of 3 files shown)


bernardd:
Looks good, thanks @studzien!

PR #1104, opened 2024-12-16
Add Bedrock Agent endpoints

anukaver-glia:
Map the Bedrock Agent and Bedrock Agent Runtime endpoints. This allows to start using the services with the `ex_aws` library.

The PR is identical to the one which introduced endpoints for Bedrock and Bedrock Runtime: https://github.com/ex-aws/ex_aws/pull/1023.

The list of supported endpoints relies on
https://docs.aws.amazon.com/general/latest/gr/bedrock.html#bedrock_region.

### DIFF
--- a/priv/endpoints.exs
+++ b/priv/endpoints.exs
@@ -156,6 +156,40 @@ chime_voice_regions = [
             "us-west-2" => %{}
           }
         },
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @anukaver-glia!

PR #1105, opened 2024-12-21
Feat/add tooling to generate endpoints

djgoku:
Adding an Elixir script to download the latest endpoints.json file from https://github.com/boto/botocore/blob/develop/botocore/data/endpoints.json, save it to `priv/endpoints.exs`, and then `mix format` it.


feld:
This would be very helpful. I've run into two services now that didn't have the partitions included in ex_aws so I've had to hack around it.

jasonlatta:
Ditto Feld's comment. Created https://github.com/ex-aws/ex_aws/pull/1147 due to missing services in partitions. But hadn;t realized there was some kind of "official" file this was replicating. So, this PR would be a better solution

Issue #1110, opened 2025-01-13
redirect vs follow_redirect vs follow_redirects


inoas-nbw:
req < 0.4.0 expects follow_redirects
req >= 0.4.0 expects redirect
hackney option may be follow_redirect

https://github.com/ex-aws/ex_aws/blob/v2.5.7/lib/ex_aws/request/req.ex#L39
https://github.com/ex-aws/ex_aws/blob/v2.5.8/lib/ex_aws/request/req.ex#L34
https://hexdocs.pm/req/0.3.12/Req.Steps.html#follow_redirects/1
https://hexdocs.pm/req/0.5.8/Req.Steps.html#redirect/1

=> as req is not stable and ex_aws does not specify req above 0.4, it should set both, follow_redirects and redirect.
also the current comment is missleading https://github.com/ex-aws/ex_aws/blob/v2.5.8/lib/ex_aws/request/req.ex#L34

PR #1111, opened 2025-01-13
req-docs

inoas-nbw:


bernardd:
Thanks @inoas-nbw.
Resolves #1110

PR #1127, opened 2025-03-25
Change Req requirement to v0.5.10+

### LINKED ISSUE #1086: Problem with Req HTTP Client
I tried to use the new Req HTTPClient, but it failed me when trying to download from S3 via stream
```
(ex_aws_s3 2.5.4) lib/ex_aws/s3/download.ex:98: ExAws.Operation.ExAws.S3.Download.stream!/2
(ex_aws_s3 2.5.4) lib/ex_aws/s3/download.ex:28: ExAws.S3.Download.build_chunk_stream/2
:erlang.binary_to_integer(["3047599"])
* 1st argument: not a binary
```

Problem is that the headers have a different format for Req.

I fixed the error by adding these lines to the response (I took them from your documentation in HTTP Client module):
``` elixir
headers =
  Enum.flat_map(headers, fn
    {name, vals} when is_list(vals) -> Enum.map(vals, &{name, &1})
    {name, val} -> {name, val}
  end)
```
 
This small change in ExAws.Request.Req fixed all my problems.

Please let me know if you need any more information

I'm happy to do the change myself if you want me to.
In that case, please give me some instructions on how to contribute.

wojtekmach:
Closes #1086

In v0.5.10 I have added `Req.get_headers_list(req_or_resp)` to convert headers to the more common list of tuples format.

Req does retries to GET/HEAD requests by default (I'm thinking about doing no retries by default in Req v1.0 though, https://github.com/wojtekmach/req/issues/383), however since ExAws.Request does retries too, I've explicitly set Req not to retry.


wojtekmach:
In Req v0.5 we had a few breaking changes, https://github.com/wojtekmach/req/blob/main/CHANGELOG.md#full-v050-changelog. If you'd rather not force users to go to v0.5, I'm happy to back port this API to v0.4.x. Let me know!

PR #1128, opened 2025-03-25
Do not warn when optional deps are not available

### LINKED ISSUE #1102: Elixir 1.17 compiler warning: Req is not defined
Elixir 1.17 raises a compiler warning that `Req` is not defined. We don't use Req in our project, and it's an optional dependency, so it should be silenced in this case.

### Environment

Erlang/OTP 27 [erts-15.1.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Elixir 1.17.3 (compiled with Erlang/OTP 27)

```
* ex_aws (Hex package) (mix)
  locked at 2.5.7 (ex_aws) 2c3c5775

* ex_aws_s3 (Hex package) (mix)
  locked at 2.5.5 (ex_aws_s3) ac34af1e
```

```
* httpoison (Hex package) (mix)
  locked at 2.2.1 (httpoison) 51364e6d

* hackney (Hex package) (rebar3)
  locked at 1.20.1 (hackney) fe9094e5
```

### Current behavior

We get this compiler warning:

```
==> ex_aws
Compiling 28 files (.ex)
    warning: Req.request/1 is undefined (module Req is not available or is yet to be defined)
    │
 24 │     |> Req.request()
    │            ~
    │
    └─ lib/ex_aws/request/req.ex:24:12: ExAws.Request.Req.request/5
```

### Expected behavior

Not to get this compiler warning when we're not using Req, as it's an optional dependency.

wojtekmach:
Closes #1102

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -25,6 +25,12 @@ defmodule ExAws.Mixfile do
         "coveralls.detail": :test,
         "coveralls.post": :test,
         "coveralls.html": :test
+      ],
+      xref: [
+        exclude: [
+          :hackney,
+          Req
+        ]
       ]
     ]
   end


bernardd:
Nice one, thanks @wojtekmach

PR #1129, opened 2025-03-27
Update Bedrock endpoints with latest AWS regions

adriencohen:
Updates the AWS Bedrock endpoints (bedrock and bedrock-runtime) with all currently supported regions according to the [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/bedrock.html).


bernardd:
Thanks @adriencohen!

PR #1131, opened 2025-03-28
Implement retries for TooManyRequestsException 

db-qstream:
Currently [retries](https://github.com/ex-aws/ex_aws/blob/main/lib/ex_aws/request.ex#L167) are only done for `ProvisionedThroughputExceededException` and `ThrottlingException`.

A few of the Cognito actions like [AdminCreateUser](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUser.html) and [AdminSetUserPassword](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminSetUserPassword.html) return `TooManyRequestsException` when the request limits are exceeded.

This PR contains fix to retry on `TooManyRequestsException`

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -172,6 +172,10 @@ defmodule ExAws.Request do
     {:retry, {type, message}}
   end
 
+  def handle_aws_error({"TooManyRequestsException" = type, message, _}) do
+    {:retry, {type, message}}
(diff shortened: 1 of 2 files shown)


bernardd:
Nice one, thanks @db-qstream!

PR #1135, opened 2025-05-12
Add stream_builder to RestQuery, fix Operation.stream implementation

lukaszsamson:
This PR adds `stream_builder` property to `RestQuery`. The `stream!` implementation added in https://github.com/ex-aws/ex_aws/commit/6cc2365b24ab71551840924220026e86495710ac could never work as intended. To fix it I added an implementation similar to the one in `Json` https://github.com/ex-aws/ex_aws/blob/main/lib/ex_aws/operation/json.ex#L63

Fixes a type system warning:
```
    warning: the 1st pattern in clause will never match:

        %{stream_builder: fun}

    because it is expected to receive type:

        dynamic(%ExAws.Operation.RestQuery{})

    typing violation found at:
    │
 32 │   def stream!(%{stream_builder: fun}, config) do
    │               ~
    │
    └─ (ex_aws 2.5.8) lib/ex_aws/operation/rest_query.ex:32:15: ExAws.Operation.ExAws.Operation.RestQuery.stream!/2
```

### DIFF
--- a/lib/ex_aws/operation/rest_query.ex
+++ b/lib/ex_aws/operation/rest_query.ex
@@ -1,7 +1,8 @@
 defmodule ExAws.Operation.RestQuery do
   @moduledoc false
 
-  defstruct http_method: nil,
+  defstruct stream_builder: nil,
+            http_method: nil,
             path: "/",
             params: %{},
             body: "",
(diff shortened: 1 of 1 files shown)


bernardd:
Thanks @lukaszsamson!

PR #1140, opened 2025-05-27
Fix build status badge

PragTob:
Seemed broken for me when I checked out the github.

image

with this it looks fine/as it should :)

Here's a small side by side:

image

Thanks for ExAws!


bernardd:
Thanks @PragTob!

PR #1153, opened 2025-07-14
Refresh session tokens

axelson:
Refresh session tokens after they expire in an hour

https://docs.aws.amazon.com/singlesignon/latest/userguide/authconcept.html#sessionsconcept

Based on this doc: https://docs.aws.amazon.com/singlesignon/latest/userguide/authconcept.html#sessionsconcept
Depending on the configured expiration time for the `aws sso login` token which can be up to 90 days, the "application session" token (which I presume is what you're using in the batch process) only receives tokens that are valid for 1 hour (I assume this is a security measure so the session can be revoked without waiting a full 90 days).

So how do you refresh the tokens, you make a `CreateToken` call: https://docs.aws.amazon.com/singlesignon/latest/OIDCAPIReference/API_CreateToken.html

Here's the base url: "https://oidc.#{region}.amazonaws.com/token"

And here's an example of the payload (where the `sso_cache` is from `~/.aws/sso/cache`):
```
payload = %{
  "grantType" => "refresh_token",
  "clientId" => sso_cache["clientId"],
  "clientSecret" => sso_cache["clientSecret"],
  "refreshToken" => sso_cache["refreshToken"]
}
```

TODO:
- Clean up so it can be upstreamed (improve expiration detection)
- Should we be writing the updated token back into the cache?

### DIFF
--- a/lib/ex_aws/credentials_ini/file.ex
+++ b/lib/ex_aws/credentials_ini/file.ex
@@ -46,15 +46,14 @@ if Code.ensure_loaded?(ConfigParser) do
       with {_, {:ok, sso_cache_content}} <-
              {:read, File.read(get_sso_cache_file(sso_cache_key))},
            {_,
-            {:ok, %{"expiresAt" => expires_at, "accessToken" => access_token, "region" => region}}} <-
+            {:ok, %{"expiresAt" => expires_at} = sso_cache}} <-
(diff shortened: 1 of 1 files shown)


axelson:
Whoops, didn't mean to open this one either. I forgot that even if you start on the pull requests tab for _your_ fork. The pull request will by default be created on the upstream repo 🤦

PR #1155, opened 2025-07-25
Try to fix access to S3 object that has question mark in the object name

speeddragon:
During a cleanup of my files (stored in DigitalOcean Spaces, equivalent to S3), I've noticed some issues accessing files with question marks (?) in the object name. E.g.: `test.jpg?id=3` and `test.jpg?`.

I'm not sure if the best solution here would be avoiding storing files with special characters, like mentioned on [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines-avoid-characters), but also found similar issues in the past, like #660 so I wanted to come up a solution for this case, but I think there would still be some other edge cases to explore regarding this.

To solve this, under the `perform/2` function in `ExAws.Operation.S3`, I've encoded the object path. 

To avoid the double encoding, I needed to remove `Url.uri_encode/1` under:
- `ExAws.Auth.signature/8`
- `ExAws.Request.Url.query/1`

I'm not familiar with other AWS services and how they handle the query params, but I believe I also would need to add `uri_encode/1` to the other `ExAws.Operation` modules (`JSON`, `Query`, `RestQuery`).

~Probably exist some other valid cases to not encode every query parameter, I believe `versionId` would be one of them, but I think there should be more.~ (This is fixed in the second commit)

I'm also not sure if DigitalOcean Spaces mimics 100% the same S3 behaviour.

I would like some guidance on whether I should invest more time in this or drop it. Maybe we shouldn't support this kind of functionality due to the complexity versus the benefits.

### DIFF
--- a/test/ex_aws/operation/s3_test.exs
+++ b/test/ex_aws/operation/s3_test.exs
@@ -157,7 +157,7 @@ defmodule ExAws.Operation.S3Test do
       :request,
       fn _method, url, _body, _headers, _opts ->
         assert url ==
-                 "https://examplebucket.s3.amazonaws.com/test%20hello%20%233.txt%3Facl%3D21?version-id=v1"
+                 "https://examplebucket.s3.amazonaws.com/test%20hello%20%233.txt%3Facl%3D21?versionId=v1"
 
         {:ok, %{status_code: 200}}
       end
@@ -168,7 +168,7 @@ defmodule ExAws.Operation.S3Test do
       | path: "test hello #3.txt?acl=21",
         headers: headers,
         bucket: "examplebucket",
-        params: %{"version-id" => "v1"}
+        params: %{"versionId" => "v1"}
     }
 
     assert {:ok, %{status_code: 200}} == S3.perform(operation, config)


lib/ex_aws/request/url.ex
```diff
@@ -14,7 +14,7 @@ defmodule ExAws.Request.Url do
     |> convert_port_to_integer
     |> (&struct(URI, &1)).()
     |> URI.to_string()
-    |> String.trim_trailing("?")
+    |> String.replace_suffix("?", "")
```
speeddragon, line 17: This is needed to support `text.txt?` object, since if the object name ends with the question mark, the `URI.to_string` will add another question mark and the `String.trim_trailing/2` will remove both.

speeddragon:
Thanks for replying @bernardd. To answer to your questions:

1. I'm not sure what could break this, since this is `ex_aws` (a core repository for other dependencies) and I've removed `|> Url.uri_encode()` from `ex_aws/auth.ex`. I would need to run tests on other projects that depend on this one. I will do this to answer this question.

2. A quick check in `boto` repository, I found https://github.com/boto/boto3/issues/3584. They had issues with `+`, so I think `?` isn't supported either. They delegated the request to other teams, but I don't think they support it.

PR #1162, opened 2025-08-20
Fix bucket_as_host with virtual_host causing double bucket in URLs

pepicrft:
> [!NOTE]
> I've written the code in this PR using Claude 4 and I reviewed the code afterwards.

## Problem
When both `virtual_host=true` and `bucket_as_host=true` are configured, S3 URLs are incorrectly constructed with the bucket domain appearing twice.

For example:
- Expected: `https://my-bucket-domain.com/key`
- Actual: `https://my-bucket-domain.com/my-bucket-domain.com/key`

This affects all S3 operations when using custom domain buckets, including multipart uploads.

## Root Cause
The `add_bucket_to_path` function in `ExAws.Operation.S3` only handled the `virtual_host` option but didn't support the `bucket_as_host` option. When `virtual_host=true`, it always constructed the hostname as `"#{bucket}.#{base_host}"`, even when `bucket_as_host=true` was set.

## Solution
Added proper support for the `bucket_as_host` option by:

1. Adding a new pattern match for when both `virtual_host: true` and `bucket_as_host: true` are set
2. When both are true: Use the bucket name directly as the hostname (don't append to base host)
3. Preserving all existing behavior for other configurations

## Changes
- Modified `add_bucket_to_path/2` in `lib/ex_aws/operation/s3.ex` to handle `bucket_as_host` option
- Added comprehensive tests to verify the new functionality

## Testing
- ✅ `virtual_host: true, bucket_as_host: true`: URLs become `https://my-bucket-domain/key`
- ✅ `virtual_host: true, bucket_as_host: false`: URLs remain `https://bucket.s3.amazonaws.com/key`
- ✅ `virtual_host: false`: URLs remain `https://s3.amazonaws.com/bucket/key`

Fixes ex-aws/ex_aws_s3#306

### DIFF
--- a/lib/ex_aws/operation/s3.ex
+++ b/lib/ex_aws/operation/s3.ex
@@ -62,6 +62,12 @@ defmodule ExAws.Operation.S3 do
       raise "#{__MODULE__}.perform/2 cannot perform operation on `nil` bucket"
     end
 
+    def add_bucket_to_path(operation, %{virtual_host: true, bucket_as_host: true} = config) do
+      # When bucket_as_host is true, use the bucket name as the full hostname
+      {put_in(operation.path, ensure_absolute(operation.path)),
+       Map.put(config, :host, operation.bucket)}
+    end
+
     def add_bucket_to_path(operation, %{virtual_host: true, host: base_host} = config) do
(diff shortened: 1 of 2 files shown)


bernardd:
Thanks @himank!

PR #1168, opened 2025-09-17
Fix geo operations

axelson:
Fix constructing geo requests

Without this change when calling a location services API this message is returned from AWS: `{"message":"Credential should be scoped to correct service: 'geo-places'. "}`

Here's an example request that works after this change (that fails on `main`):
```elixir
operation =
  %ExAws.Operation.JSON{
    http_method: :post,
    path: "/v2/suggest",
    data: %{QueryText: "Elixir", BiasPosition: [157.8, 21.3]},
    service: :places,
    headers: []
  }

ExAws.request(operation)
```

Also ensures that GovCloud supports places, currently was getting: `(RuntimeError) places not found in partition aws-us-gov`

Checklist:
* [x] Run `mix format` using a recent version of Elixir
* [x] Run `mix dialyzer` to make sure the typing is correct
* [x] Run `mix test` to ensure no tests have broken (also please make sure you've added tests for your particular change, where appropriate).

### DIFF
--- a/lib/ex_aws/config/defaults.ex
+++ b/lib/ex_aws/config/defaults.ex
@@ -77,7 +77,17 @@ defmodule ExAws.Config.Defaults do
     |> Map.merge(defaults(:timestream))
   end
 
-  def defaults(service) when service in [:places, :maps, :geofencing, :tracking, :routes] do
+  def defaults(:places) do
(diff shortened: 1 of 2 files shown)


bernardd:
Nice one, thanks @axelson!

PR #1174, opened 2025-10-15
Document security_token configuration requirement for instance roles

sebastianmarines:
When using instance roles for authentication, the configuration may fail silently or produce authentication errors if `security_token` is not properly configured. Instance roles generate temporary session tokens that must be included in the configuration alongside `access_key_id` and `secret_access_key`.


bernardd:
Nice one, thanks @sebastianmarines!

PR #1095, opened 2024-10-30
Rename follow redirect in Req adapter

eds314:
Closes #1087

This is my first PR ever!

PR #1175, opened 2025-10-15
Use http opts configured per http client for pod identity

Szetty:
`ExAws.PodIdentity` is not working if Req (`ExAws.Request.Req`) is configured as http_client:
```
iex(1)> ExAws.PodIdentity.security_credentials(%{http_client: ExAws.Request.Req})
** (ArgumentError) unknown option :recv_timeout. Did you mean :receive_timeout?
    (req 0.5.15) lib/req/request.ex:1184: Req.Request.validate_options/2
    (req 0.5.15) lib/req.ex:549: Req.merge/2
    (req 0.5.15) lib/req.ex:1093: Req.request/2
    (ex_aws 2.6.0) lib/ex_aws/request/req.ex:24: ExAws.Request.Req.request/5
    (ex_aws 2.6.0) lib/ex_aws/pod_identity.ex:64: ExAws.PodIdentity.fetch_credentials/2
    (ex_aws 2.6.0) lib/ex_aws/pod_identity.ex:20: ExAws.PodIdentity.security_credentials/1
```

PR #1189, opened 2025-12-11
fix: req support for pod identity

brentjanderson:
The current pod identity implementation provides defaults that are merged into other http options. The Pod identity request defaults include :recv_timeout which is incompatible with Req. By following #1095 we can unblock Req adapter users on pod identity implementation. A more robust fix that would also be a breaking change would involve getting the default HTTP options from the adapter itself for various use-cases, and merging those into the implementation.


brentjanderson:
Saw that this is an alternative to #1175 - the approach to each seems mutually exclusive.

Szetty:
> Saw that this is an alternative to #1175 - the approach to each seems mutually exclusive.

I like this solution better, was not aware we are already doing some renaming for Req, so this approach is more consistent and less intrusive for sure

Szetty:
@bernardd

bernardd:
Yep, looks good, thank you very much @brentjanderson, and apologies for the delay getting to this.

bernardd:
Rebased and tweaked this in #1217 - thanks very much @brentjanderson and @Szetty

PR #1193, opened 2026-01-21
fix(request): wrap HTTP adapters with Code.ensure_loaded? for optional deps

ManuzimFerreira:
## Problem

Users who don't use Req as their HTTP client see this warning during compilation:

```
warning: Req.request/1 is undefined (module Req is not available or is yet to be defined)
└─ lib/ex_aws/request/req.ex:22:12: ExAws.Request.Req.request/5
```

## Solution

Wrap HTTP client adapter modules with `if Code.ensure_loaded?` so they are only defined when the optional dependency is available at compile time.

This follows the same pattern used by [Tesla for optional adapters](https://github.com/elixir-tesla/tesla/blob/master/lib/tesla/adapter/finch.ex).

## Changes

- `lib/ex_aws/request/req.ex` - wrapped with `if Code.ensure_loaded?(Req)`
- `lib/ex_aws/request/hackney.ex` - wrapped with `if Code.ensure_loaded?(:hackney)` for consistency

### DIFF
--- a/lib/ex_aws/request/hackney.ex
+++ b/lib/ex_aws/request/hackney.ex
@@ -1,29 +1,31 @@
-defmodule ExAws.Request.Hackney do
-  @behaviour ExAws.Request.HttpClient
+if Code.ensure_loaded?(:hackney) do
+  defmodule ExAws.Request.Hackney do
+    @behaviour ExAws.Request.HttpClient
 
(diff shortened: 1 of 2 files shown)


bernardd:
Nice one, thanks @ManuzimFerreira!

Issue #1212, opened 2026-03-11
Add `base_path` / `path_prefix` configuration option


pierrelegall:
## Problem

Some S3-compatible servers require a path prefix on all requests. For example:
- A MinIO instance behind a reverse proxy at `https://example.com/minio/`
- An S3-compatible server mounted at a sub-path in a web application, like [PS3](https://github.com/pierrelegall/ps3) mounted via `forward "/s3", PS3.Router` in a Phoenix router, making the S3 endpoint `http://localhost:4000/s3/`

Currently there's no way to configure such a prefix — the URL is always built as `scheme://host:port/<operation_path>`.

## Proposed solution

Add a `base_path` config key that gets prepended to the operation path when building the request URL.

### Examples

Service-level configuration:

```elixir
# PS3 mounted at /s3 in a Phoenix app
config :ex_aws, :s3,
  scheme: "http://",
  host: "localhost",
  port: 4000,
  base_path: "/s3"

# MinIO behind a reverse proxy
config :ex_aws, :s3,
  scheme: "https://",
  host: "example.com",
  port: 443,
  base_path: "/minio"
```

Per-request override:

```elixir
"bucket_name"
|> ExAws.S3.list_objects()
|> ExAws.request(base_path: "/s3")
```

PR #1213, opened 2026-03-11
feature: set url base_path configurable

pierrelegall:
Solve #1212.


bernardd:
Nice, thanks @pierrelegall!

PR #1217, opened 2026-05-04
fix: req support for pod identity

bernardd:
The current pod identity implementation provides defaults that are merged into other http options. The Pod identity request defaults include :recv_timeout which is incompatible with Req. By following #1095 we can unblock Req adapter users on pod identity implementation. A more robust fix that would also be a breaking change would involve getting the default HTTP options from the adapter itself for various use-cases, and merging those into the implementation.


bernardd:
Rebase of #1189

PR #1232, opened 2026-06-10
Handle `HEAD` responses from hackney correctly.

estebanz01:
### Description
- This PR reintroduces the case branch where hackney does not return a body due to performing a `HEAD` request.
- This fix also fixes an incompatibility case with `ex_aws_s3` when using latest hackney for `HEAD`  or `OPTIONS` requests.

### Checklist
- [x] `mix format`.
- [x] `mix dialyzer`.
- [x] `mix test`.

### DIFF
--- a/lib/ex_aws/request.ex
+++ b/lib/ex_aws/request.ex
@@ -146,6 +146,10 @@ defmodule ExAws.Request do
   defp extract_error({:error, error}), do: error
   defp extract_error(error), do: error
 
+  def client_error(%{status_code: status, body: nil} = error, _json_codec) do
+    {:error, {:http_error, status, error}}
+  end
+
   def client_error(%{status_code: status, body: body} = error, json_codec) do
     case json_codec.decode(body) do
       {:ok, %{"__type" => error_type, "message" => message} = err} ->
--- a/lib/ex_aws/request/hackney.ex
+++ b/lib/ex_aws/request/hackney.ex
@@ -23,6 +23,9 @@ if Code.ensure_loaded?(:hackney) do
         {:ok, status, headers, body} ->
           {:ok, %{status_code: status, headers: headers, body: body}}
 
+        {:ok, status, headers} ->
+          {:ok, %{status_code: status, headers: headers, body: nil}}
+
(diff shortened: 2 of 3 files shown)


estebanz01:
Hey @bernardd just pinging you to show this PR as ready! 

At Wenia, when we did the upgrade to the latest version for hackney (we were using 1.x), the `ex_aws_s3` integration broke for us because we were doing some `HEAD` requests to see if the file exists on the S3 bucket, causing the code to generate an exception due to a pattern matching error. This fixes it and handles it to avoid further issues.

Thank you for your review!

walter:
I'm encountering the same issue when upgrading to latest `hackney` (`4.4.5`).

Earlier versions of `hackney` that _do not_ have security vulnerabilities (> `4.0.0`) all have the same API change, too. In other words no way to downgrade to a point where this is working _and_ there are no CVEs against `hackney`.

sax:
I'm running into this same problem when using `ExAws.stream!` with hackney 4.6.

OrangeDrangon:
Using this branch right now in my project and it is working as I would expect.

NexPB:
can we get a release for this 🙏🏻

walter:
Looking at the Release Notes in [this PR](https://github.com/ex-aws/ex_aws/pull/1256) to upgrade `hackney` (which is superseded by an [open PR](https://github.com/ex-aws/ex_aws/pull/1258)), it looks like the API shape that `ex_aws` expects in this case has been restored. I haven't dived deep, but that's my impression from what I read.

If that's true then maybe we just need the latest dependabot PR merged and a new release. A workaround is point at the PR branch until a release.

If anyone gets around to testing this please comment here.

estebanz01:
@Walter it seems that's not the case. Reading the changelog and the associated PR in hackney (https://github.com/benoitc/hackney/pull/917/changes) it's associated to the `Content-Length` header which was not defined instead of the shape of the response for empty-body methods. But I agree some testing is needed regardless!

lemonsrc:
Are there any remaining items blocking this from merging and being released? I would like to help move this along if I can.

PR #1239, opened 2026-06-22
Add Bedrock AgentCore endpoints

anukaver-glia:
Map the Bedrock AgentCore control plane and data plain endpoints. This allows to start using the services with the `ex_aws` library.

The list of supported endpoints relies on
https://docs.aws.amazon.com/general/latest/gr/bedrock_agentcore.html#bedrock_agentcore_region.


anukaver-glia:
@bernardd, would you be able to take a look, please? The change would unblock anyone (including my team) who wants to use this library for Bedrock AgentCore endpoints.

PR #1242, opened 2026-07-10
fix: Hackney pattern match response when body isn't returned

speeddragon:
I found a weird case where one of the images couldn't load after I upgraded to Hackney 4. It is a little bit weird that it only happens on that specific image (I didn't debug it further), but there is a case where a response is returned with no body.


walter:
This is basically a duplicate of #1232

speeddragon:
Thanks, I didn't properly search in the open PR (I only saw the issue and assumed there wasn't any PR for this).

PR #1248, opened 2026-07-31
fix: properly handle hackney head requests

OrangeDrangon:
According to :hackney.request() docs in the case of a head request it returns a 3 tuple. The current implementation only handles the 4 tuple case causing a `ClauseCaseError` when attempting to handle a head request. This was encountered when attempting to download a file using :ex_aws_s3.

https://github.com/benoitc/hackney/blob/9d684fb768837a74c4c7beff0b0e024eb9a9998a/src/hackney.erl#L570-L598

<details>
<summary>Raw docstring</summary>

```
%% @doc Make a request.
%%
%% Args:
%% - Method: HTTP method (get, post, put, delete, query, etc.).
%%   `query' is the RFC 10008 QUERY method: safe and idempotent, with a
%%   request body allowed like post.
%% - URL: Full URL or parsed hackney_url record
%% - Headers: List of headers
%% - Body: Request body (binary, iolist, {form, KVs}, {file, Path}, etc.)
%% - Options: Request options
%%
%% Options:
%% - async: true | once - Receive response asynchronously
%% - stream_to: PID to receive async messages
%% - follow_redirect: Follow redirects automatically
%% - max_redirect: Maximum number of redirects (default 5)
%% - location_trusted: If true, forward auth credentials on cross-host redirects (default false)
%% - pool: Pool name or false for no pooling
%% - connect_timeout: Connection timeout in ms (default 8000)
%% - recv_timeout: Receive timeout in ms (default 5000)
%%
%% Returns:
%% - {ok, Status, Headers, Body}: Success with response body
%% - {ok, Status, Headers}: HEAD request
%% - {ok, Ref}: Async mode - use stream_next/1 to receive messages
%% - {ok, ConnPid}: Streaming body mode (body = stream) - use send_body/2, finish_send_body/1
%% - {error, Reason}: Error
%%
%% Note: The `with_body' option is deprecated and ignored. Body is always returned directly.
```

</details>

<details>
<summary>Redacted error</summary>

```
%CaseClauseError{term: {:ok, 200, [headerlist...]}}
```

</details>

- [x] tests passing
- [x] formatted
- [ ] add  new tests
  - i did not see any existing tests on clients so not sure what to add but happy to if needed


OrangeDrangon:
oops duplicate of #1232

PR #1251, opened 2026-08-02
Fix ExAws.Request.Req: don't rewrite bodyless GETs to POST

### LINKED ISSUE #1246: ExAws.Request.Req sends signed GET requests as POST with req 0.7 (SignatureDoesNotMatch)
## Description

Req 0.7.0 introduced a breaking change: its `encode_body` step automatically converts GET requests to POST when the request body is set ([req CHANGELOG](https://github.com/wojtekmach/req/blob/main/CHANGELOG.md)):

```elixir
# req 0.7, lib/req/steps.ex
defp get_to_post(%Req.Request{method: :get, body: body} = req) when body != nil do
  %{req | method: :post}
end
```

`ExAws.Request.Req.request/5` always passes a `body:` option to `Req.request/1`, and ExAws calls it with `""` for bodyless requests. Since `"" != nil`, **every GET request made through `ExAws.Request.Req` is sent as a POST** (with `content-length: 0`) when req 0.7 is used.

For SigV4-signed requests this is fatal: ExAws computes the signature over a GET request, but the server receives a POST, so the canonical request differs and signature verification fails with `SignatureDoesNotMatch`. Every GET-based operation (ListBuckets, ListObjectsV2, GetObject, etc.) is broken; HEAD/PUT/POST/DELETE operations are unaffected because only `:get` is rewritten.

I hit this in the test suite of an S3-compatible server ([Firkin](https://harton.dev/james/firkin)) which uses `ExAws.Request.Req` as its HTTP client: after upgrading req 0.6.2 → 0.7.x, all GET tests fail with 403 `SignatureDoesNotMatch`, and the server logs show the incoming `list_buckets` request arriving as `POST /` with `content-length: 0`.

## Reproduction

```elixir
Mix.install([{:req, "~> 0.7"}, {:ex_aws, "~> 2.7"}])

# Capture the request Req actually sends, without network access.
adapter = fn request ->
  send(self(), {:wire, request.method})
  {request, Req.Response.new(status: 200, body: "")}
end

# This is what ExAws does for a bodyless signed GET (e.g. S3 ListBuckets)
# when configured with `config :ex_aws, http_client: ExAws.Request.Req`:
ExAws.Request.Req.request(:get, "https://example.com/", "", [], adapter: adapter)

receive do
  {:wire, method} -> IO.inspect(method, label: "method on the wire")
end
```

With req 0.7.x this prints `:post`. With req 0.6.x it prints `:get`.

For comparison, passing `body: nil` instead of `""` preserves the method:

```elixir
ExAws.Request.Req.request(:get, "https://example.com/", nil, [], adapter: adapter)
# method on the wire: :get
```

## Suggested fix

In `ExAws.Request.Req.request/5`, don't pass a body to Req when there isn't one, e.g.:

```elixir
body = if body == "", do: nil, else: body
```

or only include the `:body` option when it's non-empty. Happy to open a PR if that approach sounds reasonable.

## Environment

- ex_aws 2.7.0
- req 0.7.0 / 0.7.1 (works correctly on req 0.6.x)
- Elixir 1.20.2 / Erlang OTP 29

gernotkogler:
Closes #1246.

Req 0.7 [automatically converts a GET into a POST when a non-nil request body is set](https://github.com/wojtekmach/req/blob/main/lib/req/steps.ex). ExAws calls its HTTP client with `""` for bodyless requests (every S3 GET is built with `body: data[:body] || ""`), and `"" != nil`, so **every GET made through `ExAws.Request.Req` goes out as a POST**. Because SigV4 is signed over the original GET, the server rejects it with `SignatureDoesNotMatch` (or, for non-signed backends, 405).

In Req's model "no body" is `nil`, not `""`, so the adapter just needs to translate ExAws's convention:

```elixir
body = if body in [nil, ""], do: nil, else: body
```

Changes:
- fix the shipped `ExAws.Request.Req` adapter (normalise `""` → `nil`)
- add a regression test: a bodyless GET must reach the server as a GET (verified failing as `POST` under req 0.7.2 without the fix, passing with it)
- apply the same fix to the Req adapter example in the `ExAws.Request.HttpClient` moduledoc

Only `:get` is affected by Req's rewrite, so HEAD/PUT/POST/DELETE are unchanged.


sjmadsen:
This kind of addresses #1246, but it would be even better if it fixed the existing Req adapter already in ExAws:

https://github.com/ex-aws/ex_aws/blob/48785c61191355c1f893b5ace045139aa2eded82/lib/ex_aws/request/req.ex#L19

gernotkogler:
Good call — I hadn't spotted the shipped `ExAws.Request.Req` module (it postdates the version I was on). I've now fixed it there as the primary change, added a regression test (verified it fails as `POST` under req 0.7.2 without the fix), and kept the matching fix to the moduledoc example. Updated the PR to close #1246.

PR #1253, opened 2026-08-04
Send bodyless Req requests as nil, not ""

### LINKED ISSUE #1246: ExAws.Request.Req sends signed GET requests as POST with req 0.7 (SignatureDoesNotMatch)

braidonw:
req 0.7's encode_body step rewrites GET to POST whenever the body is set, and "" counts as set, so signed bodyless GETs go out as POSTs and AWS rejects them with SignatureDoesNotMatch. 

Bump the lock to req 0.7.2 so the regression test actually exercises the path.

Fixes #1246


ream88:
This bit us as well, rolled with a custom client for now.

estebanz01:
@einerzg @Alexander-Velasquez let's take a look a this PR. This might be what bit us when bumping req to 0.7

braidonw:
Yeah, we have rolled a custom client as well internally, but would be good to get this fixed globally

PR #1259, opened 2026-08-18
Use consistent URL parsing strategy for signed s3 requests

mashton:
Cause `Auth.presigned_url/9` to construct the uri query using the same S3-aware strategy used in `Url.sanitize/2`

This addresses an inconsistency between the canonical request used to compute an S3 signature and the sanitized URL sent to S3. Existing tests cover special characters in object keys, including the fragment delimiter `#`, but not the combination of `#` and a query string such as `?uploads=1` during multipart upload initiation.

Standard URI parsing treats everythign after `#` as a fragment. As a result, teh query string was omitted from the canonical request even though `Url.sanitize/2` recovered it and included it in the out going request. This results in a signature mismatch.

So here we extract the logic already used in the `Url` module that splits the URL on `?` and use it both in `Url.sanitize` and in `Auth.presigned_url` and `Auth.auth_header`.


mashton:
Please do let me know if you'd prefer an Issue, or if I've missed something entirely here! I know OSS is slammed right now and you have both my deepest appreciation and genuine sympathy 😅

